PR #6, opened 2021-07-21
add BInputGroup & BInputGroupText components + example

Geoffrey-D:


Geoffrey-D:
yes you're right. I will use the properties of BV from now on!

PR #12, opened 2021-07-26
BRow, BListGroup & BListGroupItem (+ quick fix for BIcon)

atilleul:
Hello Carlos,

We implemented all the features for BRow, BListGroup and half of the props from BListGroupItem.
BListGroupItem probably needs to use BLink as in BootstrapVue 2, I'll work on this in the forthcoming days.

About BIcon, we made a quick fix by disabling eslint for the problematic line, at least until we find a better version for this.

Best regards

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -34,7 +34,7 @@ module.exports = {
     "no-useless-constructor": "warn",
     "no-useless-rename": "warn",
     "no-useless-return": "warn",
-    "no-var": "error",
+    "no-var": "warn",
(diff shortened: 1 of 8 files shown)


cdmoro:
Awesome! 
About BIcon, I think we should do the same as BootstrapVue team, They run this script https://github.com/bootstrap-vue/bootstrap-vue/blob/dev/scripts/create-icons.js in order to get all the svg path values of the icons (https://github.com/bootstrap-vue/bootstrap-vue/blob/dev/src/icons/icons.js).

PR #13, opened 2021-07-27
BCol component 

Geoffrey-D:
Hey,

I have done the BCol component &  fix attr binding on BFormInput.

I have also moved the "bootstrap-icons" dependency from "devDependencies" to "dependencies".

Cheers

Geoffrey

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,12 +20,11 @@
     "src"
   ],
   "dependencies": {
+    "bootstrap-icons": "^1.5.0",
     "core-js": "^3.6.5",
     "vue": "^3.0.0"
   },
   "devDependencies": {
-    "bootstrap": "^5.0.2",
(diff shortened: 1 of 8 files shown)


cdmoro:
Great! the only thing is bootstrap-icons. I have doubts about moving bootstrap-icons from 'devDependencies' to 'dependencies'. `bootstrap-icons` package should be optional (as in [BootstrapVue](https://github.com/bootstrap-vue/bootstrap-vue/blob/45635d1d391263192787e9e0bddc004abe115ace/package.json#L108)), the package is necessary only in the case you use the icons and you should install manually.

Geoffrey-D:
As you wish, I liked the fact that it made the library easier to use.

Feel free to move it back to "devDependencies" :)

PR #29, opened 2021-08-12
fix #26

### LINKED ISSUE #26: b-form-select shouldn't add size="0"
b-form-select, without any given size, adds now ` size="0"` to generated markup, which is a bug.

This results in this rule being applied, which hides the dropdown icon:
https://github.com/twbs/bootstrap/blob/649c2bb0bf326db36cbbf7e72c0541b19749a70e/scss/forms/_form-select.scss#L42

mariusa:
fix https://github.com/cdmoro/bootstrap-vue-3/issues/26

### DIFF
--- a/src/components/BFormSelect.vue
+++ b/src/components/BFormSelect.vue
@@ -102,10 +102,10 @@ export default defineComponent({
       };
     });
 
-    const computedSelectSize = () => {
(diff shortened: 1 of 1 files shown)


mariusa:
Error: Could not resolve './BFormCheckbox' from src/components/index.ts
?

PR #36, opened 2021-08-13
fix #28

### LINKED ISSUE #28: b-form-select doesn't set initial value
To replicate, change this line
https://github.com/cdmoro/bootstrap-vue-3/blob/b258f565b032088afe87651141afebeeedee5931/src/App.vue#L1142
to 
```        
selected: 'second',
```

Expected: have 'Item 2' selected by default when view loads
Actual: first option is always selected by default.

mariusa:
b-form-select doesn't set initial value #28


mariusa:
unrelated build error: `Error: Could not resolve './BFormCheckbox' from src/components/index.ts`

PR #37, opened 2021-08-13
Fix build

Hunsu:
The formatting was changed. What tools do we use for this project ?


mariusa:
> The formatting was changed. What tools do we use for this project ?

I use vscode and added formatting settings in .vscode/ , trying to follow the existing project standards.
Feel free to adjust as desired.

PR #40, opened 2021-08-27
feat: Allow passing custom properties to components

AaronDewes:
This adds v-bind="$attrs" to components which didn't have it set before, which allows adding custom classes an properties to components

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -504,23 +504,23 @@
       </b-form>
     </div>
     <div>
-      <b-form-input type="text" />
-      <b-form-input type="number" />
-      <b-form-input type="email" />
-      <b-form-input type="password" />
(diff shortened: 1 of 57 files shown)


cdmoro:
I don't get it. The components already allow custom properties without this change.

If I wrote this:
``` html
<b-alert title="I'm a title" aria-label="I'm an aria label">I'm an alert</b-alert>
```

it would be rendered as:
``` html
<div class="alert alert-info" role="alert" title="I'm a title" aria-label="I'm an aria label" data-v-7ba5bd90="">
    I'm an alert
</div>
```

I didn't test all the components, but I'm pretty sure that this is a common behavior.

AaronDewes:
Then I probably did something wrong, sorry.

PR #52, opened 2021-08-30
trying to fix tests. Still getting error "Must use import to load ES Module:"

mariusa:
stuck on error "Must use import to load ES Module:". Tried various fixes. Maybe somebody else knows how to solve this?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,7 +14,6 @@ pnpm-debug.log*
 
 # Editor directories and files
 .idea
-.vscode
 *.suo
 *.ntvs*
 *.njsproj
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "editor.tabSize": 2
+}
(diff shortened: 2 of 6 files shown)


jclijmans:
I've got the unittest running by doing following starting from the main branch

installed following packages as dev dependencies : @babel/plugin-transform-runtime and  @babel/preset-env
```
   yarn add --dev @babel/preset-env
   yarn add --dev @babel/plugin-transform-runtime
```
I updated babel.config.js to be

```
module.exports = {
  "presets": ["@babel/preset-env"],
  "plugins": [
    ["@babel/transform-runtime"]
  ]
}
```

After that the unittests start running again 
Also the docs and the application itself are working

I'm no expert in node/babel/webpack, not even a novice , but this seems to be working

I've also tested it with vite-jest, like you did and that also allows me to run the tests.
I just installed vite-jest as dev dependency
`  yarn add --dev vite-jest`
And updated the test:unit to run vite-jest instead of jest

cdmoro:
Guys, do we have any news on this? We should fix this ASAP. I'm not an expert on tests nor github workflows, so I'm not much useful :cry:

mariusa:
@jclijmans 's suggestions work for me too, thanks! I'll close this PR and make a new one with those.

PR #56, opened 2021-08-31
feat: Table fields first draft

jclijmans:
First version of the table fields functionality.
Currently only key and label properties implemented

Table should work without fields, fields as just string array or fields as TableFieldObject array


src/App.vue
```diff
@@ -1236,6 +1283,12 @@ export default defineComponent({
           { age: 89, first_name: 'Geneva', last_name: 'Wilson' },
           { age: 38, first_name: 'Jami', last_name: 'Carney' }
         ];
+    const stringTableDefinitions = ref(['last_name', 'first_name', 'age']);
+    const objectTableDefinitions = ref<TableFieldObject[]>([
+      {key: 'last_name', label: "Family name"},
+      {key: 'first_name', label: "Given name"},
+      ]);
```
a-kriya, line 1290: Indent. The code below in other files has formatting issues, too.

@cdmoro I'd suggest setting up lint + prettier to auto-run on git hooks to save time for PR creators and reviewers.

jclijmans: It would be nice to have a .prettierrc that specifies the formatting rules like spaces or tabs, print width, ...
So that vscode automatically formats the code correctly
I've currently disabled FormatOnSave to not force my local rules upon this project, but then I've got to maintain it myself.

I can create one, but I'm not sure what the formatting rules are.

mariusa: I'm also using vscode and would love to commit .vscode/ with workspace settings.

cdmoro: > It would be nice to have a .prettierrc that specifies the formatting rules like spaces or tabs, print width, ...
> So that vscode automatically formats the code correctly
> I've currently disabled FormatOnSave to not force my local rules upon this project, but then I've got to maintain it myself.
> 
> I can create one, but I'm not sure what the formatting rules are.

It's a good idea to have a .prettierrc file, feel free to create it!

About the rules, I think most of the default rules are just fine. According to Prettier [official documentation](https://prettier.io/docs/en/options.html), just to recap:

- Print width: 80
- Tab width: 2
- Tabs: false
- Semicolons: true
- Quotes: true (default: false)
- Trailing commas: "es5"
- Bracket spacing: true
- Arrow function parentheses: "always"
- End of line: "lf"

mariusa: > It's a good idea to have a .prettierrc file, feel free to create it!

It's not a universal solution. I use vscode, but with the built-in formatting, no prettier. That's why I'd prefer .vscode/

But if others use something else? How do other projects handle this? @a-krya said ' lint + prettier to auto-run on git hooks ' ?

a-kriya: I don't think just having `.vscode` is enough -- it won't cover anyone not using it.

Git hooks are also important as not everyone will remember running `npm run lint` before committing.

https://vueschool.io/articles/vuejs-tutorials/eslint-and-prettier-with-vite-and-vue-js-3/
https://www.npmjs.com/package/husky

a-kriya: I can put up a separate PR with that if someone's not already on it.

cdmoro: I only have a question, this type `TableFieldObject` shouldn't be a union type with the string and object type of fields? According to the BootstrapVue table, the fields prop allows an object and string as well. This type should be something like that:

``` typescript
interface TableFieldObject {
    key: string,
    // other fields
}

type TableField = TableFieldObject | string
```

cdmoro:
I only have a question, this type `TableFieldObject` shouldn't be a union type with the string and object type of fields? According to the BootstrapVue table, the fields prop allows an object and string as well. This type should be something like that:

``` typescript
interface TableFieldObject {
    key: string,
    // other fields
}

type TableField = TableFieldObject | string
```

src/App.vue
cdmoro: It seems ok to me! Go on :sunglasses:

a-kriya: I figured I'll be doing it from all the thumbs up : ]

Will put something up soon.

jclijmans:
I've updated App.vue to use the TableField.

Indeed, TableField should be used by a consuming component.
But internally, I normalize the fields to a TableFieldObject, so that the code in BTable.vue does not have to determine if it's a string or an object each time the fields property is used.

cdmoro:
Perfect! I agree with normalizing fields, I suppose that normalization is reactive, right? If you update the fields it should update the rendered table columns too

jclijmans:
It's a computed, so it should be reactive.

But, to be sure, I've tested it out and it's indeed reactive both when it's just an array of strings and when it's an object array.

Issue #46, opened 2021-08-28
BFormCheckbox plain property not compatible with bootstrap-vue


jclijmans:
In bootstrap-vue-2, the plain property seems to behave exact the opposite of the behavior of bootstrap-vue.

In bootstrap-vue from the documentation :
property plain, type Boolean, Default false, Render the form control in plain mode, rather than custom styled mode

In bootstrap-vue-3, the plain property has default value true and it's behavior is exactly the opposite of the behavior in bootstrap-vue.
If I set the property plain to true, the checkbox is rendered in custom styling.

This seams to me to be very confusing. If I set the plain property to true, I expect it to rendered plain (As the browser renders it)
Also it's not helping people migrating from bootstrap-vue.

Is this intentionally, or is this a bug?

jclijmans:
@cdmoro : Do you agree that this is a bug and needs to be fixed?

If you agree, I'll have a look at reversing the plain property behavior

cdmoro:
Yes, this is a bug, definitely. It would be nice if you could fix it, thanks!

Issue #30, opened 2021-08-12
Please add b-form-radio-group / b-form-checkbox-group 


mariusa:
similar to https://bootstrap-vue.org/docs/components/form-radio

Geoffrey-D:
Hey @mariusa,

I have implemented the "BFormCheckboxGroup" component in the pull request https://github.com/cdmoro/bootstrap-vue-3/pull/64.

There are probably some fine tuning to be done but the basic features are working.

Let me know if you find any issues with that component :)

mariusa:
Excellent, Geoffrey! Thanks so much!
I've tested your fork with a couple of use cases and it works great :grin:   

Could you also do the `b-form-radio-group` ?

Geoffrey-D:
Yes that's planned :)

Geoffrey-D:
b-form-radio-group is waiting to be merged in https://github.com/cdmoro/bootstrap-vue-3/pull/67 :)

mariusa:
Great news!

Carlos, would you please push a new release afterwards too? There have been
quite some changes. We also need to test vite build.

Wishing you a great weekend :)

cdmoro:
I've published a new version: 0.0.2 :smile:

mariusa:
`b-form-radio-group` and `b-form-checkbox-group` work great in 0.0.3, thanks so much!

Geoffrey-D:
you are welcome! Thank you for the feedback, it's very motivating :)

PR #57, opened 2021-09-03
FormCheckbox v-model support

### LINKED ISSUE #49: BFormCheckbox does not support v-model
the BFormCheckbox component does not support v-model

      <div>{{ vartocheck }}</div>
      <b-form-checkbox v-model="vartocheck">Default</b-form-checkbox>

and the the setup method in the script
  setup() {
    const vartocheck = ref(false);
    return {
      vartocheck
    }
});

jclijmans:
v-model support for FormCheckbox. It fixes #49 
v-model supports boolean, string values set in value/unchecked-value as well as an array.

The model binds to the standard modelValue prop, so I have removed the checked prop

Doc is also updated.
As soon as I can figure out how to use vue components or real javascript code in the markdow, I'll add live examples in the documentation


cdmoro:
One thing, could you have a look at #46 please?

jclijmans:
I have raised the issue because I think the current implementation is reversed.
I did not get a response from you that indeed it should be reversed.

If so, I'll take a look at reversing the plain property

mariusa:
Thanks @jclijmans !

> as well as an array.

Does this mean #30 is no longer needed?

jclijmans:
For checkboxes where you select multiple items out of an array, the grouping is not needed. See App.vue for an example

For radio buttons where you want 1 item selected out of the array. That doesn't work with the current implementation.
There is some contention between the different components all trying to update the single model

mariusa:
Thanks for clarification!

PR #62, opened 2021-09-05
fix unit tests

mariusa:


mariusa:
Now there's a `yarn` error. I'm using only `npm`, so not sure what's the best fix (how to update the yarn lock file).

jclijmans:
After fetching the latest code, installing via yarn install --frozen-lockfile, the install failed with an error "Your lockfile needs to be updated, but yarn was run with `--frozen-lockfile`." Same error as the build.
When debugging it, the packages the install complained about were :
@babel/plugin-transform-runtime@^7.15.0' and @babel/preset-env@^7.15.4
Removing the node_modules folder and running the install again and now the install completed.

I see in the test build, that the node_modules are retrieved from the cache. (from previous runs?)
Probably there are some other versions in that cache

For the tests in the CI we could either not retrieve the node_modules from cache or run the yarn install without the --frozen-lockfile
For that you need to update the .github/workflows/test.yml file.
There are a number of install commands for different environments, all with --frozen-lockfile and using the node_module cache.
You could also update the cache keys used, so that it results in a cache miss

According to github documentation, GitHub will remove any cache entries that have not been accessed in over 7 days.
But there is currently no way to manually clear the cache.

PR #63, opened 2021-09-06
Add a working BIcon component

AaronDewes:
Based on https://github.com/dvuckovic/vue3-bootstrap-icons. Because that is licensed under the WTFPL, it can easily be added here.


jclijmans:
Great work.

Some personal remarks :
- This project is typescript oriented, so I would like to see the properties variant, size, animation to be typed. See BCard.vue on how typed properties are defined. You can make separate types for these in the /src/types folder.
- I would like the styles to be defined in the /src/styles/styles.scss file instead of defined as scoped styles. I think scoped styles are copied for each instance. Having multiple BIcon components on a page would increase the size of the page.
- Can you put a working example in App.vue? That way when someone wants to work on it, a working example to test with is readily available.

AaronDewes:
Thanks for the suggestions, I'll add all of these.

src/App.vue
```diff
@@ -1199,7 +1220,6 @@ export default defineComponent({
     const radioSelected = ref()
 
     onMounted(() => {
-      // input.value?.focus();
```
AaronDewes, line 1202: I didn't remove this, I guess it was the linter?

src/components/BIcon.vue
```diff
@@ -1,284 +1,94 @@
...
-                'spin',
-                'spin-reverse',
-                'spin-pulse',
-                'spin-reverse-pulse',
-                'throb',
-            ].indexOf(value) !== -1,
-        },
+    animation: {
+      type: String as PropType<Animation>,
     },
```
AaronDewes: Should I still include the runtime validation logic here for JS users? If yes, is using Array.contains okay?

AaronDewes:
Also, how about swichting to Yarn 3? Yarn 3 saves itself in the project directory, so everyone runs the same version.

src/components/BIcon.vue
```diff
@@ -1,51 +1,94 @@
...
     </g>
   </svg>
 </template>
 
 <script lang="ts">
-import {computed, defineComponent, PropType} from 'vue'
-import {ColorVariant, InputSize} from '../types'
+// https://github.com/dvuckovic/vue3-bootstrap-icons
+import {defineComponent, PropType} from 'vue'
+import BootstrapIcons from 'bootstrap-icons/bootstrap-icons.svg'
```
cdmoro, line 12: I have a doubt on this line. I thought the main goal is the ability to import a single SVG path (like BootstrapVue) per icon, according to this `import` if you want to use only one icon you have to import all of them, which is not the best practice.

AaronDewes: https://dvuckovic.com/2021/03/12/vue-bootstrap-icons/

> However, I found that none of them was based on the sprite method, and they opted for importing every icon on its own for tree-shaking reasons.

> For sure, this is the best option, but only when you can afford it. Again, what if you don’t know which icons will be needed at build time? You still need to import all of them at some point, but normally they will still be considered as separate resources and might result in separate requests for each icon usage. Yikes!


If you still want to use a different method, I can try to change it.

cdmoro: I think we should do the same as the BootstrapVue team, They run this script https://github.com/bootstrap-vue/bootstrap-vue/blob/dev/scripts/create-icons.js in order to get all the SVG path values of the icons in one file (https://github.com/bootstrap-vue/bootstrap-vue/blob/dev/src/icons/icons.js).

Of course, this method is a bit complex, but it allows to import a single icon instead of all of them. For now, we can keep your implementation, but I think in the future we should consider this.

src/components/BIcon.vue
jclijmans: I've looked into other components and they don't have validators. Also bootstrap-vue BIcon does not have validators for the properties.

So for me, I don't see why this component should have them. If the properties are not valid, they also put invalid classes on the elements.

PR #64, opened 2021-09-06
BDropDown (Group, Toolbar, Divider & ItemButton) + BFormCheckboxGroup + BFormCheckbox improvements

Geoffrey-D:


mariusa:
I've tested this with a couple of use cases and it works great :grin:   Thank you, Geoffrey!

PR #66, opened 2021-09-08
Fix formcheckbox and formradio. Fixes #65

### LINKED ISSUE #65: BFormChecbox and BFormRadio don't work anymore after PR #64
Simple checkbox like this doesn't work

      <div class="row">
        <b-form-checkbox v-model="checkedDefault" class="col-4">Default</b-form-checkbox>
        <div class="col-6">Checked: {{ checkedDefault }}</div>
      </div>

There is a syntax error in the BFormCheckbox template, but even with that fixed, it still doesn't work

I'll see if I can figure out what's going on and try to make it work again.

jclijmans:
Fixes issue #65 

- value property must have default value true. Bug existed prior to #64 , but somehow the new code made this causing the simple checkbox failures
- #64 removed the ability to have an array as v-model.

- additional fixes for FormRadio 

Still open issues :
- input event is removed. We need to either update the docs or make input event working again
- lots of unittests fail due to the complete change of the inner working of the FormCheckbox
   - fix unittests and remove tests no longer relevant

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -1,46 +1,6 @@
 <!-- eslint-disable vue/max-attributes-per-line vue/singleline-html-element-content-newline -->
 <template>
   <b-container class="mt-4" fluid="sm">
-    <h5 class="my-3">Checkboxes</h5>
-    <h6 class="m-2">Simple</h6>
-    <div class="m-4">
-      <b-form-checkbox
-          id="checkbox-1"
-          v-model="checkboxes.status"
-          name="checkbox-1"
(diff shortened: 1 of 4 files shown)


Geoffrey-D:
Hey @jclijmans 

Unit testing & eslint errors are fixed in https://github.com/cdmoro/bootstrap-vue-3/pull/67

Input event should also be triggered now

I still need to work on BFormRadio to have it working with BFormRadioGroup (probably next week)

PR #67, opened 2021-09-09
BFormRadioGroup (unfinished) + BFormRadio & BFormGroup improvements + fix build errors

Geoffrey-D:
Fix unit testing & eslint errors

BFormRadio, BFormRadioGroup, BFormCheckbox & BFormCheckboxGroup improvements

Due to Radio changes implied by https://github.com/cdmoro/bootstrap-vue-3/pull/66 some more changes need to be done for BFormRadioGroup to work


src/components/BFormCheckbox/BFormCheckbox.vue
```diff
@@ -57,24 +63,51 @@ export default defineComponent({
...
     const classes = getClasses(props)
     const inputClasses = getInputClasses(props)
     const labelClasses = getLabelClasses(props)
 
-    const handleClick = (checked: any) => {
+    watch(
+      () => props.modelValue,
+      (newValue) => {
+        emit('input', newValue)
+        emit('update:modelValue', newValue)
```
jclijmans, line 90: Is emitting update:modelvalue needed? An update to the modelValue prop causes the emit wich then cause modelValue to be changed by the vue binding system again

jclijmans:
I've tested the BFormRadio and it's broken again.

Setting the model value from code, does not cause the radio buttons to change.  See the example in App.vue where you can set the "individual grouped" radiobuttons using the buttons underneat

I had fixed it by removing the toggleChecked from BFormRadio.vue and useFormCheck.ts 
and by changing line 47 in useFormCheck.ts from
`    if (checked) return checked`
to
`   if (Array.isArray(propsChecked) && checked) return checked`

Maybe it's needed for BFormRadioGroup?
When I change it back, the example works but 4 unittests fail, so maybe I'm missing something or the unittests are faulty

Further, I get eslint warnings in  useFormChecks.ts for the new functions you have added like _getComputedAriaInvalid
Warnings for the use of any, missing return types, eslint(arrow-body-style) and even errors for using String and Boolean as types instead of string and boolean

Geoffrey-D:
The BFormRadio component never really fully worked on any example of https://bootstrap-vue.org/docs/components/form-radio ( even though it was passing unit testing).

This is why I rewrote the component (as well as BFormCheckbox which is OK now and works fine with both BFormCheckboxGroup & unit testing)

Before your PR the "rewritten" component of the branch https://github.com/Efficy/bootstrap-vue-3 was working with all examples of the bootstrap-vue documentation (and with the BFormRadioGroup) but was causing issue with some example of the App & unit testing.

I think that we need to come back to a situation close to https://github.com/Efficy/bootstrap-vue-3/commit/54c7177967891d3b3ce35b12d81712f829d0441b#diff-ba8bc04fc168c4e78c3d37891fd606a96651847bb84b7bca6e4ae88bb22e323d

To do that, we should focus on removing the method "useFormCheck" (that make things unnecessarly complex), and to follow what has been done on BFormCheckbox (the 2 components are very close).

I will try to have a look at it next week. In the meantime I think that this PR can be merged

jclijmans:
The main issue I have with the current PR is that updating the model in code does not cause the checkbox to change
2-way databinding should work and currently it isn't

You are right about the "useFormCheck" being to complex and it should be removed. and the work you've done so far is great.

cdmoro:
Should I merge this pull request or do you want me to wait until you finish the BFromRadioGroup component? Thanks for your awesome work!

Geoffrey-D:
I have finished the rewrite of the BFormRadio. 

Now the component is simpler, all examples are working and I have added some BFormRadioGroup examples.

Be aware that the unit testing of BFormRadio has currently 4 errors, all due to the fact that for some reason the component datas 
 (classes or model value) are not updated after a "click":

```
await input.trigger('click')
expect(label.classes()).toContain('active') // Works in browser but not in unit test
```

To my opinion @cdmoro you can merge this PR, and I would be very pleased if someone else could look at this unit-testing component update issue :)

src/components/BFormCheckbox/BFormCheckbox.vue
Geoffrey-D: You are right, thanks for pointing that out! I removed that in https://github.com/Efficy/bootstrap-vue-3/commit/db36c2c7aaba9f532a0ebb33ac5771202b2ae787 :)

jclijmans:
The unittests are failing because vue-test-utils doesn't handle 2-way data-binding
There is a lengthy discussion regarding the issue but no resolution yet. see [#279](https://github.com/vuejs/vue-test-utils-next/discussions/279)

There is a work-around by handling the update:modelValue yourself and updating the modelValue property

I'll create a PR for this.

However there is 1 test still failing for which I'm not sure that the BFormRadio implementation is correct.
Test 'works when v-model bound to an array'

The test verifies that when the BFormRadio is bound to an array, clicking a checked radio does not cause a change event.
But the current implementation does emit a change even because the value is removed from the array.
In the test the localChecked changes from ['foo', 'bar'] to ['foo'] and thus triggers a change event

When I test this in the current bootstrap-vue implementation with the cars example from App.vue, only 1 radio can be selected at anytime, a radio cannot be unchecked

I'll create an issue for it, so the implementation can be discussed and possible changed

PR #70, opened 2021-09-10
fix files published to include all dist

mariusa:


mariusa:
Please publish a new version to npm, 0.0.2 is broken due to wrong "exports" in package.json
Tested this locally.

PR #88, opened 2021-09-15
fix typescript compilation errors

Geoffrey-D:
Those fixes are required to be able to use `bootstrap-vue-3` in a typescript project.

NB: I had to remove the "autoClose" feature of the dropdown for now

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -122,7 +122,7 @@
     <h6 class="m-2">Custom Button Icon</h6>
     <b-dropdown text="Custom Button Icon" no-caret class="m-2" variant="link">
       <template #button-content>
-        <img src="@/assets/logo.png" style="height: 1em" />
+        <img src="./assets/logo.png" style="height: 1em" />
       </template>
       <b-dropdown-item href="#">Action</b-dropdown-item>
     </b-dropdown>
@@ -1296,7 +1296,7 @@
 <script lang="ts">
 import {defineComponent, onMounted, ref} from 'vue'
 import {useBreadcrumb} from './composables/useBreadcrumb'
-import BDropdown from '@/components/BDropdown.vue'
+import BDropdown from './components/BDropdown.vue'
 import TableField from './types/TableField'
 
 export default defineComponent({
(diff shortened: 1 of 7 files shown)


jclijmans:
Why doesn't the @ alias for src work?
It's defined in vite.config.ts

It's much easier to maintain your imports using the alias than working with relative folders

If @ doesn't work maybe it's defined incorrectly
Normally it should be defined as follows in vite.config.ts

import vue from '@vitejs/plugin-vue'
const path = require('path')

export default {
  alias: {
    '@': path.resolve(__dirname, './src')
  },
  plugins: [vue()]
}

Geoffrey-D:
The project I am currently using for testing is using `vue-cli`.

having absolute path like "/@" generate errors like this one:

```
 ERROR  Failed to compile with 1 error 

This dependency was not found:

* @/components/BFormRadio/BFormRadio.vue in ./node_modules/bootstrap-vue-3/src
/components/BFormRadio/index.ts

```

jclijmans:
You can also define the alias for typescript in tsconfig.json

{
  "compilerOptions": {
    .....
    "paths": {
      "@/*": ["src/*"]
    },
  },
 .........
}

Geoffrey-D:
I currently have this in my tsconfig.json:

```
"compilerOptions": {
...
    "paths": {
      "@/*": ["src/*"]
    },
...
}
```

Yet it is not working when there are absolute "@/" import in bootstrap-vue-3 files, and couldn't find a way to make it work with this synthax

jclijmans:
Ha, I now see the error in your compile
It's looking for @/components/BFormRadio/BFormRadio.vue in ./node_modules/bootstrap-vue-3/src
/components/BFormRadio/index.ts

But of course BFormRadio.vue is not in your src/components folder but in node_modules/....

Not sure how libraries should be build, but shouldn't these dependencies be resolved when you build bootstrap-vue-3 for deployment? Have you just copied the source of bootstrap-vue-3 to the node_modules folder?

cdmoro:
I think we should use relative paths instead of absolute paths for now. At least, until we can figure out how to use absolute paths. If we all agree I can merge this PR.

jclijmans:
ok for me

Geoffrey-D:
@jclijmans it has been installed using `npm i bootstrap-vue-3`

I think that we can merge it for now

Issue #141, opened 2021-10-12
BNavItem does not support links


mattmess1221:
The `<a>` tag in `<b-nav-item>` has the `href="#"` hard-coded, which makes it difficult to use it with vue-router's `<router-link>` without nesting them.

https://github.com/cdmoro/bootstrap-vue-3/blob/99c75411936f452f3020a85a7c6e014f5a7ee96e/src/components/BNavItem.vue#L3-L8

PR #143, opened 2021-10-12
NavItem : add href prop

jclijmans:
Added an href property on BNavItem to address #141

BNavItem is far from completed, but hopefully this handles the issue until the BNavItem is finished


jclijmans:
Bootstrap-Vue NavItem has the href property without default. 
That's why I didn't add a default. If you want to, I can implement the default

PR #144, opened 2021-10-12
Fix BFormInput formatter

### LINKED ISSUE #126: BFormInput formatter does not work
The formatter property for BFormInput does not work.
The format function is called, but the resulting formatted text is not put into the input control.

To test :

```
<template>
  <b-card>
    <div role="group">
      <label for="input-formatter">"Text input with formatter (on input)"</label>
      <b-form-input
        id="input-formatter"
        v-model="formatInputText"
        placeholder="Enter your name"
        :formatter="lowerCaseFormatter"
      ></b-form-input>
      <b-form-text>We will convert your name to lowercase instantly</b-form-text>
    </div>
  </b-card>
</template>

<script lang="ts" setup>
import {ref} from 'vue'

const formatInputText = ref('')
const lowerCaseFormatter = (value: string) => {
  const result = value.toLowerCase()
  return result
}
</script>

```

stevemandl:
This is an attempt to fix #126. All tests pass, but I was unable to add a test to verify this is working. I have sample code to add to a demo app to see it in action that I can add to the PR if requested.

### DIFF
--- a/src/components/BFormInput/BFormInput.vue
+++ b/src/components/BFormInput/BFormInput.vue
@@ -93,14 +93,15 @@ export default defineComponent({
   setup(props, {emit}) {
     const input = ref<HTMLInputElement>()
     let inputValue: string | null = null
-    let formatterResult: any = null
+    let neverFormatted = true
     const computedId = useId(props.id, 'input')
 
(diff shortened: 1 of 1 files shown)


cdmoro:
I understand that this PR is ready to merge, right?

PR #146, opened 2021-10-13
BPopover improvements

Geoffrey-D:


jclijmans:
Sorry for the late review

properties nonhtml and nonsanitize that are false by default, aren't those double negatives?

In my opinion it would have been better to call these properties html and sanitize with default value false

So the Popover options would be
    html: props.html,
    sanitize: props.sanitize,

Easier to understand and read.

Just my 2 cents

PR #147, opened 2021-10-13
Initial SSR test

xvaara:
DON'T MERGE!

this will need to wait for upcoming bootstrap 5.2. Testing purposes only at the moment.

Uses bootstrap from my github. so you need to remove bootstrap from node_modules and force npm/yarn to install it again. (I just nuke the lock file).


xvaara:
I also noticed that with these changes you can move to vuepress-vite, since it uses ssr to render pages. 😄

aceofwings:
Yeah I still going to be a while but I cleaned up the vuepress pages. For now boostrap will have to render client side.

VividLemon:
I'm going to close out this issue. I get that it's a test PR, but at this point there has been many library changes that render the test incompatible, and obviously merge conflicts exist. If you want to, Xvaara, you can redo this, and instead we can make a separate branch for it.

Issue #159, opened 2021-10-18
Missing b-input-group-append and b-input-group-prepend


runemy:
Hi,

Thanks for all the work that is done on this.
Guess there is some tracking on missing components and I hope b-input-group-append and b-input-group-prepend is in that list :)

https://bootstrap-vue.org/docs/components/input-group#using-sub-components

PR #161, opened 2021-10-19
Implement BInputGroup append , prepend

jclijmans:
- Implement BInputGroupAppen, BinputGroupPrepend, BInputGroupText and BInputGroupAddon
- docs

**Note** Bootstrap 5 dropped `.input-group-append` and `.input-group-prepend`. You can now just add buttons and `.input-group-text` as direct children of the input groups. 
To implement these sub-components, I had to do some hacking in styles.
The sub-components are added for compatibility reasons, the preferred way is the Bootstrap5 way.

This should handle #159

### DIFF
--- a/docs/components/InputGroup.md
+++ b/docs/components/InputGroup.md
@@ -57,7 +57,7 @@
 
 You can attach addons using either props, named slots and/or sub-components.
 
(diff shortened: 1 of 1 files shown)


jclijmans:
I've already added the comments about them not being recommended at several places in the docs.

LIke this:

**Note** Bootstrap 5 dropped `.input-group-append` and `.input-group-prepend`. You can now just add buttons and `.input-group-text` as direct children of the input groups. The sub-components are added for compatibility reasons.

If stronger wording is needed, just let me know

cdmoro:
Sorry! I totally omited it! You're right, never mind!

Issue #188, opened 2021-10-25
Add ability to import individual components


axelthat2:
If I'm not mistaken, currently we have to import the whole library to use any of the bootstrap vue components. I believe we should have ability to import individual components enabling tree shaking.

```
import { BAlert  } from "bootstrap-vue-3"
```

Also shown here: https://bootstrap-vue.org/docs/#individual-components-and-directives

PR #194, opened 2021-10-26
Upgrade vite and @vitejs/plugin-vue

jclijmans:
In response for #188
Don't know if that's al that's needed to have the ability to import individual components with tree-shaking

Updated vite to v2.6.12, updated @vitejs/plugin-vue to v1.9.3
Everything seems to work without errors or warnings

Did not update package.json because that specified 
vite 2.x.x
@vitejs/plugin-vue 1.x.x

If I need to update package.json, please let me know.


axelthat2:
I don't think we need to do `bootstrap-vue-3["default"]`.

We could do just use it as,
```
import { BAlert } from "bootstrap-vue-3"
```

or if we want to use the whole library,
```
import BootstrapVue from "bootstrap-vue-3"
```

Am I missing something here?

axelthat2:
We can set `build.rollupOptions.output.exports` to `named`.

https://github.com/rollup/rollup/issues/2106#issuecomment-377771639

axelthat2:
@jclijmans What do you think of the above?

jclijmans:
I'm sorry but I don't have experience with vite and rollup, so I don't know if this not just masks the problem.

The user should be able to use bootstrap-vue-3 by just doing as stated in the docs

```
import {createApp} from 'vue'
import BootstrapVue3 from 'bootstrap-vue-3'

import 'bootstrap/dist/css/bootstrap.css'
import 'bootstrap-vue-3/dist/bootstrap-vue-3.css'

const app = createApp(App)
app.use(BootstrapVue3)
app.mount('#app')

```
If that's no problem, then I'm ok with this solution.

axelthat2:
They can. I have tried already. One of the dev at rollup (https://github.com/rollup/rollup/issues/2106#issuecomment-377771639) already said the error message is slightly out of date. But I can verify again.

jclijmans:
Ok, I'll create a PR for this

axelthat2:
I believe the error message was for the CJS modules.

So, I think for CJS modules you have to do,

```
const BootstrapVue = require("bootstrap-vue-3")["default"]
```

Screen Shot 2021-10-31 at 8 09 15 PM

But we are not generating CJS so we are fine as you can see in the screenshot below.

Screen Shot 2021-10-31 at 8 10 07 PM

jclijmans:
Ok, the fix is in.
@axelthat2 : can you test that it solves the problems with individual import of components please?

axelthat2:
Sure!

axelthat2:
@jclijmans Alright I have tested it and works.

axelthat2:
I think you should also add this `export {plugin as BootstrapVue}`.

If you see here: https://bootstrap-vue.org/docs, you can also do,

```
import { BootstrapVue } from "bootstrap-vue"
```

jclijmans:
Done, added 
`export {plugin as BootstrapVue3}`

so you have to do 
`import { BootstrapVue3 } from "bootstrap-vue-3"`

PR #203, opened 2021-10-31
Fix #202

### LINKED ISSUE #202: BFieldGroup changes it's id whenever the input value changes
So each letter you type on the input, `b-field-group` changes it's id each time.

https://user-images.githubusercontent.com/91611516/139567458-549e0cab-afdb-4cd1-a09a-690b3066c5f9.mov

Code:
```
<script setup>
import { ref } from "vue"

const input = ref("")
</script>

<template>
  <b-form-group class="p-5">
    <b-form-input placeholder="Name" v-model="input" />
  </b-form-group>
</template>

<style></style>
```

Dependencies:

```
{
  "dependencies": {
    "bootstrap": "^5.1.3",
    "bootstrap-vue-3": "^0.0.5",
    "vue": "^3.2.16"
  },
  "devDependencies": {
    "@vitejs/plugin-vue": "^1.9.3",
    "sass": "^1.43.4",
    "vite": "^2.6.4"
  }
}
```

axelthat2:
~~So the reason it does that is because whenever the component property (v-model) changes `render()` will create a new, different component, which will have to be mounted.~~ Now the solution is to place all `render` logic inside `setup`. `setup` should however return a function which returns the element. 

So, instead of doing this,
```
render() {
  return h(...)
}
```

We do the following,
```
setup() {
  return () => h(...)
}
```

This way the component will be created only once.

~~Related issue which I faced a year ago: https://github.com/vuejs/vue-next/issues/3235#issuecomment-779308808~~

I see some few other places where you return the element inside `render`. I think we should use `setup` instead. Just let me know and I will do the changes.


jclijmans:
I started by putting render functions as return from setup. But if I recall got some issues with reactivity.
Don't remember the circumstances so it might have been a problem in my code  

Another possible solution might be to calculate the id in the setup and pass that to the render function.

But for me this fix is ok

axelthat2:
If you got issues with reactivity then I am not sure if we should merge this PR. Could you maybe try to remember what component were you coding at that time?

jclijmans:
For me this fix is ok. if you can run the samples from the docs, because those are my test cases as well when I build the BFormGroup

cdmoro:
> So the reason it does that is because whenever the component property (v-model) changes `render()` will create a new, different component, which will have to be mounted. Now the solution is to place all `render` logic inside `setup`. `setup` should however return a function which returns the element.
> 
> So, instead of doing this,
> 
> ```
> render() {
>   return h(...)
> }
> ```
> 
> We do the following,
> 
> ```
> setup() {
>   return () => h(...)
> }
> ```
> 
> This way the component will be created only once.
> 
> Related issue which I faced a year ago: [vuejs/vue-next#3235 (comment)](https://github.com/vuejs/vue-next/issues/3235#issuecomment-779308808)
> 
> I see some few other places where you return the element inside `render`. I think we should use `setup` instead. Just let me know and I will do the changes.

Return in setup function is the correct behavior, could you make the necessary changes in the other files, please?

axelthat2:
Sure!

axelthat2:
I think I found the reactivity issue. So in `BOverlay` you have this,

```
let $overlay: any = ''

if (props.show) {
....
$overlay = h(...)
}
```

Assuming we have placed all render logic inside setup, when `props.show` is toggled then overlay element will not be shown because inside `setup` the component is created only once and all the logic have been ran already. Now to fix this, all we need is to wrap the logic inside a function,

```
function getOverlay() {
  if(!props.show) return null

  return h(...)
}
```

Of course, we can use also `computed`, `ref` as we need as these functions will run whenever the dependency changes.

This is critical one to work with. We have to verify each and every component is working as expected manually (since we don't have tests for all components) after moving all logic inside setup. At this moment, I am sorry I cannot work on this as it might require a lot of time.

jclijmans:
Reading the docs, I think your assumption is not correct. Please note, this is not meant to be a harsh comment, we both just want to understand how things work.

According to my interpretation of the docs and my testing (and again I could be completely wrong ...)

- Returning a render function from setup is the same as defining it separately in the options API
The only difference is that you have access to the reactive state and the parameters given to setup (props, slots, emit, ...)
- Returning from setup means that you define the render function. It doesn't mean that it will be run only once.
- The render function does not make a new component each time it is run, it renders the component in html. The virtual dom compares the newly created html and with make updates to the dom where needed.

To test this, I updated the Overlay component to be returned from the setup and after changing all references to "this" and make sure to use .value when accessing refs or computed, everything worked like it used to. I did not have to wrap the "if props.show" in a getOverlay function.

The original problem with the FormGroup id being recreated is that the id is calculated within the render function. Meaning that with each render, a new id was calculated. The only logic within a render function should be to render the html, so calculating the id within render was definitely wrong.

I'll update the components that user render function so that the render function is returned from the setup
There are not that many components that use render, most use templates
- BIcon, BOverlay, BPagination, BTransition are the only components using the render from options api
- BProgressbar, Btable already return render from setup

axelthat2:
Hey that's okay. We all learn new things everyday. Maybe my comment on saying that component would be created every time was false. Maybe I mixed things up with the issue I was facing in past. I apologize for that.

Although I am surprised the overlay issue worked for you. Maybe I did something wrong it didn't work for me at all.

axelthat2:
Why doesn't it work for me lol?

https://user-images.githubusercontent.com/91611516/139668992-3050d2f5-e5e8-4945-b258-104b4204d75a.mov

This is the code. I just moved render logic to setup and replaced computed/refs with `.value`.
```
<script lang="ts">
import ColorVariant from '../../types/ColorVariant'
import {toFloat} from '../../utils/number'
import {computed, defineComponent, h, PropType, resolveComponent} from 'vue'
import {normalizeSlot} from '../../utils/normalize-slot'
import BTransition from '../BTransition/BTransition.vue'

const POSITION_COVER = {top: 0, left: 0, bottom: 0, right: 0}
const SLOT_NAME_DEFAULT = 'default'
const SLOT_NAME_OVERLAY = 'overlay'

interface ISlotScope {
  spinnerType: string | null
  spinnerVariant: string | null
  spinnerSmall: boolean
}

  name: 'BOverlay',
  components: {BTransition},
  props: {
    bgColor: {type: String, required: false},
    blur: {type: String, default: '2px'},
    fixed: {type: Boolean, default: false},
    noCenter: {type: Boolean, default: false},
    noFade: {type: Boolean, default: false},
    // If `true, does not render the default slot
    // and switches to absolute positioning
    noWrap: {type: Boolean, default: false},
    opacity: {
      type: [Number, String],
      default: 0.85,
      validator: (value: number | string) => {
        const number = toFloat(value, 0)
        return number >= 0 && number <= 1
      },
    },
    overlayTag: {type: String, default: 'div'},
    rounded: {type: [Boolean, String], default: false},
    show: {type: Boolean, default: false},
    spinnerSmall: {type: Boolean, default: false},
    spinnerType: {type: String, default: 'border'},
    spinnerVariant: {type: String, required: false},
    variant: {type: String as PropType<ColorVariant>, default: 'light'},
    wrapTag: {type: String, default: 'div'},
    zIndex: {type: [Number, String], default: 10},
  },
  emits: ['click', 'hidden', 'shown'],
  setup(props, {emit, slots}) {
    const computedRounded = computed(() =>
      props.rounded === true || props.rounded === ''
        ? 'rounded'
        : !props.rounded
        ? ''
        : `rounded-${props.rounded}`
    )

    const computedVariant = computed(() =>
      props.variant && !props.bgColor ? `bg-${props.variant}` : ''
    )

    const computedSlotScope = computed(() => ({
      spinnerType: props.spinnerType || null,
      spinnerVariant: props.spinnerVariant || null,
      spinnerSmall: props.spinnerSmall,
    }))

    const defaultOverlayFn = (scope: ISlotScope) =>
      h(resolveComponent('BSpinner'), {
        type: scope.spinnerType,
        variant: scope.spinnerVariant,
        small: scope.spinnerSmall,
      })

    let $overlay: any = ''

    if (props.show) {
      const $background = h('div', {
        class: ['position-absolute', computedVariant.value, computedRounded.value],
        style: {
          ...POSITION_COVER,
          opacity: props.opacity,
          backgroundColor: props.bgColor || null,
          backdropFilter: props.blur ? `blur(${props.blur})` : null,
        },
      })

      const $content = h(
        'div',
        {
          class: 'position-absolute',
          style: props.noCenter
            ? {...POSITION_COVER}
            : {top: '50%', left: '50%', transform: 'translateX(-50%) translateY(-50%)'},
        },
        normalizeSlot(SLOT_NAME_OVERLAY, computedSlotScope.value, slots) ||
          defaultOverlayFn(computedSlotScope.value) ||
          ''
      )

      $overlay = h(
        props.overlayTag,
        {
          class: [
            'b-overlay',
            {
              'position-absolute': !props.noWrap || (props.noWrap && !props.fixed),
              'position-fixed': props.noWrap && props.fixed,
            },
          ],
          style: {
            ...POSITION_COVER,
            zIndex: props.zIndex || 10,
          },
          onClick: (event: MouseEvent) => emit('click', event),
          key: 'overlay',
        },
        [$background, $content]
      )
    }

    const getOverlayTransition = () =>
      h(
        BTransition,
        {
          noFade: props.noFade,
          name: 'fade',
          appear: true,
          onAfterEnter: () => emit('shown'),
          onAfterLeave: () => emit('hidden'),
        },
        {default: () => $overlay}
      )

    if (props.noWrap) return getOverlayTransition()

    const wrapper = h(
      props.wrapTag,
      {
        'class': ['b-overlay-wrap position-relative'],
        'aria-busy': props.show ? 'true' : null,
      },
      [h('span', normalizeSlot(SLOT_NAME_DEFAULT, {}, slots)), getOverlayTransition()]
    )
    return () => wrapper
  },
})
</script>
```

What am I missing?

jclijmans:
Your code indeed does not work.

If you look at my code in PR #206, you see that I create the helper render variable in the main render function

so instead of
```
const defaultOverlayFn = (scope: ISlotScope) =>
      h(resolveComponent('BSpinner'), {
        type: scope.spinnerType,
        variant: scope.spinnerVariant,
        small: scope.spinnerSmall,
      })
...
return () => wrapper

```
I do

```
return () => {
    const defaultOverlayFn = (scope: ISlotScope) =>
        h(resolveComponent('BSpinner'), {
          type: scope.spinnerType,
          variant: scope.spinnerVariant,
          small: scope.spinnerSmall,
        })
   ....
   return wrapper
}

```
You're code gives warnings in the console
[Vue warn]: setup() should not return VNodes directly - return a render function instead. 

So you are basically returning the VNodes directly instead of a function that creates VNodes

axelthat2:
I get no such error in console and I do return a render function `() => wrapper`.

I think your solution worked because you placed the logic inside the render function. So as you said above instead of doing the following,

```
const defaultOverlayFn = (scope: ISlotScope) =>
      h(resolveComponent('BSpinner'), {
        type: scope.spinnerType,
        variant: scope.spinnerVariant,
        small: scope.spinnerSmall,
      })
...
return () => wrapper
```

You did,

```
return () => {
    const defaultOverlayFn = (scope: ISlotScope) =>
        h(resolveComponent('BSpinner'), {
          type: scope.spinnerType,
          variant: scope.spinnerVariant,
          small: scope.spinnerSmall,
        })
   ....
   return wrapper
}
```


Which means the function will be called everytime component needs to re-render which in this case is `props.show`.

PR #205, opened 2021-11-01
Component Reference Plugin

aceofwings:
Hey guys, 

I've been working on extending the vue press documentation by porting over the original BoostrapVue's docs to a plugin. This pull request serves as just an example of the current structure, feed back is much appreciated.

In the BoostrapVue a package.json was accompanied with each component describing traits. The component definition and package.json generated the component reference tables. 

Now there is a folder in docs where component references match up with there corresponding mark down pages. By using `ComponentReference` within a components doc page, it will render the necessary documentation.

### DIFF
--- a/docs/.vuepress/PluginComponentReference/client/clientAppEnhance.ts
+++ b/docs/.vuepress/PluginComponentReference/client/clientAppEnhance.ts
@@ -1,0 +1,7 @@
+import {h} from 'vue'
+import {defineClientAppEnhance} from '@vuepress/client'
+import ReferenceSection from './components/reference-section'
+export default defineClientAppEnhance(({app}) => {
+  // wrap the `<ComponentReference />` component with plugin options
+  app.component('ComponentReference', (props) => h(ReferenceSection, {}))
+})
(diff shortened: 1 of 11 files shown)


jclijmans:
Cool!!!

A couple of questions :
- I don't see the default values in the json file, does this plugin get that from the component itself?
- In bootstrap-vue, the json files are in the same folder as the components themselfs. In this PR you place them in a separate  folder. Is it possible to retrieve them from the source component folder? @cdmoro : Don't know what the preferred place would be.

I understand that the supplied Pagination.json  is a proof that it works, but :
- the name should be @bootstrap-vue-3/pagination
- we should not use the versions because the versions don't match our version and everything will be supplied in v1.0.0

aceofwings:
So to answer some of your questions. 
 
- The default values are pulled and enumerated from the component definition itself
- I was thinking the same thing and is totally doable as I believe I can pull the absolute and relative component definitions base paths from some vue utilities or properties.  If we all agree I can move to resolving component references in the component folder and  either continue using `[component-name].json` or define a `package.json`.  

I agree with the additional points as and will redefine the .json files as needed. 

Ill be working on this for the coming weeks in hopes to improve the documentation experience.

jclijmans:
Great, thx 
Looking forward to using it.
I'll wait with further documentation until this is in as not have to rework it afterwards

PR #222, opened 2021-11-09
add options.remove to install

xvaara:
add option to remove components and directives from install.

for example: `{remove: ['BIcon', 'BAvatar']}` 😉


cdmoro:
I appreciate your work but I don't really see the use of this. Because It doesn't have a real impact on the bundle size.
Please, let me know what you think.

xvaara:
Yeah, I'm using this when importing directly from bootstrap-vue-3/src/BootstrapVue, I can just import the Components I use also. I've just seen this convention in some vue libraries.

PR #233, opened 2021-11-21
Feature/cr

aceofwings:
Still a work in progress but will speed up documentation and remove errors such as missing prop declarations.

As JC suggest to move the component reference declarations to each individual component. Sadly right now with the use of vue press, I have not found a clear way to extend the page data with the component declarations.


mariusa:
thanks!

aceofwings:
Will be producing test cases for this plugin. More to come!

PR #235, opened 2021-11-23
Issue#124 incorrect values when using null value

kwiksilver3441:
When values in the :value of the option element contain null or undefined, it doesn't create value="", instead, it ignores the attribute. Thus causing incorrect reading of null values when switching to a value with null, and mounting with null as default value. Simply null-checking the value prop and assigning an empty string safely adds in an empty value attribute for correct targeting.


mariusa:
thanks!

PR #240, opened 2021-11-26
disabled is boolean, so it can be true also

xvaara:


cdmoro:
I'm a bit confused here. Since the type of `disabled` property is `boolean`, wouldn't be better put like that?

``` javascript
disabled: tab.props.disabled,
```

When `tab.props.disabled` is set to `true` the tab will be disabled.

xvaara:
I'm not sure what the original idea of checking against empty string is, I think when accessing children from `Tabs` the `<Tab disabled />` component returns an empty string. That is just a guess since I didn't write the original code. Just needed to programatically  change the boolean.

xvaara:
Are you gonna test or fix the tabs? I'd like to stop running my own fork, since this project now has releases.

aceofwings:
I am with @cdmoro on this. I am assuming getTabs are returning Tab Vnode instances. Ill have to review the documentation but it does not make sense to check for empty string if the default value is supposedly a Boolean false.

xvaara:
> I'm a bit confused here. Since the type of `disabled` property is `boolean`, wouldn't be better put like that?
> 
> ```js
> disabled: tab.props.disabled,
> ```
> 
> When `tab.props.disabled` is set to `true` the tab will be disabled.

I tested that. With that change none of the `<tab disabled>` tabs are disabled.  It comes through as empty string:
Screenshot 2022-02-23 at 15 54 40

this is with a boolean passed to it as variable:
Screenshot 2022-02-23 at 15 56 32

xvaara:
Yeah, so it is as intended. As you can se from the linked issue. 
So I think that this is the only way to do it, if we want to use vnodes.

cdmoro:
I see it, this boolean-string situation happens because we use vnodes, right? In that case the only solution is merge your code...

xvaara:
Yes.

PR #279, opened 2022-02-16
Feature/toast-fix

aceofwings:
-  Made some quick fixes to the layout styling for toasters. 
- Add Variant to ToastOptions
- Add Variant Helpers

Everything should now appear correctly.

### DIFF
--- a/docs/components/Toast.md
+++ b/docs/components/Toast.md
@@ -47,12 +47,14 @@ Use `toast.show()` method to generate your toasts. The method accepts the follow
 ></b-container>
 <b-button class="" @click="show1">Show Toast</b-button>
 <b-button class="ms-2" @click="show2">Show Toast with Body</b-button>
-<script>
+
+<script lang="ts" setup>
   import {useToast} from 'BootstrapVue/ToastPlugin'
   let toast = useToast()
   let show1 = () => {
     toast.show({title: 'example title'})
   }
+
   let show2 = () => {
     toast.show({title: 'example title', body: 'This is a toast'})
   }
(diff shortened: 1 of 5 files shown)


docs/components/Toast.md
```diff
@@ -66,17 +68,74 @@ Use `toast.show()` method to generate your toasts. The method accepts the follow
...
+```html
+<b-container
+  :toast="{root: true}"
+  fluid="sm"
+  position="position-fixed"
+  style="top:50px; left : -200px"
+></b-container>
+<b-card>
+  <b-button class="" variant="danger" @click="variantshow1">Show Danger</b-button>
+  <b-button class="ms-2" variant="Info" @click="variantshow2">Show Info</b-button>
```
mariusa, line 86: Please use same letter case: "Info" -> "info"

mariusa:
Hi @aceofwings  
I tried to use it in an app, but this doesn't work:
`import BToastPlugin from 'bootstrap-vue-3/plugins/BToast'` 

1. Would you please update the usage doc with correct syntax for 3rd party apps?

2. `Toast Components can be used standalone or part of the toast plugin.` What's the difference? What is standalone?

docs/components/Toast.md
mariusa: I changed this

aceofwings:
1. Sorry about that, Due to how boostrap-vue-3 can't be directly imported into the docs I opted for  a different way for fetching the plugin instance. The correct way is now `import {useToast} from './components/BToast/plugin'
2. You can currently use the BToast component and provide a true/false value to the modal to hide and show.
3. If no title is set.... it still show a title. This can easily be updated to behave that way. 

Apologies for some of the missing stuff.

aceofwings:
Looks like the way the library is done. It should resolve to 'boostrap-vue-3/components/ToastPlugin'. Ill test it out.

Ill look at refactoring for plugins so that it can be included as `bootstrap-vue-3/ToastPlugin'

mariusa:
How would the usage be in each case and which one do you recommend?

> I can default a global property in prime vue

You mean 'like prime vue' ? Now I use prime vue like this:

main.ts
```js
import Toast from "primevue/toast"
import ToastService from "primevue/toastservice"
...
    app.use(ToastService)
    app.use(router)
    app.component("Toast", Toast)

```

app.vue
```js
<template>
    <div>
        <router-view />
        <Toast position="bottom-center" />
    </div>
</template>

<script setup>
import { useToast } from 'primevue/usetoast'
const toast = useToast()

toast.add({ severity: 'info', summary: 'Message 1', detail: 'Message 1 Content', life: 3000 })

```

`toast.add()` is called only in app.vue by an event listener. All other modules just emit the event to show toast with some properties.

PR #293, opened 2022-02-28
Busy tables

AlteraWeb:


cdmoro:
Hello there, did you test toggling `busy` prop (with a checkbox, for example)? I'm not sure if `tBody` will change its value.

AlteraWeb:
> Hello there, did you test toggling `busy` prop (with a checkbox, for example)? I'm not sure if `tBody` will change its value.

Hello, I test it with Boolean variable, I change value by button.

AlteraWeb:
Снимок экрана 2022-02-28 в 14 22 34

cdmoro:
Good. A few things, according to [BootstrapVue table docs](https://bootstrap-vue.org/docs/components/table#table-busy-state):

- We should implement the `aria-busy` attribute
- We should add this CSS style:
``` css
table.b-table[aria-busy='true'] {
  opacity: 0.6;
}
```
- It's not mandatory the use of the `table-busy` slot. Regarding your code, the table only shows the busy state if this slot is passed.

cdmoro:
Hello there, it's me again. According to the BootstrapVue table docs, the `table-busy` slot doesn't replace the table, it shows the table but with this code inside `tbody`:

``` html
<tr role="row" class="b-table-busy-slot">
    <td colspan="3" role="cell">
        <!-- table-busy slot content -->
    </td>
</tr>
```

Could you please make those changes before merging? Thanks!

cdmoro:
Hello?

VividLemon:
After many months I'm going to close out this PR. I think `busy` is added and used in the current releases.

PR #316, opened 2022-03-13
Arrow fix dopdown

punga78:


cdmoro:
I think you use another branch to fork the project. You should make the changes on `main` branch, instead of `component/form-tags` branch. I think the best solution is to close this PR and start a new fresh one with the dropdown changes based on `main`.

Issue #120, opened 2021-09-30
Cannot use $bvModal.hide or show


loksoon:
Uncaught TypeError: _ctx.$bvModal is undefined
![Screenshot 2021-09-30 at 19 27 53](https://user-images.githubusercontent.com/43371272/135447059-42bc6604-83bb-4e9a-a9cc-a1ea526cef4d.png)
ed

cdmoro:
That's right! we haven't implemented that functionality yet. At the beginning of this project, I implemented the very first version of $bvModal, there is a function called useModal (it isn't documented but you can import it).

Honestly, I don't know which is the best strategy. If we should use [global properties](https://v3.vuejs.org/guide/migration/global-api.html#vue-prototype-replaced-by-config-globalproperties) or create a composable.

loksoon:
Thanks for a commented

loksoon:
Em i cannot find the useModal function inside your code

Don-Maxon:
when will Modal be added ?

Kenvdb:
> when will Modal be added ?

Yeah same here, interested to know.

xvaara:
There is modal, just that $bvModal isn't the way you do things in vue3, use a variable or create a reactive state.

L422Y:
The problem here is that if you use `v-b-modal`, there's no good / clean way to programmatically close modals without interacting with the DOM.

```
    <b-button v-b-modal.add-new-modal>Add New</b-button>              
```

asika32764:
For someone who is looking for closing BModal programmatically in child components, this works for me:

```ts
import { type ComponentInternalInstance, getCurrentInstance } from 'vue';

export function useClosestModal() {
  const instance = getCurrentInstance();

  if (!instance) {
    return null;
  }

  const modalComponent = findBModal(instance);

  if (!modalComponent) {
    return null;
  }

  return {
    show() {
      return modalComponent.exposed?.show()
    },
    hide(trigger = '') {
      return modalComponent.exposed?.hide(trigger)
    }
  };
}

function findBModal(component: ComponentInternalInstance) {
  if (!component.parent) {
    return null;
  }

  if (component.parent.type.__name === 'BModal') {
    return component.parent;
  }

  return findBModal(component.parent);
}

```

usage:

```ts
const currentModal = useClosestModal()!;

currentModal.hide();
```

VividLemon:
> For someone who is looking for closing BModal programmatically in child components, this works for me:
> 
[...]

PR #329, opened 2022-03-21
Implementation of $bvModal

parafeu:
Related to #120 

Hi ! This is my first contribution of an open source project, I hope it goes well!
I've made this implementation using mitt.js as global event bus. 
This could probably be replaced by a native solution, but this library is realy tiny and easy to use.

### DIFF
--- a/src/components/BModal.vue
+++ b/src/components/BModal.vue
@@ -5,7 +5,9 @@
         <div class="modal-content" :class="contentClass">
           <div v-if="!hideHeader" class="modal-header" :class="computedHeaderClasses">
             <component :is="titleTag" class="modal-title" :class="computedTitleClasses">
-              <slot name="title">{{ title }}</slot>
+              <slot name="title">
+                {{ title }}
+              </slot>
             </component>
             <button
               v-if="!hideHeaderClose"
@@ -32,8 +34,9 @@
                 :size="buttonSize"
                 :variant="cancelVariant"
                 @click="$emit('cancel')"
-                >{{ cancelTitle }}</b-button
               >
+                {{ cancelTitle }}
+              </b-button>
               <b-button
                 type="button"
                 class="btn btn-primary"
@@ -42,8 +45,9 @@
                 :size="buttonSize"
                 :variant="okVariant"
                 @click="$emit('ok')"
-                >{{ okTitle }}</b-button
               >
+                {{ okTitle }}
+              </b-button>
             </slot>
           </div>
         </div>
--- a/src/components/BModal.vue
+++ b/src/components/BModal.vue
@@ -5,9 +5,7 @@
         <div class="modal-content" :class="contentClass">
           <div v-if="!hideHeader" class="modal-header" :class="computedHeaderClasses">
             <component :is="titleTag" class="modal-title" :class="computedTitleClasses">
-              <slot name="title">
-                {{ title }}
-              </slot>
+              <slot name="title">{{ title }}</slot>
             </component>
             <button
               v-if="!hideHeaderClose"
@@ -34,9 +32,8 @@
                 :size="buttonSize"
                 :variant="cancelVariant"
                 @click="$emit('cancel')"
+                >{{ cancelTitle }}</b-button
               >
-                {{ cancelTitle }}
-              </b-button>
               <b-button
                 type="button"
                 class="btn btn-primary"
@@ -45,9 +42,8 @@
                 :size="buttonSize"
                 :variant="okVariant"
                 @click="$emit('ok')"
+                >{{ okTitle }}</b-button
               >
-                {{ okTitle }}
-              </b-button>
             </slot>
           </div>
         </div>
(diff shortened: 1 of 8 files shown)


src/components/BModal.vue
```diff
@@ -34,9 +32,8 @@
                 :size="buttonSize"
                 :variant="cancelVariant"
                 @click="$emit('cancel')"
+                >{{ cancelTitle }}</b-button
```
mariusa, line 35: pls fix formatting: ```</b-button``` and ```>``` should be on the same tag, not separate lines

src/components/BModal.vue
```diff
@@ -45,9 +42,8 @@
                 :size="buttonSize"
                 :variant="okVariant"
                 @click="$emit('ok')"
+                >{{ okTitle }}</b-button
```
mariusa, line 45: pls fix formatting: ```</b-button``` and ```>``` should be on the same tag, not separate lines

src/components/BModal.vue
mariusa: Did anything change here? If not, please make it same as previous code, so there's no diff.

xvaara:
We could just use a reactive object and watch that, no mitt needed.

But, what is the function of this feature? Or is it just for bootstrapVue compatibility? The why would you need to use a global function instead of the provided ways to show modals?

xvaara:
If you need to use bus, I think it would be better to use https://vueuse.org/core/useeventbus/ as I think we'll use that library at some point.

and @parafeu  you should turn the listener off on unmount, so that they aren't left behind.

parafeu:
> We could just use a reactive object and watch that, no mitt needed.
> 
> But, what is the function of this feature? Or is it just for bootstrapVue compatibility? And why would you need to use a global function instead of the provided ways to show modals?

The purpose of my pull request was to cover issue #120 . The advantage of a global function is that it is available in templates no matter which API is used (option or composition). If needed I can make an alternative with a composable. 



> If you need to use bus, I think it would be better to use https://vueuse.org/core/useeventbus/ as I think we'll use that library at some point.
> 
> and @parafeu you should turn the listener off on unmount, so that they aren't left behind.

Totally forgot the basics here, my bad. I'll make a new pull request shortly using vueuse instead of mitt.

mariusa:
A note on `globalProperties`, quoting Evan You:

> [globalProperties] … is essentially a legacy escape hatch and not for setup function. You should just import things directly or use provide/inject in setup functions.

But I'm not sure how to use `provide/inject` in a library. In my app, `app.vue` has `provide()`, all other `.vue` have `inject()` at top. Here `app.vue` is used for demo, being a library, not an actual app.

parafeu:
From Vue3 plugin documentation
> There is no strictly defined scope for a plugin, but common scenarios where plugins are useful include:
> - Make a resource [injectable](https://vuejs.org/guide/components/provide-inject.html) throughout the app by calling [app.provide()](https://vuejs.org/api/application.html#app-provide).

xvaara:
I'm not a fan of libraries adding to the globalProperties. You can implement it in your own code if you need.

Again, I'm not sure why you can't just use a state for showing modals? Why is this feature even needed? This really fells like it's only for legacy versions of code from vue1 days.

parafeu:
I implemented this feature because it was in the roadmap for 1.0.0. It seems indeed that it is mainly for backward compatibility with Bootstrap Vue 2

mariusa:
In this case, it seems best we revert the PR. Do you agree?
Sorry I merged without investigating more.

Marius

parafeu:
As you wish, I dont mind.

But maybe we should discuss about an alternative (using composable ?), it might be useful to control modal visibility without having access to its state.

mariusa:
> it might be useful to control modal visibility without having access to its state.

Would you please start a Discussion? And detail the use case / need from an app developer. Thanks!

Micene09:
In my company we are using a lot the old `$bvModal`, just because it was the only alternative to have a functional modal, that can be launched from any point of the code-base, not just from a component/view.

Personally, I think that at least we should have a composable alternative, this way we can bring almost the same functionality but with the new "Vue 3 fashion".

I agree with @xvaara about globalProperties, but maybe the whole [project](https://github.com/cdmoro/bootstrap-vue-3/projects/1) should be defined better with more details, even just to help the (growing) community here...I would help but I don't know how.
Do you have a Discord channel? How can we start a discussion on how to improve issue tracking and roadmap?

mariusa:
How about having a modal instance in app.vue, and using an event bus (mitt)
to emit hide/show events from anywhere?
app.vue listens to these and hides/shows with the received content .

For discussions, there's the Discussions section here on Github. Spreading
even thinner on other apps won't benefit.

Micene09:
> How about having a modal instance in app.vue, and using an event bus (mitt)
to emit hide/show events from anywhere?
app.vue listens to these and hides/shows with the received content .

Not atomic at all...
I'm taking about a "composable", spawning a modal from anywhere.
Imagine the equivalent of the $bvModal, returning a Promise that gets resolved once the modal was closed.
In so many business logics, outside a view and components logics, during a very complex Store (async) Action may happen to give a feedback using an info/warning (or whatever) modal with a message.

>For discussions, there's the Discussions section here on Github. Spreading
even thinner on other apps won't benefit.

Sorry, my bad.
I'm ready to start a new discussion or just move this thread the one creates by @parafeu.

PR #351, opened 2022-04-07
feat: generates *.d.ts files from .ts or .vue source

messenjer:
Hello, 

I saw an issue https://github.com/cdmoro/bootstrap-vue-3/issues/326 about d.ts file missing.

There is a nice plugin https://github.com/qmhc/vite-plugin-dts that generates declaration files (*.d.ts) from .ts or .vue source files when using vite in [library mode](https://vitejs.dev/guide/build.html#library-mode)

It's very simple, as vite developer experience.

May be we can use it.

Screenshot 2022-04-07 at 16 07 09

Mathieu

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,6 +70,7 @@
     "ts-jest": "^27.0.4",
     "typescript": "4.x.x",
     "vite": "2.x.x",
+    "vite-plugin-dts": "^1.0.5",
     "vue": "^3.2.31",
     "vue-loader": "17.x.x",
     "vue-router": "^4.0.12",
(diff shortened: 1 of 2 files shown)


mariusa:
Thanks! `BootstrapVue.d.ts` file is still empty though, I assume because we don't use vite in library mode?
Would you send a PR for that too?

mariusa:
Although I see this
```
[vite:dts] Start generate declaration files...
[vite:dts] Declaration files built in 6897ms.
```
I don't know where are the generated declarations

PR #354, opened 2022-04-11
Fix toast examples

sfe-efficy:


mariusa:
Thanks!

sfe-efficy:
Hi again
It's still pending to update the examples in 
https://cdmoro.github.io/bootstrap-vue-3/components/Toast.html#toasts-on-demand

And another thing, the example in that page for toast.show is not working.

I saw that toast in that example is a HTMLElement 

image

image

aceofwings:
the plugin was removed from the page due to compilation errors with ssr. Toasts will not work until a work around will come into play sadly...

PR #356, opened 2022-04-13
Objects as values in options for BFormSelect

ABraconnier:
## Current bug

Bootstrap-vue allows for objects to be passed for options values in BFormSelect. So, for example, according to this example from their doc:
```
        options: [
              { value: { C: '3PO' }, text: 'Option with object value' },
              { value: { R: '2D2' }, text: 'Another option with object value' }
        ]
```
The current implementation is buggy when trying to add objects as value. If two values with objects are provided, no matter what v-model will be, it will select the first of those two as long as it is an object too. This is because since we can't pass objects into actual HTML, values will not be the actual object but [object Object]. 

## How to reproduce

Have more than one option with object as value. You will see that when selecting one or another, the select is buggy

## Fix

I stringify to JSON any values that are passed to HTML template. OnChange, I emit the parsed value from JSON to go back to its original state. I'm not sure if it's ideal, but it's what I do as a fix for our current project. Any suggestion on how to better solve this is welcome.

### DIFF
--- a/src/components/BFormSelect/BFormSelect.vue
+++ b/src/components/BFormSelect/BFormSelect.vue
@@ -12,7 +12,7 @@
     :aria-required="required ? 'true' : null"
     :aria-invalid="computedAriaInvalid"
     v-bind="$attrs"
-    :value="modelValue"
+    :value="JSON.stringify(modelValue)"
     @change="onChange($event)"
   >
     <slot name="first" />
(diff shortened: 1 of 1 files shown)


aceofwings:
JSON encoding/ decoding internally does not really make sense. Let me see if there is a better way.

xvaara:
see #363 , objects were a problem else where also.

PR #363, opened 2022-04-21
Fix select/checkbox/radio with object

xvaara:
Using objects in select/checkbox/radio was completely broken. It always matched the first object. I removed some weird hacky stuff and used vue:s native v-model binding for those inputs.

Can you please release this and ssr fix soon, so I don't have to run a fork in production.


aceofwings:
Good stuff. This will most likely close out the pull request submitted. #356 

Will review when I get the chance

mariusa:
Thanks! Will also make a release.

PR #368, opened 2022-04-26
feat(BNavItemDropdown): pass slots to the child BDropdown

messenjer:
Hello,

When we use `BNavItemDropdown` it's not possible to pass slot to the `BDropdown`.

I added the possibility to pass slots through from `BNavItemDropdown` to `BDropdown`.

Regards

Mathieu

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -1659,22 +1659,40 @@
       >
     </div>
 
-    <!-- <b-nav pills>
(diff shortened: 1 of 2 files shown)


mariusa:
Thanks! Would be useful to add a mention in docs too (no need to re-add the sample code already in App.vue)

PR #380, opened 2022-05-05
fix: remove variant classes from OK/Cancel buttons

edwh:
First PR on this, so apologies if I miss something.

`b-modal` has `ok-variant` and `cancel-variant` to set the button variants.  These aren't working because we have explicit `btn-primary`/`btn-secondary` classes in there.  That means you get both classes inserted, and the primary wins.  We don't need these classes (I think) because `variant` will already ensure the correct one is inserted.

### DIFF
--- a/src/components/BModal.vue
+++ b/src/components/BModal.vue
@@ -28,7 +28,7 @@
               <b-button
                 v-if="!okOnly"
                 type="button"
-                class="btn btn-secondary"
+                class="btn"
                 data-bs-dismiss="modal"
                 :disabled="disableCancel"
                 :size="buttonSize"
@@ -39,7 +39,7 @@
               </b-button>
               <b-button
                 type="button"
-                class="btn btn-primary"
+                class="btn"
                 data-bs-dismiss="modal"
                 :disabled="disableOk"
                 :size="buttonSize"


MrTanoshii:
I don't think this should get merged.

The current implement has default colors as `btn-primary` and `btn-secondary` but you can actually override the colors by specifying any color you want; note that `okVariant` and `cancelVariant` are props with default only values.

I think it's best to allow the end user to theme as required rather than force a specific color.

### Feedback
PR wise (as if it were to get merged):
I think it'd make sense to remove all variant related code since that is the purpose of the PR.
That is, Lines 35, 46, 79 & 108

edwh:
Thanks for the comment. 

If you try to do that, say to secondary, you end up with a button which has both the btn-primary and btn-secondary classes set on the same element. Primary is there because it's hardcoded, and secondary gets added via the variant prop. 

In that case the btn-primary class wins out, and your attempt to override fails.

The default color values are already applied because of the default values of the variant prop. We don't need the classes to be hardcoded present too.

If that's not clear I can set up a repro so you can see it.

mariusa:
Agreed to remove the hardcoded variants.

> The current implement has default colors as btn-primary and btn-secondary

These could be props default values, not hardcoded generated html.

MrTanoshii:
Ah my bad, it's best to remove those classes

I pass color props and it overrides on my side.
For e.g.
```html
<b-modal
        id="exampleModal"
        v-model="showModal"
        size="xl"
        title="Modal title"
        okVariant="danger"
        cancelVariant="success"
        fade
        >...
</b-modal>
```

That being said, you are right in removing those classes there!
Edited my original reply to reflect that.

edwh:
@MrTanoshii Phew!  It would be a bit embarrassing if my first PR on this project was dumb.

Now I just need to work out how to apply a local build of this package into my own project, so that I can make more significant fixes.  I haven't done that before, so I need to find out the obvious way of doing it that everyone knows...

MrTanoshii:
you can use

Github Repo
```json
  "dependencies": {
    "bootstrap-vue-3": "USER/REPO",
  },
```

Local repo
```json
  "dependencies": {
    "bootstrap-vue-3": "file:PATH/TO/DIR",
  },
```

See https://docs.npmjs.com/cli/v8/configuring-npm/package-json#local-paths

PR #393, opened 2022-05-24
fix(BToast): vue has no exported member named 'Reactive'

messenjer:


messenjer:
@mariusa It's strange the build failed...

aceofwings:
Went through Vue/Core, I see no Reactive type. Core never really complained about the type not being there. Think I will push a fix and remove `Reactive`

Issue #397, opened 2022-05-28
Popover component is missing the offet parameter


edwh:
https://getbootstrap.com/docs/5.0/components/popovers/ documents this, but https://github.com/cdmoro/bootstrap-vue-3/blob/main/src/components/BPopover.vue doesn't seem to include it.

PR #398, opened 2022-06-01
Issue #397 popover doesn't contain offset

VividLemon:
Popover doesn't contain offset prop from either bootstrap vue or bootstrap. This change would be considered breaking from migrating from boostrap vue to bootstrap vue 3. Offset on bootstrap vue types are number | string. In line with bootstrap 5, offset now is [number, number] | string | function.


kwiksilver3441:
Adding "0px" doesn't actually do anything. I guess popover doesn't parse for px/rem/etc. I can resubmit if it really matters but it doesn't change anything.

mariusa:
If it doesn't do anything, would you please edit/remove that? Maybe add this comment in code. Thanks!

PR #404, opened 2022-06-07
Fixed issues revolving around broken packages.

kwiksilver3441:
The prettier plugin hasn't worked for awhile. A fix in the eslintrc file to include prettier as a plugin, and as a rule fixes prettier not working.

tsconfig file was changed to temporarily fix these two issues revolving around volar plugin: https://github.com/johnsoncodehk/volar/issues/1405 && https://github.com/johnsoncodehk/volar/discussions/592 

The prepare script was causing too many issues. It was preventing packages from being installed and hurting the development process. It should be added later when not so many breaking "vite build" options cause npm install to not install the required packages to actually develop. 

vue-tsc was added for the build command. I'm not sure why this was not already included. But to my understanding it's required for a Vite app using TS to build. In relation to https://github.com/cdmoro/bootstrap-vue-3/issues/403 it fixes the TS compilation issue, however, still running build causes _actual_ issues to appear.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -12,10 +12,12 @@ module.exports = {
     '@vue/typescript/recommended',
     'prettier',
   ],
+  plugins: ['prettier'],
   parserOptions: {
     ecmaVersion: 2021,
   },
   rules: {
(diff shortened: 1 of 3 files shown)


mariusa:
Thanks!

PR #407, opened 2022-06-09
Types fixes

kwiksilver3441:
The major change here is that the types dir index file now imports all other types, and correspondingly exports them. Files that use an types/* import, have been changed to importing only the index file. Some types files were actually .d.ts files, but not named correctly. And safe-types file was renamed to safeTypes to be consistent (some other files follow hyphen case...). This also includes a few linting issues that made it passed preexisting commits. Specifically those using non primitive types (ie Symbol vs symbol, Boolean vs boolean). Some files I came across while fixing the import type/* contained some errors before/after.


aceofwings (commented):
Real quick question 

in some component files you import a type via a relative path 
`../../types`

in other d.ts you import a 
`src/types`

just wondering why the difference.

kwiksilver3441:
> Real quick question
> 
> in some component files you import a type via a relative path `../../types`
> 
> in other d.ts you import a `src/types`
> 
> just wondering why the difference?

Some were from me specifically writing them, while others were auto generated (ctrl + space) intellisense. I'd prefer just using src/types. But ultimately they didn't make a difference. Once I'm done with all the interfaces (I think there's only two), I can go around and replace globally any ../types imports in favor of src/ . But I figure someone eventually will import a type as ../ so ¯\_(ツ)_/¯

PR #408, opened 2022-06-09
Component type files fixed up

kwiksilver3441:
Component types files are still a work in progress. This contains no changes to anything other than the types/components dir. So, should not interfere with anything. This was mainly to properly import non-native types. It's important to note that if https://github.com/cdmoro/bootstrap-vue-3/pull/407 is not merged, that some of the type exports from the fixed index file of types/index.ts will not exist.


kwiksilver3441:
Just a note. Can we use import type ... in Vue files? There should be no reason why not as it will be solved by vue-tsc. I've just noticed that it's never included in any Vue files, while being included in declaration files.

aceofwings:
import type can be used and should be used, certain ts configurations will complain if you import a declaration without specifying `import type`... good catch.

Looks good

kwiksilver3441:
> import type can be used and should be used, certain ts configurations will complain if you import a declaration without specifying `import type`... good catch.
> 
> Looks good

In reference to the other PR, while going around replacing to src/types in the (probably next PR), I'll also make sure any Vue files importing types uses import type

mariusa:
Hmm, there are conflicts now. I'm not sure which is the proper syntax to solve them (main / setup). Would you please have a look?

kwiksilver3441:
> Hmm, there are conflicts now. I'm not sure which is the proper syntax to solve them (main / setup). Would you please have a look?

My bad. Fixed.

kwiksilver3441:
In the latest, and possibly final additions to the component types files, I have added the last two additions, b-row and b-col. Inside these components, they use a function "getBreakpointProps". This is going to be an odd case as those props are defined during runtime through the function. This shouldn't be an issue, as the new type interfaces in src/types/BreakpointProps . However, it makes the function not useful, and, as the function was to reduce some typing, instead, we must declare each property individually. (As I don't know of any way we could implement this type of dynamic behavior in pure TS interface). Subsequently, as rowColsProps (and the equivalent in b-col) are now types, that makes the getClasses function also invalid. At this time, I don't have any ideas on how to make these utility functions work again. The components should be functional if one hardcodes in the values, These two cases are the only components that behave like this. Finally, could someone could review the src/types/BreakpointProps.d.ts file to make certain that those are the correct values that come from the getBreakpointProps function, I have only reviewed their behavior in code and cannot be 100% certain.

mariusa:
> I have only reviewed their behavior in code and would like an extra pair of eyes to be certain.

Looks good to me at a glance. Thanks!

PR #410, opened 2022-06-13
BAccordion & item to script setup

kwiksilver3441:
Due to a previously unknown syntax limitation https://vuejs.org/guide/typescript/composition-api.html#syntax-limitations

 `A reference to an interface or object literal type in **the same file**`

This means that the organization that was hoped to be brought from the src/types/components dir, is less useful. According to Hiws, a member of the original Boostrap-vue project, he heard that there was talks in the Vue discord server stating that importing types from other files may be released in Vue v3.3. For now, importing types from other files is not possible. So, perhaps in a future release of Vue, we may be able to clean up the code -- while also properly being able to import extension types. For now, we are stuck doing it the hard way.


kwiksilver3441:
For some odd reason, I think the original BAccordionItem.vue (the file before the move to it's own dir) file isn't deleting on the PR. Some weirdness going on there. If it doesn't get removed in this. I'll submit another PR with it removed along with the next couple of changes.

kwiksilver3441:
Also, a question, in future components, the build errors that are plaguing the library, they are usually stemmed from trying to assign a html attribute to null. This is allowed in Vue, as null or undefined values are not rendered, but TS comes with the error. If we simply leave them to be undefined, the attributes will not be rendered, this could possibly fix almost all of the "cannot assign null to HTMLelement..." errors. While also keeping them unrendered. Should we assign them to undefined instead?

aceofwings:
looks like it was deleted. I will have to review the error as I am not familiar with the use case. Would you provide an inline example?

kwiksilver3441:
> looks like it was deleted. I will have to review the error as I am not familiar with the use case. Would you provide an inline example?
> 
> looks good otherwise!

An example of our error or how vue doesn't render undefined?

aceofwings:
> An example of our error or how vue doesn't render undefined?

Yeah, its been a while since I last built the library and documentation but never came across the undefined error.

kwiksilver3441:
> > An example of our error or how vue doesn't render undefined?
> 
> Yeah, its been a while since I last built the library and documentation but never came across the undefined error.

In BFormSelect.vue 
```
Type 'string | null' is not assignable to type 'string | undefined'.
  Type 'null' is not assignable to type 'string | undefined'.ts(2322)

runtime-dom.d.ts(794, 3): The expected type comes from property 'form' which is declared here on type 'ElementAttrs<SelectHTMLAttributes>'
```
This is similar to the other issues, however, undefined is an allowed value, which means it will be properly unrendered if we explicitly set it to undefined. This is mainly a TS issue, as if we had some sort of Vue/types then it would be addressed there that null is allowed as an attributes value.

PR #412, opened 2022-06-17
BAvatar/Group, BBadge, BBreadcrumb, BLink script setup and fixes

kwiksilver3441:
Bavatar was converted to script setup syntax. 
Bavatargroup was converted to script setup syntax
Bbadge was not converted but updated to have it's functions be more strongly typed to hopefully prevent possible future issues
BBreadcrumb was converted to script setup syntax, with it's "computedItems" function renamed and return type modified to match what it actually does. Along with eliminating the error that came with BBreadcrumb interface being string | BBreadcrumbObject not assignable to v-bind
BBreadcrumb item contains strongly typed functions, as well as a reoccurring fix of the click event of an HTML element being "MouseEvent" not assignable to "PointerEvent" (We used PointerEvent, but TS said the HTML element was exporting MouseEvent)
BLink contains more strongly typed additions. as well as preferring arrow functions and moving it to keep computed functions together.
src/types files contain minor bug fixes and improvements.

Went through a few changes on this one and discarded some edits, might have gotten confused. Check carefully please


mariusa:
Great, thanks!

PR #414, opened 2022-06-20
BCard's script setup conversion and improvements

kwiksilver3441:
One commit moves all the types/components files to use @/types instead of src/types

The other contains script setup changes. In addition, it fixes some bugs of various "\*TitleTextVariant\*" Props to use TextColorVariant type.


kwiksilver3441:
It should be noted that in various components, they say an Array, type can be used, but doesn't strongly type the Array. I've chosen to use Array<string> since classes seem like they are going to be strings.

PR #417, opened 2022-06-21
Update vue test utils package

kwiksilver3441:
The main reason for the PR was to update package and state some intentions. It also contains some minor fixes in some unused types files. 

I'm going to need to change some tests in the next script setup conversions. BFormCheckbox has contained an error where input v-model cannot be assigned to null. Instead, I opt to use undefined since it's valid and behaves the same. Just the types checks expecting it toBe(null) will always issue out. 

I'm also going to opt to moving the `Array<string>`'s that I have been using to `Array<unknown>` (and retroactively fix the previous ones) as I have encountered some issues with them, and they're usually unknown to _me_. The unknowns will provide a better search ability to make them more explicit in the future as they were basically not strongly typed before anyways.

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,7 +57,7 @@
     "@vitejs/plugin-vue": "2.x.x",
     "@vue/compiler-sfc": "3.x.x",
     "@vue/eslint-config-typescript": "10.x.x",
-    "@vue/test-utils": "^2.0.0-rc.14",
+    "@vue/test-utils": "^2.0.0-rc.18",
     "eslint": "8.x.x",
     "eslint-config-prettier": "8.x.x",
     "eslint-plugin-prettier": "^4.0.0",
--- a/src/types/components/BFormCheckbox/BFormCheckbox.d.ts
+++ b/src/types/components/BFormCheckbox/BFormCheckbox.d.ts
@@ -1,5 +1,4 @@
-import type {InputSize} from '@/types'
-
+import type {ButtonVariant, InputSize} from '@/types'
 // Props
 export interface Props {
   ariaLabel: string
@@ -13,7 +12,7 @@ export interface Props {
   button?: boolean
   switch?: boolean
   disabled?: boolean
-  buttonVariant?: string
+  buttonVariant?: ButtonVariant
   inline?: boolean
   required?: boolean
   size?: InputSize
--- a/src/types/components/BFormCheckbox/BFormCheckboxGroup.d.ts
+++ b/src/types/components/BFormCheckbox/BFormCheckboxGroup.d.ts
@@ -1,5 +1,4 @@
-import type {ColorVariant, Size} from '@/types'
-
+import type {ButtonVariant, Size} from '@/types'
 // Props
 export interface Props {
   id?: string
@@ -7,7 +6,7 @@ export interface Props {
   modelValue?: Array
   ariaInvalid?: boolean | string
   autofocus?: boolean
-  buttonVariant?: ColorVariant
+  buttonVariant?: ButtonVariant
   buttons?: boolean
   disabled?: boolean
   disabledField?: string
--- a/src/types/components/BFormRadio/BFormRadio.d.ts
+++ b/src/types/components/BFormRadio/BFormRadio.d.ts
@@ -1,3 +1,4 @@
+import type {ButtonVariant} from '@/types'
 // Props
 export interface Props {
   ariaLabel: string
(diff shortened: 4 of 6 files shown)


aceofwings:
looks good here

kwiksilver3441:
I'm working on converting more, but like I said, these latest ones I'm working on contain the TS issues that break builds. 

Such as https://github.com/cdmoro/bootstrap-vue-3/pull/417/files#annotation_3808927704

How should I handle these? In order to stop the issues, I have assigned the modelValue to be any[] | Set<any> | Booleanish | undefined, however, this causes issues with invalid prop types. (It's not a failing test, just that the prop was set to the wrong type). So, I'm curious, should we align to InputHTMLAttributes.checked types and fix the issues? Or is there a different idea one may have? Ideas @mariusa or @aceofwings ? Sorry for @ ing, but this will determine how to build fails get fixed.

mariusa:
I'm sorry, but I don't have TS experience. Hope @aceofwings can chime in...

> Sorry for @ ing

No need to! Thanks for working on this and trying to figure out the best way!

kwiksilver3441:
> I'm sorry, but I don't have TS experience. Hope @aceofwings can chime in...
> 
> > Sorry for @ ing
> 
> No need to! Thanks for working on this and trying to figure out the best way!

The problem is, that in Bootstrap-vue, these values are allowed. However, according to (What I suspect is actually TSX types), that some of the values are not allowed. So it's about possibly causing a breaking change from boostrap vue to bootstrap vue 3. 

So it really comes down to, do we follow what TS says and cause what would be considered a breaking change from going to bootstrap vue to bootstrap vue 3, or attempt to extend TS ourselves

If we were to follow TS, it would make it simpler to maintain and make it more familiar to those already versed in TSX and native HTML abilities. 

But if we committed to following Bootstrap Vue, then we would need to extend TS and make it harder to maintain (assuming there was ever a change in how HTML inputs work). A pro is that Bootstrap Vue does extend the capability allowing for slightly less boilerplate in specific settings.

Anything that could be built in bootstrap vue, would still be possible, just with a slightly different way of doing them 
For example, https://bootstrap-vue.org/docs/components/form-checkbox example 1, setting the value and unchecked value to strings would not be allowed, however, a workaround in v3 would be to simply v-model to boolean and create a computed value that returns the value/unchecked value string. So it would all still be possible to achieve the same thing. But still be breaking.

So, to follow TS, or to follow Bootstrap Vue? That is the question

mariusa:
> So it's about possibly causing a breaking change from boostrap vue to bootstrap vue 3.

What would be the problem with that?
Migration from old projects to newer versions sometimes cause breaking changes. Let's do this project by the book and people who upgrade (eg. to Vue 3) would upgrade their code too. Or, use it only in new apps with best practices.

> But if we committed to following Bootstrap Vue, then we would need to extend TS and make it harder to maintain 

It's not worth the effort. We have a small js library to get to 1.0, not to extend a programming language. Let's focus on our small lib :)

kwiksilver3441:
> It's not worth the effort. We have a small js library to get to 1.0, not to extend a programming language. Let's focus on our small lib :)

It makes it sound larger than it is, but I'm pretty sure the only thing one would need to do is update the shims-vue.d.ts similar to https://stackoverflow.com/questions/35074713/extending-typescript-global-object-in-node-js

aceofwings:
> I'm working on converting more, but these latest ones I'm working on contain the TS issues that break builds.
> 
> Such as https://github.com/cdmoro/bootstrap-vue-3/pull/417/files#annotation_3808927704
> 
> How should I handle these? In order to stop the issues, I have assigned the modelValue to be any[] | Set | Booleanish | undefined, however, this causes issues with invalid prop types. (It's not a failing test, just that the prop was set to the wrong type). So, I'm curious, should we align to InputHTMLAttributes.checked types and fix the issues? Or is there a different idea one may have? Ideas @mariusa or @aceofwings ? Sorry for @ ing, but this will determine how to build fails get fixed.

Good Question Kiwi. Yes I would align to  InputHTMLAttributes.checked types. Don't worry about pming me. More issues we can solve the better. Once the library is buildable again, I can update the documentation to reflect any changes made.

aceofwings:
> > It's not worth the effort. We have a small js library to get to 1.0, not to extend a programming language. Let's focus on our small lib :)
> 
> It makes it sound larger than it is, but I'm pretty sure the only thing one would need to do is update the shims-vue.d.ts similar to https://stackoverflow.com/questions/35074713/extending-typescript-global-object-in-node-js. Though I haven't tried

We do global type definitions, for example  safe-types extends the window definition adding in additional animation frame attributes.

PR #423, opened 2022-06-24
Various script setup conversion

VividLemon:
This marks a big step in converting the library to use the better script setup syntax. 🥳 The only remaining components that require a change are the before mentioned components that are limited by Vue SFC compiler, and those that use render functions. 
My next planned steps are to first ensure that any components that aren't render functions are error free, I believe there's only one and I left it because it was an odd case. Some render functions are not error free, holding the build back. Second, is to go back and try to ensure that any props that were left without a required tag or default tag before the transition are marked optional (which may be hard, it can be a guessing game, though most props seem optional... ideas?). Finally, proceed to convert the render functions to template syntax. At which point, the majority of the library will have been converted, and will be error free.


mariusa:
Hurray!

VividLemon:
This line may be a bit weird. https://github.com/cdmoro/bootstrap-vue-3/pull/423/files#diff-d1b74bb4934c3110518ce1b566d66a192f73f3650f0c3e6280737437cc1de11dR57 I feel like since it's just a type it shouldn't make any issue between DOM/SSR, but the return type becomes NodeJS.Timeout. Not sure how that affects if it's not Node. Probably shouldn't have any affect, while fixing the error of trying to set a function to a number.

PR #432, opened 2022-07-04
Global Comps starter idea file

VividLemon:
Feel free to not merge this. (Hopefully "yet"). This registers typings for global components. So, for when the library is built, anyone that uses it will get proper typing support in their templates -- the same that one would get from importing the component directly. The only issue is, it will cause a bunch of errors in App.vue (technically they would be considered runtime errors atm) But keep this idea in mind. Unless people think the idea is good, I can work on fixing up the app.vue, then the PR can reviewed again


VividLemon:
It should be noted, I'm pretty sure this would add the typings for when the installing the package, but tbh I don't exactly know how the packaging process works 🤔

aceofwings:
I have to seen. I know early on in the year this library did not have  boostrap.d.ts. The inclusion allowed for users to import components and then have a good typed experience.  Does a global.d.ts from a library get registered when installed?

VividLemon:
> I have to seen. I know early on in the year this library did not have boostrap.d.ts. The inclusion allowed for users to import components and then have a good typed experience. Does a global.d.ts from a library get registered when installed?

That didn't really make a whole lot of sense, but I think I understand. Im pretty confident that adding the package to a project via npm would include the global component types. However, I'm not 100%.

 If it didn't, it would take just a bit of tooling around to get it to include, but I imagine it wouldn't be difficult and I know it can be done. Though, I'm still pretty sure it would be included without any additional adjustments.

I'll probably work on some of the compile time issues in app.vue with this tomorrow to get it hopefully into the next release. @aceofwings you had said you would attempt to resolve the remaining errors in the library, have you made any progress? I can commit to trying to fix them, but I may make a mistake and need to be vigilant

aceofwings:
> That didn't really make a whole lot of sense, but I think I understand. Im pretty confident that adding the package to a project via npm would include the global component types. However, I'm not 100%.
> 
> 

Yeah sorry, just doing some research. This looks good and can be merged in!

VividLemon:
>  This looks good and can be merged in!

On my next commit fixing some of the issues in app.vue it will be ready. But once I do, I'm going to request it to be reviewed. Some of the minor changes to fix the compile issues seem to have been from removed functionality. I just want to ensure that the functionality was removed on purpose and there are no glooming issues. -- Such as (From what I recall) B-table responsive prop not allowing the use of "xs", to name an example

VividLemon:
> Does a global.d.ts from a library get registered when installed?

Yes, I ran a build and it gets included. I suspect it's from vite-plugin-dts, so no extra config required

PR #434, opened 2022-07-05
Emitting click for splitted  and toggle buttons

sfe-efficy:
Hi!
It was pending to emit the click when we split the button for a dropdown like here:

https://bootstrap-vue.org/docs/components/dropdown#listening-to-dropdown-changes-via-root-events:~:text=is%20shown.%20Cancelable-,click,-event%20%2D%20Native

Regards


mariusa:
Thanks!

PR #435, opened 2022-07-05
Fixing building errors

sfe-efficy:


src/components/BFormGroup/BFormGroup.vue
```diff
@@ -69,20 +69,18 @@ export default defineComponent({
     const ariaDescribedby: string | null = null as string | null
     const breakPoints = ['xs', 'sm', 'md', 'lg', 'xl']
 
-    const getAlignClasses = (props: any, prefix: string) => {
-      const alignClasses = breakPoints.reduce((result: string[], breakpoint) => {
```
mariusa, line 73: Are you sure this is no longer needed? Without testing, seems like `breakPoints.reduce()` returns `result`, but what happens to it?

VividLemon: It appears fine. The {} in the arrow function was redundant.

sfe-efficy: The alignClasses variable was redudant, the IDE was suggesting to do directly the return without generating a variable to return again the data

VividLemon (approved):
I would prefer the use of "any" to be removed. But overall everything looks good.

sfe-efficy:
> I would prefer the use of "any" to be removed. But overall everything looks good.

Normally I prefer too, but if I'm honest, I'm don't have too much knowledge about the return type data from the only any introduced.

In the future, it can configure warnings for used of "any" and sometimes we can use manual castings to avoid it

src/components/BFormGroup/BFormGroup.vue
aceofwings: Agreed, good to remove.

src/components/BTable/BTable.vue
```diff
@@ -44,15 +44,15 @@ export default defineComponent({
     const computedFields = computed(() => itemHelper.normaliseFields(props.fields, props.items))
 
     return () => {
-      let theadTop: VNode | null
-      theadTop = null
+      let theadTop: VNode | null = null
```
aceofwings, line 47: Due to slots potentially returning an array of VNodes, I believe we can simplify this to

`Vnode[] | null`

src/components/BTable/BTable.vue
```diff
@@ -44,15 +44,15 @@ export default defineComponent({
...
       if (slots['thead-top']) {
-        theadTop = slots['thead-top']()
+        theadTop = slots['thead-top']() as unknown as VNode
```
aceofwings, line 49: Thus we do not need to the following above but assert optional null

`slots['thead-top']?.()`

Some of the patches `as unknown as Vnode` can be replicated in the same way. Some will need more non-trivial tweaking. 

Let me know what you think.

src/components/BTable/BTable.vue
```diff
@@ -44,15 +44,15 @@ export default defineComponent({
...
       }
 
       let theadSub: VNode | null
```
aceofwings, line 52: `      let theadSub: VNode[] | null`

src/components/BTable/BTable.vue
```diff
@@ -63,7 +63,7 @@ export default defineComponent({
                 scope: 'col',
                 class: [field.class, field.thClass, field.variant ? `table-${field.variant}` : ''],
               },
-              slots['thead-sub']({items: computedFields.value, ...field})
+              slotReference({items: computedFields.value, ...field}) as unknown as VNode
```
aceofwings, line 66: ` slots['thead-sub']?.({items: computedFields.value, ...field})`

For some reason even after checking for the existence of the thead-sub, the editor I am using is still asserting the need for an optional.

aceofwings (commented):
Looks good.

Maybe some minor tweaks.

Anything to get the library buildable again. Means documentation can be updated!
Good stuff.

mariusa:
Thanks @sfe-efficy ! I'll merge as soon as you reply to @aceofwings 's comments

src/components/BTable/BTable.vue
sfe-efficy: Changed. At the end I simplify this line and combine with another suggestions here, so using the optional check "?" with undefined possibility we can inherit the return type and the IDE don't complain anymore.

image

src/components/BTable/BTable.vue
sfe-efficy: Same reply from here
https://github.com/cdmoro/bootstrap-vue-3/pull/435#discussion_r914454202

src/components/BTable/BTable.vue
sfe-efficy: I changed a little bit the code.

image

I guess your IDE was complaining because in my previous commit the "slotReference" should continue having the undefined possible situation, maybe depending the IDE could be more or less smart to discard the undefined possible value.

In my case with the previous code the IDE can detect that the undefined possibility was removed (by the previous if). So, doing the same check in a immutable variable should be easier for a IDE to determinate that possibility was removed.

image

aceofwings:
Thanks for the merge mariusa! Changes look great and further stuff can be improved later in the project. Thanks @sfe-efficy for bringing the errors out of the build. Think the next is to recompile documents and push toward a stable release.

PR #448, opened 2022-07-07
chore: update package-lock.json

messenjer:
I updated the `package-lock.json` file.

We still have the file [.npmrc](https://github.com/cdmoro/bootstrap-vue-3/blob/main/.npmrc), is it useful?

With this content :

```
package-lock=false
```


VividLemon:
I was also wondering why I had a package lock even though it was disabled, just 10 minutes ago

PR #450, opened 2022-07-08
fix(docs): fix boostrap-vue-3 import

messenjer:
When starting the documentation site (local or build), the bootstrap-vue-3 library was no longer loaded.

It was loaded via a local plugin `/docs/.vuepress/PluginComponentReference`, I moved it to the client.ts file https://v2.vuepress.vuejs.org/advanced/cookbook/usage-of-client-config.html#setup 

I also upgraded the version of vue-press-next


messenjer:
I just saw the breaking vuepress breaking change here : 

https://github.com/vuepress/vuepress-next/blob/eb962a057a91142a46a071896de418bc9903c388/CHANGELOG.md#breaking-changes-1

> core: clientAppEnhanceFiles, clientAppRootComponentFiles and clientAppSetupFiles hooks are removed, use clientConfigFile hook instead
> core: conventional file .vuepress/clientAppEnhance.{js,ts} has been renamed to .vuepress/client.{js,ts}, and the usage has been changed too

I fixed these 2 issues.

PR #454, opened 2022-07-08
fix(workspaces): update release please bot to not affect major versions

VividLemon:


VividLemon:
So far having a bad day with getting github and commits to work 😞 Accidentally pushed a commit prematurely and +

PR #455, opened 2022-07-08
fix(workflows): release please bot will not update major versions for…

VividLemon:
… pre 1.0

Release please bot will now cause minor bumps for breaking changes and patch bumps for feat changes (rather than major/minor respectively)

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -10,4 +10,6 @@ jobs:
       - uses: google-github-actions/release-please-action@v3
         with:
           release-type: node
-          package-name: release-please-action
+          package-name: package.json
+          bump-minor-pre-major: true
+          bump-patch-for-minor-pre-major: true


mariusa:
Hi @VividLemon , would you know why the `.github/workflows/npm-publish.yml` won't auto-publish at https://www.npmjs.com/package/bootstrap-vue-3 when having a release made by the release bot?

I see the release appears at https://github.com/cdmoro/bootstrap-vue-3/releases , just like a manual release.

Thanks

VividLemon:
> Hi @VividLemon , would you know why the `.github/workflows/npm-publish.yml` won't auto-publish at https://www.npmjs.com/package/bootstrap-vue-3 when having a release made by the release bot?
> 
> I see the release appears at https://github.com/cdmoro/bootstrap-vue-3/releases , just like a manual release.
> 
> Thanks

I'm a bit unsure. But what it appears is that the bot's PR doesn't contain GitHub action "checks" https://github.com/cdmoro/bootstrap-vue-3/pull/461/checks . So it's possible that the bot doesn't attempt to use actions for itself. 

Rather than trying to fix why bots do and don't do certain things, it could be easier to have the release-please bot publish itself by following these steps https://github.com/google-github-actions/release-please-action#automating-publication-to-npm . I believe you could have it run package.json's "publish" script. But you would still need to utilize the NPM_TOKEN.

VividLemon:
@mariusa It seems the bot failed on attempting to run npm ci on the latest PR, ci can be a bit more aggressive then simple npm i. It seems npm ci has an issue with with how the library is set up with the docs dir. When I remove it and try npm ci it works fine. I can do a temporary fix to have it run npm install instead, but I think a better option may be to convert the library into a typical monorepo setup, and possibly use yarn workspaces instead. It's not certain that it would fix the npm ci issue, and it would regardless be transformed into the yarn equivalent, but I think it would work better. What do you think?
I've also been trying out pnpm, and it's very quick with installs n things... Just a different alternative to think about

mariusa:
Thanks for looking into this!

>  but I think a better option may be to convert the library into a typical monorepo setup, and possibly use yarn workspaces instead

Would it be possible to please use npm workspaces? npm is still bundled with node, would prefer not to add another dependency.

VividLemon:
> Would it be possible to please use npm workspaces?

It would be possible, but I still think it would be better to switch to pnpm. A lot of the major open source libraries, such as Vue core, vitest, and vite have all switched to pnpm because of the beneficial performance they have managed to gather. I think it would be nice to switch to them instead. Though, it's up to you. Though, with having node installed, all one would need to do is "corepack enable" to activate pnpm

mariusa:
Ok. Also, this had no impact: https://github.com/npm/feedback/discussions/341 , based on npm changes and meeting notes.
So let's go with your suggestion with pnpm.

> with having node installed, all one would need to do is "corepack enable" to activate pnpm

Thanks, done

PR #464, opened 2022-07-11
fix(463): replace alias links back to relative

VividLemon:
It seems like a lot of my ideas that I think would be great and useful seem to not be 😞 The disappointing part was that this is the same as #5 ... Though I am not familiar on how npm decides to release things. I did not figure that it would remove the vite config, designating how to use the alias. Regardless, it should fix any issue relating to using aliases. 

I am trying to find a way to have error checking in .d.ts files... It can be easy to attempt to import a missing file. I checked them all this time, but it would be nice to have error detection in .d.ts files.


VividLemon:
If this does get pulled, it would be beneficial to quickly do a new release. As I think 2 current issues pertain to this.

PR #477, opened 2022-07-12
Chore/tsconfig vuepress

messenjer:
To avoid errors like that, https://github.com/cdmoro/bootstrap-vue-3/runs/7303991400?check_suite_focus=true

```
TS2688: Cannot find type definition file for '@vuepress/client/types'.
  The file is in the program because:
    Entry point of type library '@vuepress/client/types' specified in compilerOptions
```

I moved the specific '@vuepress/client/types' typing to `./docs`

I updated also the `package-lock.json`

### DIFF
--- a/docs/tsconfig.json
+++ b/docs/tsconfig.json
@@ -1,0 +1,7 @@
+{
+  "extends": "../tsconfig",
+  "compilerOptions": {
+    "types": ["vitest/globals", "@vuepress/client/types", "vite/client"]
+  },
+  "include": ["./**/*.ts", "./**/*.vue", ".vuepress/**/*.ts", ".vuepress/**/*.vue"]
+}
--- a/package-lock.json
+++ b/package-lock.json
@@ -49,9 +49,10 @@
     "docs": {
       "version": "1.0.0",
       "devDependencies": {
-        "@vuepress/plugin-search": "^2.0.0-beta.41",
+        "@vuepress/plugin-register-components": "^2.0.0-beta.48",
+        "@vuepress/plugin-search": "^2.0.0-beta.48",
         "bootstrap-vue-3": "file:../",
-        "vuepress": "^2.0.0-beta.41"
+        "vuepress": "^2.0.0-beta.48"
       },
       "peerDependencies": {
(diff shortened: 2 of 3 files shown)


VividLemon:
If you're able to run npm ci, you could probably revert https://github.com/cdmoro/bootstrap-vue-3/commit/879ba47eb5d2d24092826b0d8fa7970c3ec169cb . I was discussing with mariusa about switching the library to a monorepo so docs and the library can exist better together.

messenjer:
I have often thought that it would be a good idea to switch to a monorepo with the library, the app, docs, … Which one would you like to use ? Lerna, pnpm, turborepo, … A first solution could be to move /docs to a subfolder for example apps/docs and set up the npm workspace properly.

VividLemon:
I prefer pnpm as the performance they've squeezed out is pretty substantial. I've just started using it on other projects and there was a noticeable difference from using yarn. Plus, a lot of major libraries are using their workspaces, Vue/core, vite, vitest, vuepress, all libraries we use, I figure there has to be some good reason why they are all using pnpm. But mariusa isn't too fond of requiring extra steps and dependencies when switching off pure npm.

mariusa:
> But mariusa isn't too fond of requiring extra steps and dependencies when switching off pure npm.

True :) But since you guys made substantial contributions lately, I'd be ok with `pnpm` based on your reasoning above, but only if `npm` can't do the job (it also supports namespaces). We can leave swiching to `pnpm` only for performance reasons as another discussion.

VividLemon:
> it also supports namespaces

As I understand it, one of the major issues with using npm based workspaces is that it's flattened node_modules architecture means you're duplicating many packages. Bloating the folders and causing long install times. Whereas something like pnpm creates a "pinia like" single source of truth, where a dependency is a single source, and is given copies of instead. Causing less duplication and makes it great for workspaces. In addition, I think with npm we would need to add an eslint rule for no extraneous imports, as I don't think npm contains a built in rejection for trying to import from a different repo in the monorepo. It's your call though mariusa, I'm fine with whatever. Lmk what package manager you would prefer and I'll start working out a monorepo.

PR #478, opened 2022-07-13
List group item `to` prop

Trif4:
Turns list group items into BLinks when the `to` prop is set.

This is my first time editing a library, so let me know if I missed anything!


VividLemon:
Though it may be fine now and would require a bit of a refactor for the component, hopefully when the ability to extend props comes out, we will need to extend BLink props.

PR #482, opened 2022-07-13
refactor: remove extraneous import

VividLemon:
@vue/shared is not included in the docs direct dependencies, causing an extraneous dependency. Rather than include it as a dependency. I have copied the code used from vue and implemented it into a util function. Vue wraps the function in a caching wrapper, but I don't think the caching behavior is necessary for this use case.

I did only test to make sure that the docs portion built. I also attempted to use docs:dev, but it only came up with a generic 404 before and after the change so I don't think anything's broken. I really don't think that "cacheStringFunction" that vue contains is necessary, but I am not certain. Regardless, extraneous imports need to be fixed before migrating to pnpm and a proper monorepo workspace, as they will generally cause breaking, almost impossible to find, issues.


VividLemon:
If you want to read a little more on extraneous imports: https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-extraneous-dependencies.md , they only occur in monorepos so the rule is not auto included, I don't think.

PR #483, opened 2022-07-13
refactor(pnpm): migrate to pnpm and proper workspaces

VividLemon:
DON'T MERGE! This contains issues that will currently break the workflows, and some package scripts, among possibly others. At this point it is purely for review and to give notes! It also contains two fixes for https://github.com/cdmoro/bootstrap-vue-3/pull/482 and https://github.com/cdmoro/bootstrap-vue-3/pull/481 (I closed them and merged them here). 

List of major changes (beyond renames) 
1. release-please.yml Mentioned in #481, I think part of the issue with release please was using an old node version. Regardless, this will need to be changed again in a future commit to fix the npm -> pnpm for workflows. (haven't done it yet) -- I'm pretty certain it was because it was using an old node/npm version, a pnpm noted that npm cli changed recently at some point, so I think the reason why the entire library hasn't broke down is because they changed how peers work
2. .eslintrc , relatively unchanged beyond removing the explicit paths, in favor of a global removal, example: it had ignored ./dist and docs/.vuepress/dist and was changed in favor of simply "dist" -- to account for the fact that ./dist doesn't exist and was moved to package
3. .npmrc , we didn't actually use package-lock=false, so it was removed and added auto-install-peers=true
4. Readme.md , fix a docs issue. 
5. the main package.json was refactored to only include the workspace scripts and dependencies. This is one place I may have issues on
6. The packages/bootstrap-vue-3/README.md is a duplicate of the workspace readme. Unsure if we want to change these.
7. In relation to 5, package.json for bootstrap vue package, this contains the relevant parts for itself. -- Read bottom note !important
8. src/utils/normalize-slot.ts @vue/runtime-core was an extraneous import, VNode is exported directly from Vue, so the two imports were merged into "from 'vue'"
9. Docs had attempted to extend bootstrap-vue-3's tsconfig. Remade tsconfig to another package and exported it's tsconfig for both bootstrap-vue-3 and docs to extend from (good practice).
10. components/component-doc.vue , in relation to #482 it contained an extraneous import from @vue/shared . Replaced it with the copied function from @vue/shared to a new docs/util file. 
11. .vuepress/client.ts , The path for the bootstrap-vue-3 dist changed. It builds, but I am uncertain if this is the best way to achieve this using workspaces
12. util/ folder, read 10. Basically a copy of @vue/shared
The few remaining issues pertain to moving to a workspace, docs/package.json file change, etc. 

Note from 7, in both bootstrap-vue-3 and docs, in pnpm (as well as yarn, I don't know why npm is the oddball), peerDependencies are not automatically downloaded when using pnpm install (or yarn install). I have instead added bootstrap (both of them had that peer) to it's dev dependency as well as remained as a peer. The library will not build without them (as they were referenced but didn't exist). I think this is the correct way to do this, as downloading the package wouldn't include the dev dependencies, and would give notes when adding the package to require the peer. (Note I got confirmation from a pnpm dev that adding as a dev dependency is the correct way)

Please review carefully to see if any scripts/packages in any package.json are in the correct spots, there may be some in bootstrap-vue-3 that are supposed to be in the workspace package.json, and that any other changes don't contain side effects. If anyone has any additional notes on the turbo.json file that they would like to see in the ci, lmk. I'll begin to start cleaning up the workflows and make sure they are also using pnpm and the turbo pipeline.


packages/bootstrap-vue-3/package.json
```diff
@@ -0,0 +1,94 @@
...
+  "types": "./dist/BootstrapVue.d.ts",
+  "workspaces": [
+    "docs/"
+  ],
+  "private": false,
+  "scripts": {
+    "dev": "vite",
+    "build": "vue-tsc --noEmit && vite build",
+    "serve": "vite preview",
+    "docs:dev": "pnpm exec --workspace=docs vuepress dev docs",
```
aceofwings, line 32: I recommend separating pnpm commands into their respective packages. The commands for building and developing docs should go into the docs package.json. Finally in the main package.json, perhaps for ease of use in deployment tools and so forth is too add commands

`      "build:docs": "pnpm -filter=docs build", `

Does the workspace optional argument work with pnpm?

VividLemon: This is a feature of using turborepo. If you observe in each package, docs and bootstrap vue packages, they both contain their own dev/build commands. When running the workspace "build" command, it will run each packages build command in parallel. To run a specific package build, CD into the package and run that specific one. This can be removed, it's up to preference, but quickens the workflow. Feel free to read up on the turborepo docs. https://turborepo.org/ it's up to everyone if they want it removed. It's not specifically necessary to have for a monorepo, just qol.

VividLemon: Technically that package should not contain docs: commands. Still got some work to do for tomorrow.

aceofwings: To your note 7. Peer dependencies and as you confirmed, stop being downloaded after a certain version of node. Their was an earlier issue back in Jan where we discussed dependencies of bootstrap. Why in our docs we discuss about requiring the user to install both Boostrap and boostrap-vue-3. I believe as a dev dependency it makes sense.

aceofwings: Finally I added a dev branch so that we can flush out all issues before we merge into main.

VividLemon: K. Having it as a peer is good,  there was one issue where the person had used a custom ssr friendly version of bootstrap to remove some issues. So it's great to have the option rather than force a specific version.

aceofwings: > This is a feature of using turborepo. If you observe in each package, docs and bootstrap vue packages, they both contain their own dev/build commands. When running the workspace "build" command, it will run each packages build command in parallel. To run a specific package build, CD into the package and run that specific one. This can be removed, it's up to preference, but quickens the workflow. Feel free to read up on the turborepo docs. https://turborepo.org/ it's up to everyone if they want it removed. It's not specifically necessary to have for a monorepo, just qol.

In that case , I think its looking good. Never heard of turbo until now.

VividLemon: Me neither. Not like I work on monorepos every day ¯\_(ツ)_/¯ but they're cool techs

packages/bootstrap-vue-3/package.json
```diff
@@ -0,0 +1,94 @@
...
+  },
+  "peerDependencies": {
+    "bootstrap": "5.x.x"
+  },
+  "dependencies": {
+    "vue": "3.x.x"
+  },
+  "devDependencies": {
+    "bootstrap": "5.x.x",
+    "tsconfig": "workspace:*",
```
messenjer, line 50: It should be

```suggestion
    "tsconfig": "*",
```

packages/bootstrap-vue-3/package.json
```diff
@@ -0,0 +1,94 @@
...
+  },
+  "files": [
+    "dist",
+    "src",
+    "nuxt.js"
+  ],
+  "types": "./dist/BootstrapVue.d.ts",
+  "workspaces": [
+    "docs/"
+  ],
```
messenjer, line 26: You can remove this `workspaces`, we define the workspaces in the root `package.json`

```suggestion
```

package.json
```diff
@@ -1,97 +1,26 @@
...
-    "dist",
-    "src",
-    "nuxt.js"
-  ],
-  "types": "./dist/BootstrapVue.d.ts",
+  "private": true,
   "workspaces": [
-    "docs/"
+    "packages/*"
   ],
```
messenjer, line 5: Here, I will define the workspaces like that:

```
"workspaces": [
    "apps/*"
    "packages/*"
  ],
```

And move the previous `./docs` or now `./packages/docs` to `./apps/docs`

After that we can move also the App.vue from the `./packages/boostrap-vue-3` to `./apps/playground` for example

packages/docs/.vuepress/client.ts
```diff
@@ -7,7 +7,7 @@ import {h} from 'vue'
 export default defineClientConfig({
   async enhance({app, router, siteData}) {
     if (!__VUEPRESS_SSR__) {
-      const {BootstrapVue3, BToastPlugin} = await import('../../dist/bootstrap-vue-3.es')
+      const {BootstrapVue3, BToastPlugin} = await import('../../bootstrap-vue-3/dist/bootstrap-vue-3.es')
```
messenjer, line 10: Would this work?

`const {BootstrapVue3, BToastPlugin} = await import('bootstrap-vue-3')`

packages/docs/package.json
```diff
@@ -11,9 +11,11 @@
     "bootstrap": "5.x.x"
   },
   "devDependencies": {
+    "bootstrap": "5.x.x",
+    "tsconfig": "workspace:*",
```
messenjer, line 15:
```suggestion
    "tsconfig": "*",
```

packages/docs/package.json
```diff
@@ -11,9 +11,11 @@
...
     "@vuepress/plugin-search": "^2.0.0-beta.48",
     "@vuepress/plugin-register-components": "^2.0.0-beta.48",
-    "bootstrap-vue-3": "file:../",
+    "bootstrap-vue-3": "workspace:*",
```
messenjer, line 18:
```suggestion
    "bootstrap-vue-3": "*",
```

packages/bootstrap-vue-3/package.json
VividLemon: without the workspace: it broke the lib. Could not find the tsconfig. Are you sure it isn't required?

messenjer: Did you try a npm install at the root of the folder ?

packages/docs/package.json
messenjer: The same, Forget my comment then...

packages/docs/package.json
messenjer: The same, Forget my comment then...

pnpm-workspace.yaml
```diff
@@ -0,0 +1,2 @@
+packages:
```
messenjer, line 1: Add apps like here : https://github.com/vercel/turborepo/blob/main/examples/with-pnpm/pnpm-workspace.yaml 

` - "apps/*"`

packages/bootstrap-vue-3/package.json
messenjer: > Finally I added a dev branch so that we can flush out all issues before we merge into main.

Could we already merge into dev branch ?

PR #484, opened 2022-07-14
refactor(pnpm): migrate to pnpm and proper workspaces

VividLemon:
Remade pr to go to dev branch. There still contains some issues. 
1. I want to be certain that the package.json's contain their own required dependencies. 
2. I moved the original app.vue to it's own apps/playground/app.vue. This would be fine, if it was working. 
3. Last thing is fixing the workflows. 
If someone can double check that each package contains it's own required dependencies, that would be great. Perhaps ensuring that their scripts are correct too. I moved a lot, eslint, prettier, etc to the main workspace deps. Also, if someone could figure out why exactly the app.vue refuses to run, that would be awesome. I really think those are the only three things to do. I can do the workflows when everythings usable.


apps/docs/package.json
```diff
@@ -11,9 +11,11 @@
     "bootstrap": "5.x.x"
```
VividLemon, line 11: This isn't supposed to be a peer dependency for the docs

apps/docs/.vuepress/client.ts
```diff
@@ -7,7 +7,7 @@ import {h} from 'vue'
 export default defineClientConfig({
   async enhance({app, router, siteData}) {
     if (!__VUEPRESS_SSR__) {
-      const {BootstrapVue3, BToastPlugin} = await import('../../dist/bootstrap-vue-3.es')
+      const {BootstrapVue3, BToastPlugin} = await import('../../../packages/bootstrap-vue-3/dist/bootstrap-vue-3.es')
```
VividLemon, line 10: Use messenjer not from other pr

apps/playground/.gitignore
```diff
@@ -0,0 +1,24 @@
+# Logs
```
VividLemon, line 1: This shouldn't have a separate gitignore

apps/playground/.vscode/extensions.json
```diff
@@ -0,0 +1,3 @@
+{
```
VividLemon, line 1: This could be removed, but could also be moved to workspace .vscode

apps/playground/README.md
```diff
@@ -0,0 +1,16 @@
+# Vue 3 + TypeScript + Vite
```
VividLemon, line 1: Remove this. Maybe redo it to have just a small excerpt that it's for dev

apps/playground/src/App.vue
```diff
@@ -1891,10 +1891,13 @@ import {
   Ref,
   ref,
 } from 'vue'
-import {useBreadcrumb} from './composables/useBreadcrumb'
-import type {TableField, TableItem} from './types'
-import {BvEvent} from './utils/bvEvent'
-import {ToastInstance, useToast} from './components/BToast/plugin'
+import {useBreadcrumb} from '../../../packages/bootstrap-vue-3/src/composables/useBreadcrumb'
```
VividLemon, line 1894: These imports are wrong. I believe this was because vscode tried to adjust the auto imports. I think this is why dev isn't running

apps/playground/src/components/HelloWorld.vue
```diff
@@ -0,0 +1,38 @@
+<script setup lang="ts">
```
VividLemon, line 1: Remove this

packages/bootstrap-vue-3/README.md
```diff
@@ -0,0 +1,91 @@
+<p align="center">
```
VividLemon, line 1: One of these is a duplicate. Pick one

packages/bootstrap-vue-3/package.json
```diff
@@ -0,0 +1,83 @@
+{
```
VividLemon, line 1: Double check these imports and scripts

package.json
```diff
@@ -1,97 +1,35 @@
 {
```
VividLemon, line 1: Double check these imports and scripts

turbo.json
```diff
@@ -0,0 +1,16 @@
+{
```
VividLemon, line 1: Modify this before workflows, workflows is last thing to fix.

VividLemon (commented):
Todos

apps/docs/.vuepress/client.ts
VividLemon: In addition this seems like a vscode auto import issue.

apps/playground/src/App.vue
VividLemon: In dist, we should expose these all in root. Or, at least expose the actual useful things, such as types.

packages/bootstrap-vue-3/src/BootstrapVue.ts
```diff
@@ -0,0 +1,307 @@
...
+    })
+
+    createBreadcrumb(app)
+  },
+}
+
+export {plugin as BootstrapVue3}
+export default plugin
+
+import BAccordion from './components/BAccordion/BAccordion.vue'
```
VividLemon, line 121: I will end up compressing this down, it contains duplicate imports atm. But when doing this, the lib will successfully pull up template intellisense when using global components. Removing the need for that global-components.d.ts file, that only worked when the packages were separated... (I didn't know it didn't work, nice to see the workspace architecture doing it's job!)

VividLemon:
If you review the unresolved conversations, you can see the last major issues I have. In summary, 
1. the dts() plugin is doing it's job, but is having issues with missing some declaration files. I think some are related to using pnpm soft links, but I think some just generally haven't been working. I added in some options to have it show the error messages. Some help on figuring out how to fix them would be amazing.
2. After the migration to a proper workspace architecture, it was shown that some things were not able to be imported by a user of the lib. I don't actually think someone could import the "/types" files. Among this issue, some other things are not able to be imported, none of the composables are able to be imported, which causes apps/playground/app.vue to fail, since it is not able to actually get the composable, it's not allowed (by package.json). There may be more, but those are the ones I noticed. I can implement the ability to expose these, but I would like to know which ones I should expose, and which ones are only to be used internally... Another thing which took a few hours to figure out, and never ending up to resolve it, was managing to expose specific folders from package.json. I managed to expose the /types/index.ts and a /composables/index.ts (that I made) and exporting them from the main BootstrapVue package (So you could access a type for example from import type {SomeType} from 'bootstrap-vue-3', which worked. But should it be exposed to the root? Or should it be exposed to 'bootstrap-vue-3/composables' or 'bootstrap-vue-3/types'? I tried to do the latter, but it didn't work... Kept popping up issues. 

Those are the two major issues, the items not exposed from packages/bootstrap-vue-3 cause an issue in apps/playground/app.vue, since they are not able to access the requested resource, and the dts() issue causes some components not to have .d.ts files -- though, they could be able to access them since they **do** exist in the vue-tsc compiled dist js file. I imagine though, it would render a "cannot get types for this" issue when trying to import that component though. Some help would be appreciated. After that (and some cleanup in workspaces, and that BootstrapVue.ts file), it could technically be merged into main.

packages/bootstrap-vue-3/src/BootstrapVue.ts
VividLemon: I compressed it down, just need to ensure that nothing was accidentally broken. It shouldn't be, but things could have gotten weird by messing with such an essential file 😵‍💫

mariusa:
@aceofwings would you have any suggestions on the 2 remaining issues?

VividLemon:
I'm not 100% on how workflows will run with the "needs:" on jobs. However, if it does modify it's working dir, that means that what we could do instead is merge all workflows into a "ci" workflow, it would run through everything, and reduce duplication. Right now, all workflows clone the repo, install pnpm, set node version to 16, install dependencies, build the library. It's all a bit redundant. If the "needs:" works how I think it does, I could change it and possibly make running the workflows significantly faster. -- Not that it is needed that much.

messenjer:
What a great job!

I had tried last Friday to fork your branch to fix some issues and I had encountered all the same problems. Vuejs dependency, composables not being exposed, ...

I think releasing the app.vue helped to see all these issues.

It's impressive all the fixing you were able to do.

messenjer:
For the vue dts files, may be we can use vue-tsc instead of vite-plug-in-dts https://github.com/johnsoncodehk/volar/discussions/640#discussioncomment-1555479 

I'm on holiday for 2 weeks (without laptop), when I get back I'll be happy to contribute too.

VividLemon:
Some notable changes in the above 3 commits, BPopover was reverted, read it's long description. In short, was just easier to revert it to pre script setup, than to work around Vue SFC compiler

The other one is more interesting, I managed to get vite-plugin-dts to work properly and compile each dts file. Some were caused by the bug mentioned in the commit description. Some of the other components like BTable were not compiling either, but converting them to use pure TS interfaces, liked the others, made everything work. __don't ask me why this works__ 😕 

The only remaining issue preventing build is the issue of exposing those spoken about elements for the package. Composables, types, etc. I was thinking about just exposing everything for the time being (to get it to build and whatnot), and then considering pulling back. 

Lastly, a temporary fix from a bug noticed by moving to the workspace setup was that in apps/app.vue at line 372, doing the `@submit.stop.prevent` caused an error that "undefined" was not assignable to () => void. Using `@submit.stop.prevent` should be possible and allowed. I submitted a temporary fix to stop the error, not the problem.

Like I said, the only thing left to get this workable is correctly exposing types and stuff.

aceofwings:
Have you looked into [this](https://stackoverflow.com/questions/62518396/importing-from-subfolders-for-a-javascript-package#:%7E:text=Votes-,13,-This%20is%20possible)

I have not built the lib in a while so I can't comment on the structure. Think it will be ok to merge to see how everything functions. 

Make sure the build are composables and types into the root folder of Dist.

VividLemon:
Despite the fact that the checks pass, and it builds and tests properly, I'm having issues actually getting things to render in dev. Just posts blank screens.. I removed that deploy.sh script for docs since the workflow action should do it for you, and it didn't seem to get used anywhere in the app... It can be added back, there was some pnpm/npm things to work out though.

VividLemon:
> Have you looked into [this](https://stackoverflow.com/questions/62518396/importing-from-subfolders-for-a-javascript-package#:%7E:text=Votes-,13,-This%20is%20possible)

This is somewhat helpful, but the question is, do we actually want to subroute these files? I'm more asking about how the structure of the dist be. Should we include composables under 'bootstrap-vue-3' or should they be 'bootstrap-vue-3/composables', should it be 'bootstrap-vue-3' or bootstrap-vue-3/types'. The post you showed may be helpful, but how would we prefer to export these items? That's my question.

aceofwings:
Should we include composables under 'bootstrap-vue-3' or should they be 'bootstrap-vue-3/composables', should it be 'bootstrap-vue-3' or bootstrap-vue-3/types'.

Seeing how Vue handles typing and other frameworks, I would vote to keep them within the root. Perhaps for plugins or modules it may be fore a different story but for components, mixins , directives, composables and so forth might be best to keep to the root.

VividLemon:
> Seeing how Vue handles typing and other frameworks, I would vote to keep them within the root. Perhaps for plugins or modules it may be fore a different story but for components, mixins , directives, composables and so forth might be best to keep to the root.

Can you go through and get me a list of which of all of those are only to be used internally? I only need to implement that feature, and fix some weird issues with the apps not displaying, and then I am finished with this change.

aceofwings:
@VividLemon. So I looked through the library, and I can safely say that only the components should be exposed.

VividLemon:
> @VividLemon. So I looked through the library, and I can safely say that only the components should be exposed.

The docs say that "these are a list of types you can use". https://cdmoro.github.io/bootstrap-vue-3/getting-started/#types . It would make sense that if there is a prop exposed, that it's type should be exposed as well ~~ So someone could declare a variable as a type, then assign it to a prop. Are you certain that there are other things that shouldn't be exposed? Because I think I disagree...

VividLemon:
> @VividLemon. So I looked through the library, and I can safely say that only the components should be exposed.

In the latest commits, I have reviewed and exported any types that are used as a prop for a component. The docs mention "you can use these exported types" and all of them on that list are included in those used in props, and are exported. There is a few additions. Any type/util/composable/directive that is exported, is highlighted with a jsdoc @external identifier indicating that it is exported from BootstrapVue.ts. This is how I decided to highlight that we are exporting it. In addition, some components that don't export also include some jsdoc annotations, but they're just auto-generated stuff, and not in depth at all.

VividLemon:
In my unreleased updates I have, I have it capable of outputting the apps/playground. However, I have noticed an issue with the utils/index.ts page. importing the items from utils/dom causes an error since the next import is utils/env. I'm not sure why it works like this, but having this in this order causes it to say "cannot use has_window_support before initialization". This is probably going to be a reversion of the index.ts system, which is annoying... I'm not sure why it would work like that. I may open an issue to vite or vue-tsc...

VividLemon:
If you could merge this request into dev please. Thanks. Then, after it is merged into dev, if someone, or multiple people, could checkout the branch and test it. Considering the scale of the change, I would prefer a few people to review the behavior. -- Especially the ./apps/docs. I was unsure of how it exactly worked before the change, so I may have gotten things wrong. Like, perhaps the ./scripts/deploy.sh could be reintroduced, but from my glances I didn't think it was useful, but it could have been... I don't know.

PR #492, opened 2022-07-20
Added BD component. Props: gap, variant (Default: grid)

martinjack:
* Added **BD** component. Props: **gap**, variant (**Default: grid**)
```

### CODE REVIEW

**Comment by VividLemon** [comment]
I appreciate you taking the time to make things standard with including the types, global-components, etc. It would be super helpful if you also used conventional commit messages as well, so they can be standard and the workflow bots can auto publish changelogs as well. I'm assuming that "BD" would be considered an exclusive component to bootstrap-vue-3, I was wondering if there isn't already a wrapper that could simply include a prop modifier for including the behavior, without including a pretty barebones component? I'm not against it, I'm just wondering if it's super necessary. In addition, with all the wrappers around bbuttons especially, they have caused this issue https://github.com/cdmoro/bootstrap-vue-3/issues/470 (not your fault, just the wrapping of divs around buttons caused it)

**Comment by martinjack** [comment]
Ok, edited message)

**Comment by VividLemon** [comment]
> Ok, edited message)

That's not quite what I meant. The name of the PR doesn't affect the commit. Instead, when you're going to do "git commit -m ...", you should instead follow the conventional commit guideline. So, for this, I would do 
```bash
git commit -m "feat: add bd component"
```

martinjack:
> > Ok, edited message)
> 
> That's not quite what I meant. The name of the PR doesn't affect the commit. Instead, when you're going to do "git commit -m ...", you should instead follow the conventional commit guideline. So, for this, I would do
> 
> ```shell
> git commit -m "feat: add bd component"
> ```
> 
> It's mentioned in the root README but this is a link to their site https://www.conventionalcommits.org/en/v1.0.0/ . It's not a lot of reading, mostly examples.

Ok, edited again

VividLemon:
Do you also mind clarifying if there is no alternative option, without adding in a new component? I feel like there may be other options here.

PR #495, opened 2022-07-22
Fix BTooltip updates for Bootstrap 5.2

Atokulus:
Dear community, dear @cdmoro 

I hope you are all well :)

This pull request will add support for Bootstrap 5.2's new Tooltip API using the `Tooltip.setContent()` method for updating the Tooltip.

As a side effect it will also fix a bug with previous versions of `bootstrap-vue-3` using Bootstrap 5.2, which caused an unshown tooltip to flash up when setting its content via the old way on initialization.

Question remains whether we should keep support for Bootstrap 5.1 and lower?

Best regards
Markus


mariusa:
> Question remains whether we should keep support for Bootstrap 5.1 and lower?

no :)

PR #497, opened 2022-07-23
fix: allow workflow bot to use peerDep

VividLemon:
This was also fixed in the dev branch, but I might as well add it here. This fix should allow workflows to properly install peerDependencies and close https://github.com/cdmoro/bootstrap-vue-3/issues/467 (This could have also been fixed by having the bot use Node 16, rather than Node 12)

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -11,6 +11,7 @@
     "bootstrap": "5.x.x"
   },
   "devDependencies": {
+    "bootstrap": "5.x.x",
     "@vuepress/plugin-search": "^2.0.0-beta.48",
     "@vuepress/plugin-register-components": "^2.0.0-beta.48",
     "bootstrap-vue-3": "file:../",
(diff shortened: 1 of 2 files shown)


mariusa:
> This could have also been fixed by having the bot use Node 16, rather than Node 12

A good solution, keep in mind for future. No need to support old versions with our latest version.

VividLemon:
> > This could have also been fixed by having the bot use Node 16, rather than Node 12
> 
> A good solution, keep in mind for future. No need to support old versions with our latest version.

No, I mean the bot pulls node version 12 to use for itself, doesn't have anything to do with our app, we're all good 👌

mariusa:
Right. I don't know why it runs 2 times :( As you mentioned, it finally published this time.

VividLemon:
> Right. I don't know why it runs 2 times :( As you mentioned, it finally published this time.

I did a little test to see something, 
Before: 
image
After
image

Turns out if npm has both a script that is builtin, and a script that is made, it decides to just use both 🤣 . I will post an issue in npm. But I tested pnpm and it doesn't behave like that.

mariusa:
Appreciate creating an issue with npm!

PR #503, opened 2022-07-26
refactor(playground): use layout in components

VividLemon:


VividLemon:
All i have left to do is reintroduce an app.vue to the main package for dev. Not sure if it should be bare bones or a copy of playground.

PR #506, opened 2022-07-27
docs: move how to publish release to contribution

VividLemon:


VividLemon:
I was thinking, since vue is considered a peer dependency, should it be along side the installation script... But then this really got me thinking, did Vue even need to be a direct dependency 🤔 Those using Nuxt would not need to have Vue installed directly. Would the library work if Vue was moved as a devDependency? 

Deep thoughts...

mariusa:
¯\_(ツ)_/¯

PR #514, opened 2022-07-29
style: activate eslint rules for .vuepress files

VividLemon:
@aceofwings what do you think about activating lint rules for .vuepress folder? If you're not a fan, close the PR. There's a lot of ts files in .vuepress that could be cleaned.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,1 @@
+!apps/docs/docs/.vuepress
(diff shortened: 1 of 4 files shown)


aceofwings:
I will have to take a look. Even though dev was merged into master, I am still reviewing the code/project structure and so forth. I agree with the step forward.

VividLemon:
> I will have to take a look. Even though dev was merged into master, I am still reviewing the code/project structure and so forth. I agree with the step forward.

It seems like everything is working in the docs. Im fairly sure the merge caused a re deploy of the docs, from what I could see in the actions. 

Maybe I was doing the dev wrong because everytime I pulled up the docs on dev it didn't have styles and things. But the code is not any different beyond the extraneous import fixes

PR #519, opened 2022-07-31
feat: navbar

DavidGalicia:
I added a BNavbar component that implements the `variant` and `toggleable` props of the original nav bar. I also added additional components related to BNavbar such as BNavbarToggle. The BCollapse component was updated with a `is-nav` prop.

Here are screen snippets of the Navbar in use:

Navbar expanded:
demo0

Navbar collapsed with menu closed:
demo1

Navbar collapsed with menu opened:
demo2


DavidGalicia:
Closed per Issayah's request

PR #526, opened 2022-08-02
ci: finish release please fix

VividLemon:
Many commits and test fixes later, I finally managed to fix the release please bot to use the workspace and publish it to npm.

### DIFF
--- a/.github/workflows/release-please.yaml
+++ b/.github/workflows/release-please.yaml
@@ -22,18 +22,13 @@ jobs:
         node-version: [16.x]
 
     steps:
-      - name: Clone repository
-        uses: actions/checkout@v3
-        # these if statements ensure that a publication only occurs when
-        # a new release is created:
-
       - name: Use release please
         uses: google-github-actions/release-please-action@v3
         id: release
         with:
           release-type: node
           path: packages/bootstrap-vue-3
-          package-name: bootstrap-vue-3
+          package-name: bootstrap-vue-4
           bump-minor-pre-major: true
           bump-patch-for-minor-pre-major: true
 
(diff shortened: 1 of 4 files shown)


VividLemon:
I also made release-please a bit more resilient so it installs/builds/tests, just as npm-publish.yaml does, before publishing. Since --filter uses that package.json, it actually doesn't run the roots "prepare" script, it uses the package.json prepare script, which doesn't exist. What I noticed in my testing was, that it will still publish, even if ./dist doesn't exist. So it would publish without the types. The additions will safeguard that, since there is no prepare, but just something to note.

(I thought about including a prepare script for that package, but husky is a root dependency and not one for that workspace, so rather than duplicate the dependency and bloat it a bit, I chose to just run all the typical ci suites... Technically, this would be **after** a persons PR got merged, therefore, already running the test suites, therefore running the test twice. Once **before** a person's merged, and once **after** the bots version change gets merged)

VividLemon:
Lastly, it should be noted that post 0.2, the README for the ./packages/bootstrap-vue-3 is slimmed down a lot. This will affect the README on npm once a 0.2.x is released. Should we duplicate it to make it look nice for npm, make it different, or just leave it the way it is?

aceofwings:
> Lastly, it should be noted that post 0.2, the README for the ./packages/bootstrap-vue-3 is slimmed down a lot. This will affect the README on npm once a 0.2.x is released. Should we duplicate it to make it look nice for npm, make it different, or just leave it the way it is?

Yes would be great

VividLemon:
> > Lastly, it should be noted that post 0.2, the README for the ./packages/bootstrap-vue-3 is slimmed down a lot. This will affect the README on npm once a 0.2.x is released. Should we duplicate it to make it look nice for npm, make it different, or just leave it the way it is?
> 
> Yes would be great

Which one? 😆

mariusa:
I think it's fine the way it is. People can always click on github repo.

Issue #547, opened 2022-08-08
FormCheckbox switch style is broken


mariusa:
This used to work in 0.1.13
https://cdmoro.github.io/bootstrap-vue-3/components/FormCheckbox.html#individual-checkbox-switch-style

`form-switch` needs to be added to `<div class="form-check form-switch">`
https://getbootstrap.com/docs/5.2/forms/checks-radios/#switches

mariusa:
https://github.com/cdmoro/bootstrap-vue-3/blob/main/packages/bootstrap-vue-3/src/components/BFormCheckbox/BFormCheckbox.vue

// TODO switch is not used

VividLemon:
I see the issue, switch is used in the `getClasses` external method. It grabs the props directly and sidesteps the resolveBooleanish function. Related to https://github.com/cdmoro/bootstrap-vue-3/pull/544 , since it is boolean | string, doing only "switch", resolves the prop to an empty string, which is considered false. 

I was curious on how the switch was working at first, it didn't seem like it was used anywhere in the component. Tomorrow, I will check through each component and see if there is similar style of offloading props to a external function. If possible, while going through each of those functions, I would like to strongly type them (for example): 
getClasses on this function could be switched to
```ts
(props: {plain?: T, button?: T, inline?: T, switch?: T, size?: T) => 
``` 
It's not a necessary change, but while I'm there. My proposed changes would be that above, then pass in an object literal like so:
```ts
const classes = getClasses(props)
// to
const classes = getClasses({switch: switchBoolean.value}) // etc etc
```
This should fix all the issues. Does that sound ok to you?

mariusa:
Yes! Thank you :)

VividLemon:
I may have to change it up a little bit though. If you simply did ({switch: switchBoolean.value}) the classes would lose reactivity. Doing `getClasses(props)` is fine, but doing `getClasses({foo: props.foo})` will break everything. I'd just need to use `toRefs` util

VividLemon:
[...]

Issue #548, opened 2022-08-09
Unable to add Custom Class on PopOver component


unickortega:
I am also unable to receive event (show, shown, etc) from popover component.

PR #549, opened 2022-08-09
fix: #547 use reactive items for composables

VividLemon:
After the Booleanish update, many composables that grabbed the props object directly would fail to use the resolved booleanish type. This led to issues. Those composables were refactored to use reactive objects, that contained toRef synced copies of the either prop, or other reactive value. Read: https://vuejs.org/api/composition-api-setup.html#accessing-props

> If you really need to destructure the props, ***or need to pass a prop into an external function while retaining reactivity***, you can do so with the toRefs() and toRef() utility APIs:

Some other notable changes, 

- booleanish resolver was moved to a composable just like the above mention, it is no longer an inline computed, which can make changing it easier.
- A small addition to bpopover, the `Popover` class already contains a customClass attribute, all that needs to be done is to pass in a new prop to allow that functionality to fix -- part of -- #548 
- useId was not reactive. It follows the same principles as above. 
- mathMin/mathMax external functions were quite pointless. They didn't add anything to the function besides a polluted import. I normally don't like changing peoples code unless it's needed for a feature, but this was very unnecessary. 
- useAlignment was refactored to use a Ref<> rather than the entire props object. 
- BFormRadio had an index file that exported only one of the two .vue files in that dir. No other component follows this principle
- The internal _getComputedAriaInvalid in useFormCheck's function for resolving ariaInvalid was pulled out and extracted as a utility function in case other components need it. There is one that seems like it should need it, and a note was added for it. 
- Also, a type was added as AriaInvalid = 'grammar' | 'spelling' | Booleanish | undefined

Also, in general, some changes I included also are refactored to be more strongly typed 💪


mariusa:
Fantastic!

PR #559, opened 2022-08-15
fix: unify aria-invalid behavior

VividLemon:


mariusa:
Please merge as you see fit :)

PR #568, opened 2022-08-19
chore: deps should have explicit numbering

VividLemon:
Beyond the peer dependencies, many devDependencies contained non-explicit numbering such as vue-router:4.x.x, These should be replaced with explicit supported versions, in addition, those supported versions also contain the ^ symbol, functioning similar to .x.x syntax. But being more explicit. Perhaps I could go through and audit each package, and determine if they should be a static version, a patch only ~ version, or keep as a ^ verison.

I'm not going to merge this one before I get a general idea on if people think this is a good idea. But I think simply the idea that, as a user, seeing "We explicitly support v5.2 of Bootstrap" is a bit more useful, with the intention that `^` brings, indicating that we have plans to support futures as well.

### DIFF
--- a/apps/playground/package.json
+++ b/apps/playground/package.json
@@ -13,7 +13,7 @@
     "bootstrap": "^5.2.0",
     "bootstrap-vue-3": "workspace:*",
     "vue": "^3.2.37",
-    "vue-router": "4"
+    "vue-router": "4.1.3"
   },
   "devDependencies": {
     "tsconfig": "workspace:*",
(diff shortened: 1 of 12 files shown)


mariusa (approved):
>  These should be replaced with explicit supported versions, in addition, those supported versions also contain the ^ symbol, functioning similar to .x.x syntax. But being more explicit.

Agree with this. Thanks!

VividLemon:
K

PR #571, opened 2022-08-20
Remotes/ace/main

aceofwings:
New Documentation for B-Nav, B-Link and Grid System.

Fix for Dark Mode on Component Reference Plugin


VividLemon:
Nice work. If you'd like, I can work on the docs for navbar components. I've never made a doc reference though, and it would take time out of my building tests, up to you. 

Is there any functional reason why you added back in the `name` property in components? I removed them awhile ago since script setup components don't contain the name, and to keep everything uniform... actually they will all be replaced after all.

aceofwings:
I can do the docs on the navbar if you are still doing tests or jump on tests 

I added the name back for the moment as the docs rely on fetching the component and pulling its name from their definition. I think going forward we will assume the name provided in the reference.json and ill have to address the plugin accordingly. 

I was going back and forth about how to add custom types to the page as currently types are inferred as primitives and I think i have an idea but it will take some time. 

Will also move forward with the convention

VividLemon:
@aceofwings two things I noticed from the docs update, First, feel as if the grid components could have the example cols be a bit more prominent, the contrast is extremely low.
image

Second, the dark mode button has low contrast in light mode 
image
vs 
image
Not sure if this one happened because of you, or if I just never noticed it before. Probably the former. Just some notes for making docs a tad better, if you could possibly take a look.

aceofwings:
I noticed the dark mode button did not contrast well. I believe its just the choice of our navbar at the moment. Will have to target it in a later fix.

PR #598, opened 2022-08-29
BTable sortable and BModal lazy props

VividLemon:
This is a basic addition of sortable and modal lazy

BModal lazy only renders it's content as lazy, and does not actually fetch anything lazily. [devhus](https://github.com/devhus), the original author, had made a <Suspense> addition, but from all of my testing, it didn't actually do anything. It may be possible in the future to further refine this as an async component, but at this time I'm unsure how.

### DIFF
--- a/packages/bootstrap-vue-3/src/components/BModal.vue
+++ b/packages/bootstrap-vue-3/src/components/BModal.vue
@@ -22,8 +22,7 @@
           </div>
           <div class="modal-body" :class="computedBodyClasses">
             <Suspense v-if="!lazyBoolean || (lazyBoolean && modelValueBoolean === true)">
-              <!-- <slot /> -->
-              <ModalContentComponent></ModalContentComponent>
+              <slot />
             </Suspense>
           </div>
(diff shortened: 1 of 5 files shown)


VividLemon:
@devhus Beyond some of the issues that I showed in the other PR. ie duplicate class rules, syntax issues, etc, the biggest change I made was the removal of Suspense and that system. From my testing, it didn't appear to have any affect on the loading of the component. So, at this point `props.lazy` only renders the content in the DOM lazily. I'm more curious with this addition, with the ability for the _user_ to use a <Suspense>/async component, that _then_ it could be asyncronously used. In addition, the v-if check that was added for lazy, could be expanded to include the header/footer, as to further make it 'lazy'. So, I think at this point I'm clear to merge this, but the `lazy` behavior could be possibly expanded upon in the future.

VividLemon:
BEGIN_COMMIT_OVERRIDE
feat(BTable): add sorting option props

feat(BModal): support lazy loading
END_COMMIT_OVERRIDE

PR #601, opened 2022-08-29
BModal lazy prop improvement, BButton loading prop

devhus:
**BModal lazy prop improvement**
- Included the BModal header & footer slots to be lazy-rendered, Which gives makes the component lazy-render work correctly when giving a component to the header/footer slots that require to be mounted after the modal was shown.
- Added an additional variable ``lazyLoadCompleted`` to make sure that the modal's lazy rendered content stays shown until the modal was fully hidden. 

Example of the lazy-rendered modal content on hide/close without the ``lazyLoadCompleted`` variable:
https://www.loom.com/share/e027cb1c0b674c53adbbf906737c0dc2

After adding  ``lazyLoadCompleted` variable & condition:
https://www.loom.com/share/2fc9bc3ce1694336822d34b6a936104c

**BButton loading prop**
- Added a new prop to BButton component giving it the ability to show a BSpiiner when needed, also added a few custom CSS attrs to the BSpinner element to render responsively with the button's size & bootstrap's ``fw-bold`` class.

BButton working example: 
```vue
<template>
  <b-container>
    <div class="d-flex gap-2 align-items-center">
      <b-button
        variant="primary"
        :loading="btnLoading"
        @click="btnLoading = !btnLoading"
        :size="'sm'"
      >Click to {{ btnLoading ? 'stop' : 'start' }} loading!</b-button>
      <b-button
        variant="primary"
        :loading="btnLoading"
        @click="btnLoading = !btnLoading"
        :size="'md'"
      >Click to {{ btnLoading ? 'stop' : 'start' }} loading!</b-button>
      <b-button
        variant="primary"
        :loading="btnLoading"
        @click="btnLoading = !btnLoading"
        :size="'lg'"
      >Click to {{ btnLoading ? 'stop' : 'start' }} loading!</b-button>
    </div>
  </b-container>
</template>

<script  lang="ts" setup>
import { ref } from 'vue';

const btnLoading = ref(false);
</script>
```

BEGIN_COMMIT_OVERRIDE
feat(BButton): add loading prop that makes spinner
feat(BModal): lazily load entire modal when prop lazy

END_COMMIT_OVERRIDE


packages/bootstrap-vue-3/src/components/BButton/BButton.vue
```diff
@@ -1,14 +1,15 @@
 <template>
   <component :is="computedTag" class="btn" :class="classes" v-bind="attrs" @click="clicked">
+    <BSpinner v-if="loading" class="btn-spinner" :small="size != 'lg'"></BSpinner>
```
VividLemon, line 3: I would convert this to instead be 
```vue
<slot name="loading">
    <BSpinner v-if="loading" class="btn-spinner" :small="size != 'lg'" />
</slot>
```
This would allow for people to specify how they want their loading spinner to be.

packages/bootstrap-vue-3/src/components/BModal.vue
```diff
@@ -2,7 +2,15 @@
   <teleport to="body">
     <div :id="id" ref="element" class="modal" :class="modalClasses" tabindex="-1" v-bind="$attrs">
       <div class="modal-dialog" :class="modalDialogClasses">
-        <div class="modal-content" :class="contentClass">
+        <div
```
VividLemon, line 5: Good addition to make the entire modal lazy loaded. I saw the same behavior of the modal not doing it's animation, so I decided not to touch the v-ifs at that time. 👍

VividLemon (commented):
Besides the one issue, I approve.

packages/bootstrap-vue-3/src/components/BButton/_button.scss
```diff
@@ -2,3 +2,24 @@
   background: transparent;
   padding: 0px;
 }
+
+.btn {
```
VividLemon, line 6: Are you sure there's not some built in classes/props that could be used, instead of adding custom css?

VividLemon:
Please, if you intend to continue contributing, you will need to have your commit messages conform to [conventional commits](https://www.conventionalcommits.org/)

packages/bootstrap-vue-3/src/components/BButton/BButton.vue
devhus: I  thought about that but you already have access to the default slot where you can add your custom loading component and it works the same, for example:
```vue
<b-button>
<CustomLoaderComponentThatShowsWhenPropPassed :loading="true" />
Submit
</b-button>
```
Which seems to be much better then having a slot for it like that:
```vue
<b-button>
<template #loading>
<CustomLoaderComponent />
</template>
Submit
</b-button>
```
So, the loading prop on BButton works as a quick solution when you don't need a custom loader but you can still customize its style in your global CSS files as i did in the button's SCSS file

packages/bootstrap-vue-3/src/components/BButton/_button.scss
devhus: I don't think there are any built-in classes/props to achieve the same result, As you can see it only overwrites the ``<BSpinner>`` component CSS to have sizes that fit with the button's size without touching the ``<BSpinner>``  logic/code for sizing.

devhus:
> Please, if you intend to continue contributing, you will need to have your commit messages conform to [conventional commits](https://www.conventionalcommits.org/)

I will take a look and try my best to follow it in any upcoming commits. Sorry if  i am doing something wrong, It's my first time contributing an open-source library.

packages/bootstrap-vue-3/src/components/BButton/BButton.vue
devhus: eqeqeq was fixed and committed.

VividLemon: I was thinking the loading addition should be more similar to https://vuetifyjs.com/en/components/buttons/#loaders 
It removes the text inside while loading, shows a spinner, OR you also have the option to use the slot, which you could overwrite the spinner behavior with something like "loading..." text or whatnot. I simply think their solution looks nice and it's something we could mimic, since Bootstrap has no preference

VividLemon:
> I will take a look and try my best to follow it in any upcoming commits. Sorry if i am doing something wrong, It's my first time contributing an open-source library.

It's nothing detrimental, but it causes issues with the bot that parses through and automates the changelog file, and tag releases. For example, since the last one didn't contain the conventional commit, it didn't actually give any notice in the changelog. Which could be harmful to those updating. But after I got into using conventional commits after the decision to use them here, I exclusively use them on all projects now.

packages/bootstrap-vue-3/src/components/BButton/BButton.vue
devhus: I see, now it does look like a useful thing to have. There you go 
image

packages/bootstrap-vue-3/src/components/BButton/BButtonLoading.vue
```diff
@@ -0,0 +1,22 @@
...
+
+<script setup lang="ts">
+import {InputSize} from 'src/types'
+
+type Props = {
+  loading: boolean
+  mode: 'fill' | 'inline'
+  size: InputSize
+}
+const props = defineProps<Props>()
```
VividLemon, line 21: remove `const props =` when props is not used. Otherwise an eslint error will occur for no unused vars

packages/bootstrap-vue-3/src/components/BButton/BButton.vue
```diff
@@ -22,14 +27,14 @@ export default defineComponent({
...
-    size: {type: String as PropType<InputSize>},
+    size: {type: String as PropType<InputSize>, default: 'md'},
     squared: {type: [Boolean, String] as PropType<Booleanish>, default: false},
     tag: {type: String, default: 'button'},
     target: {type: String as PropType<LinkTarget>, default: '_self'},
     type: {type: String as PropType<ButtonType>, default: 'button'},
     variant: {type: String as PropType<ButtonVariant>, default: 'secondary'},
     loading: {type: Boolean, default: false},
-    loadingDisabled: {type: Boolean, default: false},
+    loadingMode: {type: String as () => 'fill' | 'inline', default: 'inline'},
```
VividLemon, line 37: I think `String as PropType<'fill' | 'inline'>` may work better.

packages/bootstrap-vue-3/src/components/BButton/BButtonLoading.vue
```diff
@@ -0,0 +1,22 @@
+<template>
+  <div
+    v-if="loading"
+    class="btn-loading"
+    :class="{'mode-fill': mode == 'fill', 'mode-inline': mode == 'inline'}"
+  >
+    <slot v-if="loading">
```
VividLemon, line 7: The second v-if is redundant as it would not be rendered if the parent isn't rendered.

packages/bootstrap-vue-3/src/components/BButton/BButtonLoading.vue
```diff
@@ -0,0 +1,22 @@
+<template>
+  <div
+    v-if="loading"
+    class="btn-loading"
+    :class="{'mode-fill': mode == 'fill', 'mode-inline': mode == 'inline'}"
```
VividLemon, line 5: eqeqeq

packages/bootstrap-vue-3/src/components/BButton/BButton.vue
```diff
@@ -22,14 +27,14 @@ export default defineComponent({
...
     pressed: {type: [Boolean, String] as PropType<Booleanish>, default: false},
     rel: {type: String, default: undefined},
-    size: {type: String as PropType<InputSize>},
+    size: {type: String as PropType<InputSize>, default: 'md'},
     squared: {type: [Boolean, String] as PropType<Booleanish>, default: false},
     tag: {type: String, default: 'button'},
     target: {type: String as PropType<LinkTarget>, default: '_self'},
     type: {type: String as PropType<ButtonType>, default: 'button'},
     variant: {type: String as PropType<ButtonVariant>, default: 'secondary'},
     loading: {type: Boolean, default: false},
```
VividLemon, line 36:
```ts
type: [Boolean, String] as PropType<Booleanish>
```
And, of course, resolve the `Booleanish` to `boolean` by using the `useBooleanish()` function

packages/bootstrap-vue-3/src/components/BButton/BButtonLoading.vue
```diff
@@ -0,0 +1,22 @@
...
+    <slot v-if="loading">
+      <BSpinner class="btn-spinner" :small="size !== 'lg'"></BSpinner>
+    </slot>
+  </div>
+</template>
+
+<script setup lang="ts">
+import {InputSize} from 'src/types'
+
+type Props = {
```
VividLemon, line 16:
```ts
interface Props {
loading?: Booleanish
mode?: 'fill' | 'inline'
size?: InputSize
}

const loadingBoolean = useBooleanish(toRef(prop, 'loading'))
```
Though, Booleanish option isn't too useful for an internal item, I guess if it's there >.>

VividLemon (changes requested):
Once that's done, I think it's at a great point to merge.

packages/bootstrap-vue-3/src/components/BButton/BButton.vue
VividLemon: Looks great!

VividLemon:
I'm also kind of skeptical of the use of decomposed .vue files when they are only used in a single component. Tbh, the use of a nested BButtonLoaading makes testing a bit more difficult since in every test I would need to check for the existence of the sub-component, it may be different if the sub-component is a huge bit of logic, but it's overhead. It's up to preference I guess.

VividLemon (commented):
Last fix

devhus:
> I'm also kind of skeptical of the use of decomposed .vue files when they are only used in a single component. Tbh, the use of a nested BButtonLoaading makes testing a bit more difficult since in every test I would need to check for the existence of the sub-component, it may be different if the sub-component is a huge bit of logic, but it's overhead. It's up to preference I guess.

BButtonLoading is removed, and your other suggestions are done 😄

VividLemon:
My last note must have been removed from the commit. But, be careful when using intellisense for importing 

Package imports are perfectly fine `import {ref} from 'vue'` is perfectly fine.

Routes like `import {type} from 'src/types' will break the component when in library mode. 

Not intuitive at all, but I've dealt with the problem before. Use relative paths when vite is in library mode

devhus:
> My last note must have been removed from the commit. But, be careful when using intellisense for importing
> 
> Package imports are perfectly fine `import {ref} from 'vue'` is perfectly fine.
> 
> Routes like `import {type} from 'src/types' will break the component when in library mode.
> 
> Not intuitive at all, but I've dealt with the problem before. Use relative paths when vite is in library mode
> 
> > I've used a child component so i don't increase the BButton component code
> 
> I get why you'd do it, decomposition of components is a good strategy, but sometimes it's better to not

Noted. 👍🏼

PR #608, opened 2022-08-30
b-table #head(), #cell() and field attribute

cattaneoinfoesse:
A possible solution for the issue #375 + field "field" in data

PR #609, opened 2022-08-30
BContainer tag prop, BTable items/rows selection

devhus:
BREAKING CHANGES: .container & .container-fluid CSS classes now have a default 'block' display to behave as a 'div' element when the BContainer tag is changed
BREAKING CHANGE: BTable 'sorted' event output value was changed to an object with keys [sort(string), desc(boolean)] for better ts type support.
BREAKING CHANGES: BTable sortInternal prop default value is now 'false'

fix: added a few missing props to the BTable.d.ts file
fix: replaced BTable emits definition array with a typescript interface for better type support
fix: update the data type of a few of BTable props

feat: Added 'tag' prop to BContainer component
feat: Added BTable selectable prop that adds the selection functionality to the table
feat: Added BTable select-mode prop with a value of [multi, single, range] to expand how the table selection can be used.
feat: Added Btable select-head prop to show an extra Head & Cell to the table
feat: Added selectHead & selectCell slots to customize the selection of the extra head & cells.
feat: BTable selection-variant prop was added, giving the ability to use bootstrap's colors as bg to the selected rows

### DIFF
--- a/packages/bootstrap-vue-3/src/components/BTable/BTable.vue
+++ b/packages/bootstrap-vue-3/src/components/BTable/BTable.vue
@@ -173,7 +173,7 @@ const props = withDefaults(defineProps<BTableProps>(), {
   responsive: false,
   small: false,
   striped: false,
-  sortInternal: true,
+  sortInternal: false,
   selectable: false,
   selectHead: true,
   selectMode: 'single',
--- a/packages/bootstrap-vue-3/src/components/BTable/BTable.vue
+++ b/packages/bootstrap-vue-3/src/components/BTable/BTable.vue
@@ -4,6 +4,11 @@
       <thead>
         <slot v-if="$slots['thead-top']" name="thead-top" />
         <tr>
+          <th v-if="addSelectableCell">
+            <slot name="selectHead">
+              {{ typeof selectHead === 'boolean' ? 'Selected' : selectHead }}
+            </slot>
+          </th>
           <th
             v-for="field in computedFields"
             :key="field.key"
@@ -55,8 +60,14 @@
         <tr
           v-for="(tr, ind) in computedItems"
           :key="ind"
-          :class="tr._rowVariant ? `table-${tr._rowVariant}` : null"
+          :class="getRowClasses(tr)"
+          @click.prevent="onRowClick(tr, ind, $event)"
         >
+          <td v-if="addSelectableCell">
+            <slot name="selectCell">
+              <span :class="selectedItems.has(tr) ? 'text-primary' : ''">🗹</span>
+            </slot>
+          </td>
           <td
             v-for="(field, index) in computedFields"
             :key="field.key"
@@ -112,7 +123,7 @@
 
 <script setup lang="ts">
 // import type {Breakpoint} from '../../types'
-import {computed, toRef} from 'vue'
+import {computed, ref, toRef, useSlots} from 'vue'
 import {useBooleanish} from '../../composables'
 import type {
   Booleanish,
@@ -144,6 +155,9 @@ interface BTableProps {
   sortBy?: string
   sortDesc?: boolean
   sortInternal?: boolean
+  selectable?: boolean
+  selectHead?: boolean | string
+  selectMode?: 'multi' | 'single' | 'range'
 }
 
 const props = withDefaults(defineProps<BTableProps>(), {
@@ -159,9 +173,22 @@ const props = withDefaults(defineProps<BTableProps>(), {
   small: false,
   striped: false,
   sortInternal: true,
+  selectable: false,
+  selectHead: true,
+  selectMode: 'single',
 })
 
-const emits = defineEmits(['update:sortBy', 'update:sortDesc', 'sorted'])
+interface BTableEmits {
+  (e: 'rowSelected', value: TableItem): void
+  (e: 'rowUnselected', value: TableItem): void
+  (e: 'selection', value: TableItem[]): void
+  (e: 'update:sortBy', value: string): void
+  (e: 'update:sortDesc', value: boolean): void
+  (e: 'sorted', ...value: Parameters<(sort?: {by?: string; desc?: boolean}) => any>): void
+}
+
+const emits = defineEmits<BTableEmits>()
+const slots = useSlots()
 
 const captionTopBoolean = useBooleanish(toRef(props, 'captionTop'))
 const borderlessBoolean = useBooleanish(toRef(props, 'borderless'))
@@ -171,6 +198,8 @@ const footCloneBoolean = useBooleanish(toRef(props, 'footClone'))
 const hoverBoolean = useBooleanish(toRef(props, 'hover'))
 const smallBoolean = useBooleanish(toRef(props, 'small'))
 const stripedBoolean = useBooleanish(toRef(props, 'striped'))
+const sortInternalBoolean = useBooleanish(toRef(props, 'sortInternal'))
+const selectableBoolean = useBooleanish(toRef(props, 'selectable'))
 
 const classes = computed(() => [
   'table',
@@ -185,13 +214,16 @@ const classes = computed(() => [
     'table-borderless': borderlessBoolean.value,
     'table-sm': smallBoolean.value,
     'caption-top': captionTopBoolean.value,
+    'b-table-selectable': selectableBoolean.value,
+    [`b-table-select-${props.selectMode}`]: selectableBoolean.value,
+    'b-table-selecting user-select-none': selectableBoolean.value && isSelecting.value,
   },
 ])
 
 const itemHelper = useItemHelper()
 const computedFields = computed(() => itemHelper.normaliseFields(props.fields, props.items))
 const computedItems = computed(() =>
-  props.sortInternal
+  sortInternalBoolean.value === true
     ? itemHelper.sortItems(props.fields, props.items, {key: props.sortBy, desc: props.sortDesc})
     : props.items
 )
@@ -200,6 +232,10 @@ const responsiveClasses = computed(() => ({
   'table-responsive': typeof props.responsive === 'boolean' && props.responsive,
   [`table-responsive-${props.responsive}`]: typeof props.responsive === 'string',
 }))
+
+const addSelectableCell = computed(
+  () => selectableBoolean && (props.selectHead || slots.selectHead !== undefined)
+)
 
 const isSortable = computed(
   () =>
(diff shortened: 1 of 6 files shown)


devhus:
A working example:
```vue
<template>
  <b-container class="py-4">
    <b-form-group
      label="Selection mode:"
      label-for="table-select-mode-select"
      label-cols-md="4"
    >
      <b-form-select
        id="table-select-mode-select"
        v-model="selectMode"
        :options="modes"
        class="mb-3"
      ></b-form-select>
    </b-form-group>

    <b-table
      :items="items"
      :fields="fields"
      :select-mode="selectMode"
      responsive="sm"
      ref="selectableTable"
      selectable
      @row-selected="onRowSelected"
      @row-unselected="onRowUnSelected"
      @selection="onSelection"
      :select-cell="false"
    >
      <template #selectHead>Is Selected?</template>
      <template #selectCell>✔</template>
      <!-- Example scoped slot for select state illustrative purposes -->
      <template #cell(selected)="{ rowSelected }">
        <template v-if="rowSelected">
          <span aria-hidden="true">&check;</span>
          <span class="sr-only">Selected</span>
        </template>
        <template v-else>
          <span aria-hidden="true">&nbsp;</span>
          <span class="sr-only">Not selected</span>
        </template>
      </template>
    </b-table>
    <p>
      Selected Rows:<br>
      {{ selected }}
    </p>
    <p>
      Internal Selected Rows:<br>
      {{ internalSelected }}
    </p>
  </b-container>
</template>

<script  lang="ts" >
import { defineComponent } from "vue";
import { BTableProps } from "./types/components";

  data() {
    return {
      modes: ['multi', 'single', 'range'],
      fields: ['isActive', 'age', 'first_name', 'last_name'],
      items: [
        { isActive: true, age: 40, first_name: 'Dickerson', last_name: 'Macdonald' },
        { isActive: false, age: 21, first_name: 'Larsen', last_name: 'Shaw' },
        { isActive: false, age: 89, first_name: 'Geneva', last_name: 'Wilson' },
        { isActive: true, age: 38, first_name: 'Jami', last_name: 'Carney' }
      ],
      selectMode: 'multi' as BTableProps['selectMode'],
      selected: [],
      internalSelected: [],
    }
  },
  methods: {
    onRowSelected(item: any) {
      this.selected.push(item);
    },
    onRowUnSelected(item: any) {
      this.selected.splice(this.selected.indexOf(item), 1);
    },
    onSelection(items: any) {
      this.internalSelected = items;
    },
  }
});
</script>
```

More logic to be added later to give the ability to update selectedItems array through props using v-model, and provide few methods to control the internal selectedItems array (such as, selectAllRows, clearSelected, selectRow(index))

VividLemon:
If you could possibly address https://github.com/cdmoro/bootstrap-vue-3/pull/608. Feel free to give them a message, figure out a solution, or copy both the changes together, as to not have any merge conflicts.

VividLemon:
If you could possibly address https://github.com/cdmoro/bootstrap-vue-3/pull/608. Feel free to give them a message, figure out a solution, or copy both the changes together, as to not have any merge conflicts. @cattaneoinfoesse

devhus:
> If you could possibly address #608. Feel free to give them a message, figure out a solution, or copy both the changes together, as to not have any merge conflicts. @cattaneoinfoesse

done.

packages/bootstrap-vue-3/src/components/BTable/BTable.vue
```diff
@@ -201,6 +235,10 @@ const responsiveClasses = computed(() => ({
   [`table-responsive-${props.responsive}`]: typeof props.responsive === 'string',
 }))
 
+const addSelectableCell = computed(
+  () => selectableBoolean && (props.selectHead || slots.selectHead !== undefined)
```
VividLemon, line 239: selectableBoolean.value

packages/bootstrap-vue-3/src/components/BTable/BTable.vue
```diff
@@ -144,6 +155,10 @@ interface BTableProps {
   sortBy?: string
   sortDesc?: boolean
   sortInternal?: boolean
+  selectable?: boolean
```
VividLemon, line 158: You included this as a `useBooleanish` option, but did not set the type to `Booleanish`. That and sortInternal, not that it was Booleanish though either. If you could handle all of them, ie sortDesc, sortInternal, selectable, that would be great.

VividLemon: sortInternal doesn't have a default value. Perhaps it should be false by default?

packages/bootstrap-vue-3/src/components/BTable/BTable.vue
VividLemon: !!props.selectHead
Just since it could be possible that things could get confused.

packages/bootstrap-vue-3/src/types/components/BTable/BTable.d.ts
```diff
@@ -16,6 +16,9 @@ export interface Props {
   small?: boolean
   striped?: boolean
   variant?: ColorVariant
+  sortInternal?: boolean
```
VividLemon, line 19: At this time you don't _technically_ need to bother with keeping this list updated. Once Vue supports type imports for props and emits, all the component .d.ts files will need to be updated.

VividLemon (changes requested):

packages/bootstrap-vue-3/src/types/components/BTable/BTable.d.ts
devhus: noted

packages/bootstrap-vue-3/src/components/BTable/BTable.vue
devhus: > You included this as a `useBooleanish` option, but did not set the type to `Booleanish`. That and sortInternal, not that it was Booleanish though either. If you could handle all of them, ie sortDesc, sortInternal, selectable, that would be great.

done


> sortInternal doesn't have a default value. Perhaps it should be false by default?

It has a default value as true, changed it to false.

packages/bootstrap-vue-3/src/components/BTable/BTable.vue
devhus: > !!props.selectHead Just since it could be possible that things could get confused.

Both the pro and slot doing kinda the same job, passing either of both will show/replace the select head & cells. 
The prop is a quicker method to change the header as a text.
The slot is a better method to change the element as a whole.

What is your suggestion?

packages/bootstrap-vue-3/src/components/BTable/BTable.vue
VividLemon: > It has a default value as true, changed it to false

It did? I must have not seen it then

packages/bootstrap-vue-3/src/components/BTable/BTable.vue
```diff
@@ -158,10 +173,24 @@ const props = withDefaults(defineProps<BTableProps>(), {
...
+  selectionVariant: 'primary',
 })
 
-const emits = defineEmits(['update:sortBy', 'update:sortDesc', 'sorted'])
+interface BTableEmits {
+  (e: 'rowSelected', value: TableItem): void
+  (e: 'rowUnselected', value: TableItem): void
+  (e: 'selection', value: TableItem[]): void
+  (e: 'update:sortBy', value: string): void
+  (e: 'update:sortDesc', value: Booleanish): void
```
VividLemon, line 188: I think it's best to follow the practices of other components here. Basically, when a prop type is only `boolean` it can be converted to `Booleanish`, immediately in script it is resolved down to `boolean` and only the `useBooleanish` value is used throughout the script. 

Basically, I think you should convert this emit value back to `boolean`, and only use the resolved `boolean` values throughout the script, and never pass `Booleanish` values around. There's nothing wrong with passing them around, I think it's best to just keep everything on the same page and using pure `boolean` values.

packages/bootstrap-vue-3/src/components/BTable/BTable.vue
VividLemon: Do you mean for example that: 
```ts
<slot name="title">
   {{props.title}}
</slot>
```
is your issue here, and is redundant? Or that there is something wrong with this particular `||` statement? I'm a little confused.

packages/bootstrap-vue-3/src/components/BTable/BTable.vue
devhus: done, now it does feel right.

packages/bootstrap-vue-3/src/components/BTable/BTable.vue
devhus: I got confused as well. I don't think there is an issue in that part, just trying to understand your point.
What is the issue here exactly?

VividLemon: Since props.selectHead is type `boolean | string`, a compiler _could_ get confused by the string part. replacing `selectableBoolean.value && (props.selectHead || slots.selectHead !== undefined)` with `selectableBoolean.value && (!!props.selectHead || slots.selectHead !== undefined)` is "kinda, sorta, good enough" to ensure it checks against boolean values. Just js weirdness

VividLemon: Like, doing something like 
```ts
interface Props {
someProp?: boolean | string
}
const classes = computed<Record<string, boolean>>(() => ({
   someClass: props.someProp
})
```
Would cause a ts error, since string can't resolve to boolean. This is just an example, but it illustrates the point. This is probably a non-issue since doing `foo && bar` should make everything boolean... but JS is weird

devhus: Good, I've learned something new today.
It is done.

VividLemon: It will still do something like this when is string:
```ts
const a = ''
console.log(!!a) // false
const b = 'a' // adding even a single character
console.log(!!b) // true
```
So doing !! might give a weird issue if a user ever passes an empty string, but I think it's usually expected.

ahmedrazaa:
seems working fine for me except to know one thing, how i can change icon for not selected item it always show ✔ for all rows.<template v-slot:cell(selected)="rowSelected ">
          
          <template v-if="rowSelected">
          <span aria-hidden="true">&check;</span>
          <span class="sr-only">Selected</span>
        </template>
        <template v-else>
          <span aria-hidden="true">&nbsp;</span>
          <span class="sr-only">Not selected</span>
        </template> 
       </template> 
it looks like that part of code isn't working, always show ✔ for all rows

devhus:
> seems working fine for me except to know one thing, how i can change icon for not selected item it always show ✔ for all rows.<template v-slot:cell(selected)="rowSelected ">
> 
> ```
>       <template v-if="rowSelected">
>       <span aria-hidden="true">&check;</span>
>       <span class="sr-only">Selected</span>
>     </template>
>     <template v-else>
>       <span aria-hidden="true">&nbsp;</span>
>       <span class="sr-only">Not selected</span>
>     </template> 
>    </template> 
> ```
> 
> it looks like that part of code isn't working, always show ✔ for all rows

with the upcoming release update, it will be customizable using CSS. Please take a look at #643

VividLemon:
I think it would be nice if additionally those values could be modified using slots, for those like me that don't like touching css.

yashwp:
I'm using `0.16.6` version of this package. But I cannot use `rowSelected` to render checkboxes for each row and select/deselect on row selection.

When I can expect this fix?

```
<b-table responsive hover :items="items" :fields="columns" tbody-tr-class="cursor-pointer"
               selectable select-mode="multi" @row-selected="onRowSelect" :select-head="true"
               @row-unselected="onRowDeselect">
        <template #cell(null)="{ rowSelected }">
          {{ rowSelected }}
          <input type="checkbox" :checked="rowSelected">
        </template>
      </b-table>
      ```

noque-lind:
Same issue for me on 17.6 as well!

PR #624, opened 2022-09-05
BTable sticky header, columns, and selection. Along with a few fixes/additions

devhus:
feat(BTable): completed the ``TableItem.sticky-column`` prop logic, according to bootstrap-vue2 docs
feat(BTable): added ``sticky-header`` prop feature, according to bootstrap-vue2 docs
feat(BTable): new ``sticky-select`` prop was added, the prop converts the selection column to the sticky mode when the selectable prop is passed as true to the table
feat(BTable): row's ``_showDetails`` prop logic was completed
feat(BTable): added ``busy`` prop logic, the prop shows a spinner and message that can be replaced through a slot template ``#table-busy``

### DIFF
--- a/packages/bootstrap-vue-3/src/components/BTable/BTable.vue
+++ b/packages/bootstrap-vue-3/src/components/BTable/BTable.vue
@@ -104,7 +104,7 @@
           </tr>
 
           <tr v-if="tr._showDetails === true && $slots['row-details']" :class="getRowClasses(tr)">
-            <td :colspan="computedFields.length + (selectableBoolean ? 1 : 0)">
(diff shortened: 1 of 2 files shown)


devhus:
There will be no more commits to this PR, only if needed a few commits for fixes for the review.

rubebop:
Since it was merged into the main, I guess this was already released, right?
I wrote [here](https://github.com/bootstrap-vue-next/bootstrap-vue-next/discussions/1225) some question concerning the `sticky-header` in a BTable: it still doesn't seem to work → I am missing something?

PR #632, opened 2022-09-07
fix(composables): Spread out children from Fragment slots

arrudaje:
# Describe the PR

Vue3 introduced the concept of Fragments, which allow placing multiple elements as roots of a component, but also they're used to group elements iterated through in a `v-for` directive.

For example, when we write
```vue
<template>
    <b-form-radio-group>
        <b-form-radio v-for="option in options" :value="option">
            {{ option }}
        </b-form-radio>
    </b-form-radio-group>
</template>
<script>
export default {
    computed: {
        options() {
            return [ 1, 2, 3 ];
        }
    }
}
</script>
```

it compiles (roughly) to:
```json
{
  "type": "Symbol(Fragment)",
  "children": [
    {
      "type": {
        "__name": "BFormRadio"
      }
    },
    {
      "type": {
        "__name": "BFormRadio"
      }
    },
    {
      "type": {
        "__name": "BFormRadio"
      }
    }
  ]
}
```

So this PR makes the conversion of slots into elements recognise that, if we have fragments like this in the slots, we need to spread out the children so that they are rendered as well, otherwise they will be filtered out by the `.filter()` chain that disregards all nodes that are not of the expected type.

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

- [x] Pull request title and all commits follow the [**Conventional Commits**](https://www.conventionalcommits.org/) naming convention **This is very important, as the `CHANGELOG` is generated from these messages, and determines the next version type**

### DIFF
--- a/packages/bootstrap-vue-3/src/composables/useFormCheck.ts
+++ b/packages/bootstrap-vue-3/src/composables/useFormCheck.ts
@@ -84,6 +84,8 @@ const getGroupClasses = (items: {
     [`btn-group-${items.size}`]: items.size,
   }))
 
+// TODO this function is similarly used in BTabs and may be capable of being a util function
+// Investigate if it can be done to reduce complexity
 /**
  * @param slots
  * @param nodeType
--- a/packages/bootstrap-vue-3/src/composables/useFormCheck.ts
+++ b/packages/bootstrap-vue-3/src/composables/useFormCheck.ts
@@ -92,9 +92,9 @@ const getGroupClasses = (items: {
  */
 const slotsToElements = (slots: Array<any>, nodeType: string, disabled: boolean) =>
   slots
-    .reduce((acc: Array<any>, slot: any) => slot.type.name === 'Symbol(Fragment)' ?
+    .reduce((acc: Array<any>, slot: any) => slot.type.toString() === 'Symbol(Fragment)' ?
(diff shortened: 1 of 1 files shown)


arrudaje:
Resolves #633

VividLemon:
I can confidently say that _I have no idea what's happening here_

arrudaje:
@VividLemon this is a problem I faced while implementing bootstrap-vue-3 in my project. I have a list of options that come from a service, and additionally to the radios showing up, I want to add a little icon with a tooltip beside some of the options, thus the need to make them a slotted component. If you try to do something similar to what I demonstrated you will see that the radios are not rendered. This is because the `slotsToElements` function filters out everything that's not a `BFormRadio` component, but it just happens that when you iterate over a list to generate radios, they actually end up as children of a Fragment component, that's the root of the slot, and when you place this inside a BFormRadioGroup component, it gets lost along the way.

My fix was just to check if one of the root elements of the slot was a Fragment instead of a proper component, and in a positive case, extract the children in order to make them root elements, so the `filter` chain can properly let the intended BFormRadios to pass.

VividLemon:
What you have sounds like it could be correct, but I find no difference in the example you gave before and after.

arrudaje:
@VividLemon my fix had some bugs, I'm sorry for not testing before. I have fixed it and now it's working

PR #647, opened 2022-09-11
feat(tooltip): :sparkles: Handle tooltip config as value instead of m…

imanmalekian31:
# Describe the PR

Added handling values directive beside modifiers.

## PR checklist

- [x] Feature - `feat(...)`


VividLemon:
I'm curious on what exactly the purpose is... What issue does it seek to resolve? Ease of use, compatibility, etc?

imanmalekian31:
It had been handled in bootstrap vue and because of that, I added it.
And it's so useful, instead of directives.

PR #650, opened 2022-09-12
Fix backdrop being rendered over modal-dialog

Atokulus:
Currently the DOM order results in the backdrop being drawn over the modal-dialog itself, making the modal inoperable.


Atokulus:
Example screenshot:

backdrop_overlays_modal_itself

VividLemon:
A small reproduction would be useful, as I'm not seeing an issue with mine.

Atokulus:
Indeed, this seems to be a bug with my way of importing Bootstraps latest CSS variables, mainly the z-indices of the modal components.

https://github.com/cdmoro/bootstrap-vue-3

VividLemon:
I don't think using the pure scss file will import _scoped_ css, which seems to be the issue. Though, I'm surprised Vite doesn't code split the css, especially in a component scoped style

PR #655, opened 2022-09-12
docs: add some missing components

VividLemon:
@aceofwings I'm still having issues with getting some `<component-reference>` items to use the json file instead of relying on the automatic types. I don't quite understand because Accordion.md works just fine, but when I use the made Offcanvas.md file, it doesn't use the .json types I made. The way the pages work is a bit weird to me. If you could review the `apps\docs\docs\component-references\schema.json` file, and remark about how the schema can be improved, especially with your docs slot changes. There was some portions when reviewing other json files that confused me, so notes or fixing them yourself would help a lot. It sucks that these things can't be auto generated. If you could elaborate on how exactly the system works between the .md files, the component reference, and how it choses between auto-generating and the json files, that would be a big help.


aceofwings:
Yep, even though the slots pull request was merged I still can't figure out what is wrong. I must be using row details incorrectly as @devhus  example works and functions as expected. 

Also instead of using name as an key to target a prop using must use prop as the key eg.
```
{prop: "myProp",
type: "Booleanish",
description: "example:"
}

**Comment by VividLemon** [comment]
> Also instead of using name as an key to target a prop, you must use prop as the key eg.

I don't understand what you mean. Could you provide a real example? All of my additions are similar to this: 
```json
{
 "name": "bodyScrolling",
 "type": "Booleanish"
},
```

aceofwings:
Instead of this

```json
{
 "name": "bodyScrolling",
 "type": "Booleanish"
},
```

you must do this
```json
{
 "prop": "bodyScrolling",
 "type": "Booleanish"
},
```

Change all your "name" keys to "prop" keys and your Accordion.json should work

VividLemon:
I'm 



Incompetent

aceofwings:
lolol no worries

Issue #657, opened 2022-09-13
The close button in the top right of <BModal> does not emit an update to v-model value.


Atokulus:
Hey community
I hope you are all well.

When showing a BModal and closing the modal with the close button (cross button) in the top right, the modal's v-model binding is not updated, leading to issues when wanting to reshow the modal.

See https://codesandbox.io/s/modal-close-bug-9sqsl6?file=/src/App.vue for a reproduction of this issue.

Best regards
Markus

PR #658, opened 2022-09-13
Possible fix for close button in the top right of modal to emit an update to v-model value.

Atokulus:
This is a possible fix for #657: The close button in the top right of <BModal> does not emit an update to v-model value.


Atokulus:
Alternative would be to explicitly register a Vue click handler on the close button.

VividLemon:
https://github.com/cdmoro/bootstrap-vue-3/pull/661 preferred

PR #661, opened 2022-09-13
BModal header's close button not updating v-model value

devhus:
This issue mentioned in https://github.com/cdmoro/bootstrap-vue-3/pull/658
He was correct in the second comment https://github.com/cdmoro/bootstrap-vue-3/pull/658#issuecomment-1245293371 
The close event should be fired through vue's click handler, Events should never depend on Bootstrap's event callbacks in such matter since we can't fully control the events coming from Bootstrap

BEGIN_COMMIT_OVERRIDE
fix(BModal): fixed the modal's close button located in the header where the button click was not updating ``v-model`` value
END_COMMIT_OVERRIDE


Atokulus:
Hey there
Just realized there still another case, where the binding is not updated: #674

PR #677, opened 2022-09-14
A possible solution for BModal escape-key close event

devhus:
PR for issue https://github.com/cdmoro/bootstrap-vue-3/issues/674
____
I have tried my best to make the events vue-based only but it was not working out smoothly. that's the best result I've got so far.
This PR will fix the issue and also make sure escape-key closing works fine with nested modals since bootstrap's modal does not really support multiple/nested modal instances

example of working code
```vue
<div>
      <b-button @click="showModal = true">Show Modal</b-button>

      <b-modal v-model="showModal" id="my-modal">
        Hello From My Modal!

        <b-button @click="showNestedModal = true">Show Modal</b-button>

        <b-modal v-model="showNestedModal" id="my-modal">Hello From Nested Modal!</b-modal>
      </b-modal>
    </div>
```

### DIFF
--- a/packages/bootstrap-vue-3/src/components/BModal.vue
+++ b/packages/bootstrap-vue-3/src/components/BModal.vue
@@ -1,7 +1,7 @@
 <template>
   <teleport to="body">
     <div
-      :id="elementId"
+      :id="computedId"
       ref="element"
       class="modal"
       :class="modalClasses"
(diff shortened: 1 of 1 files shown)


packages/bootstrap-vue-3/src/components/BModal.vue
```diff
@@ -176,6 +184,9 @@ const titleSrOnlyBoolean = useBooleanish(toRef(props, 'titleSrOnly'))
 
 const lazyLoadCompleted = ref(false)
 
+const bsId = ref(getId('bmodal'))
+const elementId = computed(() => props.id ?? bsId.value)
```
VividLemon, line 188: If bsId is only used here. It's probably better to follow the standard practice and use the _useId_ composable
```ts
const computedId = useId(toRef(props, 'id'), 'modal')
```

PR #678, opened 2022-09-15
An attempt to implement items provider

devhus:
this PR is an attempt to implement the feature mentioned the in issue https://github.com/cdmoro/bootstrap-vue-3/issues/673
____

I've tried my best to bring this feature on but not quite sure if did it right.
Note that the provider is now a standalone prop not as it used to be in BSVue2 and should be passed through the ``provider`` prop instead of the ``items`` prop

fully working example:
```vue
<template>
  <b-container class="py-5">
    <!-- User Interface controls -->
    <b-row>
      <b-col lg="6" class="my-1">
        <b-form-group
          v-slot="{ariaDescribedby}"
          label="Sort"
          label-for="sort-by-select"
          label-cols-sm="3"
          label-align-sm="right"
          label-size="sm"
          class="mb-0"
        >
          <b-input-group size="sm">
            <b-form-select
              id="sort-by-select"
              v-model="sortBy"
              :options="sortOptions"
              :aria-describedby="ariaDescribedby"
              class="w-75"
            >
              <template #first>
                <option value="">-- none --</option>
              </template>
            </b-form-select>

            <b-form-select
              v-model="sortDesc"
              :disabled="!sortBy"
              :aria-describedby="ariaDescribedby"
              size="sm"
              class="w-25"
            >
              <option :value="false">Asc</option>
              <option :value="true">Desc</option>
            </b-form-select>
          </b-input-group>
        </b-form-group>
      </b-col>

      <b-col lg="6" class="my-1">
        <b-form-group
          label="Filter"
          label-for="filter-input"
          label-cols-sm="3"
          label-align-sm="right"
          label-size="sm"
          class="mb-0"
        >
          <b-input-group size="sm">
            <b-form-input
              id="filter-input"
              v-model="filter"
              type="search"
              placeholder="Type to Search"
            />

            <b-input-group-append>
              <b-button :disabled="!filter" @click="filter = ''">Clear</b-button>
            </b-input-group-append>
          </b-input-group>
        </b-form-group>
      </b-col>

      <b-col lg="6" class="my-1">
        <b-form-group
          v-slot="{ariaDescribedby}"
          v-model="sortDirection"
          label="Filter On"
          description="Leave all unchecked to filter on all data"
          label-cols-sm="3"
          label-align-sm="right"
          label-size="sm"
          class="mb-0"
        >
          <div class="d-flex gap-2">
            <b-form-checkbox v-model="filterOn" value="name">Name</b-form-checkbox>
            <b-form-checkbox v-model="filterOn" value="age">Age</b-form-checkbox>
            <b-form-checkbox v-model="filterOn" value="isActive">Active</b-form-checkbox>
          </div>
        </b-form-group>
      </b-col>

      <b-col sm="5" md="6" class="my-1">
        <b-form-group
          label="Per page"
          label-for="per-page-select"
          label-cols-sm="6"
          label-cols-md="4"
          label-cols-lg="3"
          label-align-sm="right"
          label-size="sm"
          class="mb-0"
        >
          <b-form-select id="per-page-select" v-model="perPage" :options="pageOptions" size="sm" />
        </b-form-group>
      </b-col>

      <b-col sm="7" md="6" class="my-1">
        <b-pagination
          v-model="currentPage"
          :total-rows="totalRows"
          :per-page="perPage"
          align="fill"
          size="sm"
          class="my-0"
        />
      </b-col>
    </b-row>

    <!-- Main table element -->
    <b-table
      v-model:sort-by="sortBy"
      v-model:sort-desc="sortDesc"
      v-model:busy="busy"
      :items="items"
      :fields="fields"
      :current-page="currentPage"
      :per-page="perPage"
      :filter="filter"
      responsive
      :filterable="filterOn"
      small
      :sort-internal="true"
      :provider="myProvider"
      :no-providing="['paging']"
      @filtered="onFiltered"
    >
      <template #cell(name)="row"> {{ row.value.first }} {{ row.value.last }} </template>

      <template #cell(actions)="row">
        <b-button size="sm" class="mr-1" @click="info(row.item, row.index, $event.target)">
          Info modal
        </b-button>
        <b-button size="sm" @click="row.toggleDetails">
          {{ row.detailsShowing ? 'Hide' : 'Show' }} Details
        </b-button>
      </template>

      <template #row-details="row">
        <b-card>
          <ul>
            <li v-for="(value, key) in row.item" :key="key">{{ key }}: {{ value }}</li>
            <b-button size="sm" @click="row.toggleDetails"> Toggle Details </b-button>
          </ul>
        </b-card>
      </template>
    </b-table>

    <!-- Info modal -->
    <b-modal
      :id="infoModal.id"
      v-model="infoModal.open"
      :title="infoModal.title"
      ok-only
      @hide="resetInfoModal"
    >
      <pre>{{ infoModal.content }}</pre>
    </b-modal>
  </b-container>
</template>

<script lang="ts">
import { defineComponent } from 'vue'
import { ColorVariant, TableItem } from './types'
import { BTableProviderContext } from './types/components'

  data() {
    return {
      items: [
        {isActive: true, age: 40, name: {first: 'Dickerson', last: 'Macdonald'}},
        {isActive: false, age: 21, name: {first: 'Larsen', last: 'Shaw'}},
        {
          isActive: false,
          age: 9,
          name: {first: 'Mini', last: 'Navarro'},
          _rowVariant: 'success' as ColorVariant,
        },
        {isActive: false, age: 89, name: {first: 'Geneva', last: 'Wilson'}},
        {isActive: true, age: 38, name: {first: 'Jami', last: 'Carney'}},
        {isActive: false, age: 27, name: {first: 'Essie', last: 'Dunlap'}},
        {isActive: true, age: 40, name: {first: 'Thor', last: 'Macdonald'}},
        {
          isActive: true,
          age: 87,
          name: {first: 'Larsen', last: 'Shaw'},
          _cellVariants: {age: 'danger', isActive: 'warning'} as Record<any, any>,
        },
        {isActive: false, age: 26, name: {first: 'Mitzi', last: 'Navarro'}},
        {isActive: false, age: 22, name: {first: 'Genevieve', last: 'Wilson'}},
        {isActive: true, age: 38, name: {first: 'John', last: 'Carney'}},
        {isActive: false, age: 29, name: {first: 'Dick', last: 'Dunlap'}},
      ],
      fields: [
        {key: 'name', label: 'Person full name', sortable: true, sortDirection: 'desc'},
        {key: 'age', label: 'Person age', sortable: true, class: 'text-center'},
        {
          key: 'isActive',
          label: 'Is Active',
          formatter: (value: any, key: any, item: any) => (value ? 'Yes' : 'No'),
          sortable: true,
          sortByFormatted: true,
          filterByFormatted: true,
        },
        {key: 'actions', label: 'Actions'},
      ],
      totalRows: 1,
      currentPage: 1,
      busy: false,
      perPage: 5,
      pageOptions: [5, 10, 15, {value: 100, text: 'Show a lot'}],
      sortBy: '',
      sortDesc: false,
      sortDirection: 'asc',
      filter: '',
      filterOn: [],
      infoModal: {
        open: false,
        id: 'info-modal',
        title: '',
        content: '',
      },
    }
  },
  computed: {
    sortOptions() {
      // Create an options list from our fields
      return this.fields.filter((f) => f.sortable).map((f) => ({text: f.label, value: f.key}))
    },
  },
  mounted() {
    // Set the initial number of items
    this.totalRows = this.items.length
  },
  methods: {
    myProvider(
      ctx: BTableProviderContext,
      callback: (items: Array<TableItem>) => Promise<TableItem[] | undefined>
    ) {
      //ctx: {
      //   filter?: string
      //   currentPage?: number
      //   perPage?: number
      //   sortBy?: string
      //   sortDesc?: Booleanish
      // }
      const items: Array<TableItem> = [
        {isActive: false, age: 89, name: {first: 'Geneva', last: 'Wilson'}},
        {isActive: true, age: 38, name: {first: 'Jami', last: 'Carney'}},
        {isActive: false, age: 27, name: {first: 'Essie', last: 'Dunlap'}},
        {isActive: true, age: 40, name: {first: 'Thor', last: 'Macdonald'}},
      ]

      // console.log('Items set through a return', items);
      // // You can return an arroy of items
      // return items;

      // also you can return undefined and set the items through the callback function
      // callback(items)
      //   .then((items) => {
      //     console.log('Items set through callback function', items);
      //   })
      //   .finally(() => this.busy = false);
      // return undefined;

      // and finally you can return a promise that resolves an array of array OR undefined with setting the items through the callback function
      return new Promise<Array<TableItem> | undefined>((resolve, reject) => {
        setTimeout(() => {
          resolve(items)
        }, 3000)
      })
    },
    info(item: any, index: any, button: any) {
      this.infoModal.title = `Row index: ${index}`
      this.infoModal.content = JSON.stringify(item, null, 2)
      this.infoModal.open = true
    },
    resetInfoModal() {
      this.infoModal.title = ''
      this.infoModal.content = ''
    },
    onFiltered(filteredItems: any) {
      // Trigger pagination to update the number of buttons/pages due to filtering
      this.totalRows = filteredItems.length
      this.currentPage = 1
    },
  },
})
</script>

```


makroxyz:
I think it's better to stay close to BV2 ```api-url``` prop as possible, this also for an easy switch

devhus:
> I think it's better to stay close to BV2 `api-url` prop as possible, this also for an easy switch

Isn't api-prop just a prop that you pass to the table and the table passes it back to the provider function? if not please explain what is it for.

makroxyz:
> > I think it's better to stay close to BV2 `api-url` prop as possible, this also for an easy switch
> 
> Isn't api-prop just a prop that you pass to the table and the table passes it back to the provider function? if not please explain what is it for.

Yes it is. But there's no need to add ```provider``` property... ```items``` prop can be a also Function. 
It should be as close as possbile to the BV2... most people need to switch when BTable is ready

devhus:
> > > I think it's better to stay close to BV2 `api-url` prop as possible, this also for an easy switch
> > 
> > 
> > Isn't api-prop just a prop that you pass to the table and the table passes it back to the provider function? if not please explain what is it for.
> 
> Yes it is. But there's no need to add `provider` property... `items` prop can be a also Function. It should be as close as possbile to the BV2... most people need to switch when BTable is ready

Didn't seem like a big difference to me but much cleaner to work with a standalone prop for each feature.
I am just following the KISS rule -_0_-

VividLemon:
It was recently discussed with pi0, an owner of bootstrap-vue, that BTable has been a bloated component on their end, and that in a V3, it may be a good idea to scale back some of the features. It was also talked that maybe a more advanced table component can take over some of the features there, so perhaps it may be a good idea to start thinking about scaling some features back, and thinking about creating a separate component?

devhus:
> It was recently discussed with pi0, an owner of bootstrap-vue, that BTable has been a bloated component on their end, and that in a V3, it may be a good idea to scale back some of the features. It was also talked that maybe a more advanced table component can take over some of the features there, so perhaps it may be a good idea to start thinking about scaling some features back, and thinking about creating a separate component?

I really feel that BTable of BSVue3 is not even close to being bloated even with adding provider logic to it.
The table now covers the most important features of BSVue2 table which makes it easier to switch from BSV2 to BSV3 with minimum modifications. 
But... I agree that any more features should be implemented in another component.

VividLemon:
> I really feel that BTable of BSVue3 is not even close to being bloated even with adding provider logic to it. The table now covers the most important features of BSVue2 table which makes it easier to switch from BSV2 to BSV3 with minimum modifications. But... I agree that any more features should be implemented in another component.

Features like filtering is realistically something that a person could develop themselves with very minimal code. The point is, perhaps sometime soon it may be beneficial to put many of these features into a new component.

devhus:
> > I really feel that BTable of BSVue3 is not even close to being bloated even with adding provider logic to it. The table now covers the most important features of BSVue2 table which makes it easier to switch from BSV2 to BSV3 with minimum modifications. But... I agree that any more features should be implemented in another component.
> 
> Features like filtering is realistically something that a person could develop themselves with very minimal code. The point is, perhaps sometime soon it may be beneficial to put many of these features into a new component.

I agree but why bother to reinvent the wheel and write minimal code when you can use a component to do it for you ? 😆  
Anyway, Have you reviewed the PR?

aceofwings:
Good points. Remember that BTableSimple Is the dumb down component. I don't see a need to scale back features unless we want to consider this a plugin on its own or be designed to be enhanced with minimal effort via a pluggable interface. 

I'll give my thoughts on the actually code in a bit once I read it

VividLemon:
> I agree but why bother to reinvent the wheel and write minimal code when you can use a component to do it for you ? 😆 Anyway, Have you reviewed the PR?

I've been busy, no I have not.

makroxyz:
> I agree but why bother to reinvent the wheel and write minimal code when you can use a component to do it for you ? laughing Anyway, Have you reviewed the PR?

I totally agree. 
In my opinion BTable is a "key" component and should be as close as possible to the "old one".

packages/bootstrap-vue-3/src/components/BTable/BTable.vue
```diff
@@ -449,6 +464,33 @@ const handleRowSelection = (row: TableItem, index: number, shiftClicked = false)
...
+  if (response instanceof Promise) {
+    response
+      .then((items) => {
+        if (!Array.isArray(items)) return
+        itemHelper.updateInternalItems(items)
+      })
+      .finally(() => (internalBusyFlag.value = false))
+    return
+  }
+  itemHelper.updateInternalItems(response)
```
VividLemon, line 490: This function, and the clone deep function may be converted to async. Since deep cloning can be possibly quite long, depending on the scale of the items, it may be best to convert the deep clone function to async, and async this as well. Just a thought

I overhauled the utils, so cloneDeep util is different than what it used to be. It may be a good idea to sync your lib.

devhus: I hope i got this right.

packages/bootstrap-vue-3/src/components/BTable/BTable.vue
```diff
@@ -481,15 +481,17 @@ const callItemsProvider = () => {
   if (response === undefined) return
   if (response instanceof Promise) {
     response
-      .then((items) => {
+      .then(async (items) => {
         if (!Array.isArray(items)) return
-        itemHelper.updateInternalItems(items)
+        const internalItems = await itemHelper.updateInternalItems(items)
```
VividLemon, line 486: Don't mix await in callbacks. Either convert to a full async function, or continue doing .then chains. (They can be flakey when error handling, not that they throw errors

packages/bootstrap-vue-3/src/components/BTable/BTable.vue
VividLemon: > I hope that i've addressed this correctly

I do think there is an async related issue in there. But I will look a bit more when I have the time

packages/bootstrap-vue-3/src/components/BTable/BTable.vue
devhus: Not quite sure about this point since it's a simple variable set. however, I've removed the await operator from within the promise let me know if this is enough

VividLemon:
[changes.zip](https://github.com/cdmoro/bootstrap-vue-3/files/9592475/changes.zip)
I've attempted to resolve some of the asynchronous issues. That file contains the changes that I have made. (I probably could have found a way to push to the pr, but idk). 

At this point, I have to say that testing the component is on you. Prior to this PR, I was willing to go through and test the component when it was done. However, due the the additional complexity added here, I won't be able to test this thoroughly enough without a major time investment on getting familiar with the code. If you include the changes from above, and you believe everything works, I will (review it again, just in case as always) merge the PR. But, it is up to you now to write the test file for this component. 

Like I had said though, I am fully willing to drop support for this feature, and scale back some of the features on BTable. But, it's up to you, just know that I would heavily prefer _you_ to write the test file if you are committed to including this feature.

devhus:
> VividLemon

I have written some tests but note that i am not very good at writing tests so, you might need to take a deep look into the test file as well.

packages/bootstrap-vue-3/src/components/BTable/BTableContainer.vue
```diff
@@ -1,21 +1,24 @@
 <template>
-  <div v-if="responsive" :class="responsiveClass">
+  <table v-if="!responsive" :class="classes">
```
VividLemon, line 2: When you have items at the root like this, classes can be bound from the parent, and don't need a prop for it.

parent.vue
```vue
<template>
  <child class="foo" />
</template>
```
child
```vue
<template>
  <h1>foo</h1>
// still gets the classes, because it is at the root
</template>
<script></script>
```
No props needed. Should work automatically

There is a distinction however, that parent elements that have a _child_ that needs a class, sometimes require a prop. 

Like, BCaard.vue follows the idea of needing a prop, since the parent prop needs to pass it to the child prop

My point is that a prop for classes is a bit redundant when it's not a parent

VividLemon: Maybe it's different when there's a not a single root element though, idk

VividLemon: Never mind, I guess from the test it says that it is different when there's two roots. Lmao

Anyways, you can use type ClassValue to designate dynamic classes. Technically it's not a 100% perfect TS type, but it does the job.

VividLemon:
> @VividLemon I have written some tests but note that i am not very good at writing tests so, you might need to take a deep look into the test file as well.

It will be a work in progress. It's just a lot to test when there's a lot of different moving parts.

packages/bootstrap-vue-3/src/components/BTable/BTableContainer.vue
devhus: When ``responsive`` prop is true the root element will be a DIV, If we passed the table's classes to the container like this it will add the table's classes to the root DIV which is not correct

PR #684, opened 2022-09-19
Fix renamed bootstrap 5 utilities classes

bettysteger:
# Describe the PR

Dropdown **right** property was not working, because attaching the old utility CSS class: `dropdown-menu-right` instead of the new `dropdown-menu-end`.  Throughout the docs I also renamed the old utility CSS classes. 

## PR checklist

- [x] Bugfix :bug: - `fix(BDropdown)`
- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

- [x] Pull request title and all commits follow the [**Conventional Commits**](https://www.conventionalcommits.org/) naming convention **This is very important, as the `CHANGELOG` is generated from these messages, and determines the next version type**

### DIFF
--- a/apps/docs/docs/components/ButtonToolbar.md
+++ b/apps/docs/docs/components/ButtonToolbar.md
@@ -52,12 +52,12 @@
   <b-card>
     <div>
       <b-button-toolbar aria-label="Toolbar with button groups and input groups">
-        <b-button-group size="sm" class="mr-1">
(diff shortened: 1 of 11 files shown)


VividLemon:
If you want to go the extra mile, you can also rename some of the props to be their corresponding names. Such as replacing prop right with prop end, etc. Alternatively, you could also allow for _both_ and have both prop right and prop end, I know one component does that somewhere.

PR #687, opened 2022-09-19
test: tabs.spec.ts

VividLemon:
@devhus 
Your tests are at a good start, but need some work. So I'm going to use this to clear some misconceptions I think you may have in your tests. 

1. You are correct in that you are testing for the existence of a component. Your use of .exists() is correct. However, when you have already checked for the existence of a component, you should instead use `.get()` to get the component, rather than `.find()`..... `.get()` will throw an error when it does not exist, whereas a `.find()` will have a failing test. Basically, use `.find()` only when checking the existence, and `.get()` after you have tested the existence. 
2. In your first test, you had started testing the inner workings of the BTableContainer.vue file. You were checking that it's `tag is table by default`, `tag is div when prop responsive`. This is not up to the BTable.spec.ts test, it is it's own component test. Instead of testing that it is the correct tag, in BTable.spec.ts, you only need to test that you are passing in the correct values to the component props, it should be on it's own test to handle how the props are used. 
3. You have used `expect($table.classes().includes('border-danger')).toBe(true)`. This works, but there's a simpler alternative to do. `expect($table.classes()).toContain('border-danger')`. They essentially do the same thing, but it's nicer to use the testing utilities. 

Overall it's a good start. The basis of a component test is to test that things are being rendered correctly. That html elements and components are where they're supposed to be, that classes exist on those elements, that attributes exist. The internal implementations do not matter, if they did, they would be moved to a utility or a composable and would receive further testing. Also, events need to be tested, that they are being emitted when actions happen, etc. I can continue to try to write some tests for the table, but the biggest issue is ensuring that the asynchronous changes properly cause things to render correctly, that filter props render correctly, etc.

### DIFF
--- a/packages/bootstrap-vue-3/src/components/BTable/b-table.spec.ts
+++ b/packages/bootstrap-vue-3/src/components/BTable/b-table.spec.ts
@@ -3,6 +3,7 @@ import {afterEach, describe, expect, it} from 'vitest'
 import {ref} from 'vue'
 import {TableField, TableItem} from '../../types'
 import BTable from './BTable.vue'
+import BTableContainer from './BTableContainer.vue'
 
 const fields: Array<TableField> = [{key: 'name'}, {key: 'age'}]
 const items: Array<TableItem> = [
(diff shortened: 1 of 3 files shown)


devhus:
@VividLemon Thanks for all the tips, i am glad by your since these are my first test ive ever write.

VividLemon:
> @VividLemon Thanks for all the tips, i am glad by your review since these are my first test ive ever write.

It can be difficult to start writing tests, it can be confusing. But once you comprehend how they're  working, it's an immense help to all your applications. They are a huge help when making updates especially, cause then you'll in now if one of your small changes has broken everything

VividLemon:
To clarify point number 2, the reason why this is an issue is that it causes brittle tests. If the nested component makes a change that's necessary, both the parent and child need to fix their tests. Ultimately, it's not the parents job to test a child's code, and when it breaks, it breaks both tests. It's only the parents job to give the child the correct tools

PR #702, opened 2022-09-23
fixed BInputGroup awareness of its nested children borders radius

devhus:
solves https://github.com/cdmoro/bootstrap-vue-3/issues/470


VividLemon:
Prefer https://github.com/cdmoro/bootstrap-vue-3/pull/703 solution

Issue #742, opened 2022-10-11
BTable never hides the busy indicator when using provider function


diyan-gochev-at-ft-dot-com:
### Describe the bug

BTable never hides the busy indicator when using provider function

### Steps to reproduce the bug

1. Create BTable
2. Set provider function
3. Load fake data with Promise and setTimeout
4. Loader never hides.

### Expected behavior

After loading the data, the busy indicator should be hidden.

### Versions

**Libraries:**

    "bootstrap": "^5.2.0",
    "bootstrap-vue-3": "^0.3.11",

**Environment:**

- Device: PC
- OS: Windows 10
- Browser: Chrome
- Version 106.0.5249.91 (Official Build) (64-bit)

### Demo link

https://replit.com/@DiyanGochev1/BootstrapVueTableWithProvider#index.html

diyan-gochev-at-ft-dot-com:
@VividLemon Hi, I have attached demo link. Do we have any update?

VividLemon:
> @VividLemon Hi, I have attached a demo link. Do we have any update?

https://github.com/cdmoro/bootstrap-vue-3/issues/740#issuecomment-1273918021 same applies

diyan-gochev-at-ft-dot-com:
> > @VividLemon Hi, I have attached a demo link. Do we have any update?
> 
> [#740 (comment)](https://github.com/cdmoro/bootstrap-vue-3/issues/740#issuecomment-1273918021) same applies

I completely understand, but it seems easy to fix and the impact seem to be huge.

VividLemon:
> I completely understand, but it seems easy to fix and the impact seem to be huge.

I'm guessing the `internalBusy` ref used in the component is not being set back to false. If you happen to find where it is, you are welcome to submit a PR.

diyan-gochev-at-ft-dot-com:
I am using bootstrap-vue-3.es.js which does not have un-minified version to troubleshoot/debug the source code. The project is not TS, it's pure JS.

diyan-gochev-at-ft-dot-com:
@VividLemon 
<img width="484" alt="image" src="https://user-images.githubusercontent.com/97102927/195164612-405d4352-736c-4185-8cb0-a2d74285d93c.png">

I think this is the problem, the should be the opposite, without '!' (exclamation mark), but check again. This is my first read of the source code.

diyan-gochev-at-ft-dot-com:
@VividLemon PR has been created. Please review the changes.

[...]

PR #744, opened 2022-10-11
fix(BTable): [#742] BTable with provider never hides the busy loader

diyan-gochev-at-ft-dot-com:
# Describe the PR

BTable with provider never hides the busy loader

## PR checklist

- [X] Bugfix :bug: - `fix(#742 )`


VividLemon:
That's


Hillarious

PR #755, opened 2022-10-20
BTable event filter, BFormInput highlighting animation

devhus:
- fix(BTable): Enchanted btable event filtering to prevent the unnecessary need to add event.modifiers to children's events - fixes https://github.com/cdmoro/bootstrap-vue-3/issues/723 issue
- feat(BFormInput): added highlight method to active a simple highlight animation for a few seconds - demo: https://www.loom.com/share/ca8d77063a7741c0a8c0e8d559ee8220


VividLemon:
I organized a bunch of files. Please update your code to fit in the correct organization pattern. I forget the exact pattern I conformed to.

devhus:
> I organized a bunch of files. Please update your code to fit in the correct organization pattern. I forget the exact pattern I conformed to. Something like: imports -> props -> emits -> useSlots & attrs, inject, constants, other composables, useBooleanishes composables, refs, computeds, methods, event methods (view BModal, methods explicitly used for events),watchers, lifecycle hooks. (The end ones sometimes get weird since sometimes ts will complain about usage before defined, so it's "fine" to change the order a bit)

i think it's now in the correct order, hopefully

VividLemon:
> * feat(BFormInput): added highlight method to active a simple highlight animation for a few seconds

Add an optional prop to this to disable the affect.

VividLemon:
Also in my last merged PR I removed Bootstrapjs from BModal. I broke the functionality for lazy loading modals, so I removed it. The problem was, the animation was not being correctly displayed as the elements to display the animation to were being removed before the animation was done. Causing it to look similar to having prop "noFade" on, but only when 'leaving'. If you could possibly look at implementing lazy loading back into the modal, that would be awesome.

devhus:
> > * feat(BFormInput): added highlight method to active a simple highlight animation for a few seconds
> 
> Add an optional prop to this to disable the affect.
> 
> psuedo
> 
> ```ts
>     interface Props { noHighlight?: Booleanish }
>     withDefaults(defineProps<Props>(), { noHighlight: false }
>     const noHighlightBoolean = useBooleanish(toRef(props, 'noHighlight'))
>     const highlight = () => {
>       if (isHighlighted.value === true || noHighlightBoolean.value) return
>       isHighlighted.value = true
>       setTimeout(() => {
>         isHighlighted.value = false
>       }, 2000)
>     }
> ```
> 
> Also, I'm not sure where highlight function is used. It never seems to be called? I'm not sure how it's working for you. It also may be better to have this not be the _default_ behavior, as it's not default bootstrap behavior, just a thought. Basically the same code I wrote, but in reverse. prop highlightOnHover?: Booleanish, default: false, if(!highlightOnHoverBoolean.value) ...

No need for a no-highlight prop, you highlight the input when you need to by calling the exposed method "highlight"
example: 
```vue
<b-form-input ref="myInputRef " ...>

const myInputRef = ref()
myInputRef.highlight();
```

devhus:
> Also in my last merged PR I removed Bootstrapjs from BModal. I broke the functionality for lazy loading modals, so I removed it. The problem was, the animation was not being correctly displayed as the elements to display the animation to were being removed before the animation was done. Causing it to look similar to having prop "noFade" on, but only when 'leaving'. If you could possibly look at implementing lazy loading back into the modal, that would be awesome.

i will take a look as soon as i am free

devhus:
@VividLemon There is A LOT of stuff broken in BModal atm, z-indexing, nested modals, lazy rendering, and might have more hidden bugs, i am still diagnosing it

PR #763, opened 2022-10-23
CSS possibilities correction for BTable sort-icon (Very important for UX)

devhus:
So, I did an oppise when creating the logic behind BTable -> Sort Icon styling where the sort icon elements are removed from the dom until the column's header is sorted, Which prevents a lot of styling possibilities for these icons.
Now the icon elements will be always rendered in the dom without appearing until the header is sorted.
This will allow having one global CSS code for all your website tables instead of forcing you to add slot-template inside each table to make simple customizations to the icons like this one
image

### DIFF
--- a/packages/bootstrap-vue-3/src/components/BModal.vue
+++ b/packages/bootstrap-vue-3/src/components/BModal.vue
@@ -23,7 +23,15 @@
         @keyup.esc="onEsc"
       >
         <div class="modal-dialog" :class="modalDialogClasses">
-          <div class="modal-content" :class="contentClass">
+          <div
+            v-if="
+              !lazyBoolean ||
+              (lazyBoolean && lazyLoadCompleted) ||
+              (lazyBoolean && modelValueBoolean === true)
+            "
+            class="modal-content"
+            :class="contentClass"
+          >
             <div v-if="!hideHeaderBoolean" class="modal-header" :class="headerClasses">
               <slot name="header">
                 <component
@@ -99,7 +107,7 @@
 
 <script setup lang="ts">
 // import type {BModalEmits, BModalProps} from '../types/components'
-import {computed, ref, toRef, useSlots, watch} from 'vue'
+import {computed, nextTick, ref, toRef, useSlots, watch} from 'vue'
 import {useBooleanish, useId} from '../composables'
 import type {Booleanish, ClassValue, ColorVariant, InputSize} from '../types'
 import {isEmptySlot} from '../utils'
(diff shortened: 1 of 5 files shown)


amery:
as of 94284b4 `@typescript-eslint/no-this-alias` on `src/utils/dom.ts:220` remains but it's no longer exploding regarding `Element`

FYI @VividLemon

amery:
it's worth mentioning eslint also _fixed_ style on these files
```
packages/bootstrap-vue-3/src/components/BFormInput/BFormInput.vue
packages/bootstrap-vue-3/src/constants/env.ts
packages/bootstrap-vue-3/src/utils/dom.ts
```

devhus:
> as of [94284b4](https://github.com/cdmoro/bootstrap-vue-3/commit/94284b4d6e9a7031b5ae09b8056a8e8189e62082) `@typescript-eslint/no-this-alias` on `src/utils/dom.ts:220` remains but it's no longer exploding regarding `Element`
> 
> FYI @VividLemon

So now we just need to figure out how to silent ESlint for that part, right?

amery:
> > as of [94284b4](https://github.com/cdmoro/bootstrap-vue-3/commit/94284b4d6e9a7031b5ae09b8056a8e8189e62082) `@typescript-eslint/no-this-alias` on `src/utils/dom.ts:220` remains but it's no longer exploding regarding `Element`
> > FYI @VividLemon
> 
> So now we just need to figure out how to silent ESlint for that part, right?

I don't believe that to be a solution, but it's @VividLemon decision.

this commit silences it, https://github.com/amery/bootstrap-vue-3-ssr/commit/e4aca21b0e143fa8d0c81896ee7367d13ac8651f

devhus:
@VividLemon It's now up to you, should we silence it or refactor the code to somehow work without ``this`` ?
Note that the function is not really called unless the browser is not supporting the "closest" function, I am also not very sure that the ``this`` operator can be removed since it references the Element element which is only known as ``this``

VividLemon:
> @VividLemon It's now up to you, should we silence it or refactor the code to somehow work without `this` ? Note that the function is not really called unless the browser is not supporting the "closest" function, I am also not very sure that the `this` operator can be removed since it references the Element element which is only known as `this`

Im going to need to do more research on how things are functioning.

amery:
> @VividLemon It's now up to you, should we silence it or refactor the code to somehow work without `this` ? Note that the function is not really called unless the browser is not supporting the "closest" function, I am also not very sure that the `this` operator can be removed since it references the Element element which is only known as `this`

I don't know much about the dark arts of front end development but I assume you can call a function using `this` as an argument that can in turn return the closure function using it

devhus:
@amery example ?

amery:
@devhus @VividLemon please excuse my javascript illiteracy but... why not this?
```diff
--- a/packages/bootstrap-vue-3/src/utils/dom.ts
+++ b/packages/bootstrap-vue-3/src/utils/dom.ts
@@ -216,8 +216,7 @@ export const matches = (el: Element, selector: string) =>
 /* istanbul ignore next */
 export const closestEl =
   ELEMENT_PROTO?.closest ||
-  function (this: Element, sel: string) {
-    let el = this
+  function (el: Element, sel: string) {
     if (!el) return null
     do {
       // Use our "patched" matches function
```

devhus:
@amery I don't think it will work, since the `this` argument is not really an argument it's more like a type declaration for `this` within that function.
Since the function is used as a prototype replacement it will be called by javascript in a way that references the ``Element`` as ``this`` within the defined function and so ``this`` is the only option to really get the ``Element`` we are working with. That's why my first way to go was to since ESlint.

This logic is used everywhere and in a lot of libraries
https://github.com/jonathantneal/closest/blob/master/src/index.js
https://github.com/topics/closest

devhus:
@VividLemon I've committed an update to silence the Eslint trigger so you can merge the PR if you want to, I will not be here in the next few days so be free to merge or search for another way to do it

Issue #790, opened 2022-11-04
How to extend ColorVariant, ButtonVariant, etc.?


esl51:
**What you are trying to do**
I can extend bootstrap using scss. For example, add my own colors, sizes, etc.
But I can't use them.

**What you have already tried**
```vue
<b-button variant="purple">Purple button</b-button>
```

**Any error messages you are getting**
TS2322: Type '"purple"' is not assignable to type 'ButtonVariant | undefined'.

**Environment:**
 - OS: MacOS Monterey 12.6
 - Version: 0.3.12

VividLemon:
Fundamentally the code would need to be rewritten on our end to support that. One would need to somehow override the ColorVariant interface. If it was directly modified in the dist, then it would be written over on subsequent installs. Not very intuitive. 

You could attempt to typecast - `'purple' as unknown as ColorVariant`

Doing as you reference would be a nice thing to support, but would require hefty discussions on how to best achieve it.

aceofwings:
Please note that the prop type only applies the variant class to the given component . Make sure to follow color guide [here](https://getbootstrap.com/docs/5.0/customize/color/). You can also look at bootswatch for theming. This issue is beyond the scope of bootstrapvue.

esl51:
> This issue is beyond the scope of bootstrapvue.

I can create [my own color maps](https://getbootstrap.com/docs/5.0/customize/sass/#add-to-map) with bootstrap

```scss
// Create your own map
$custom-colors: (
  "purple": $purple-500;
);

// Merge the maps
$theme-colors: map-merge($theme-colors, $custom-colors);
```

and then use them:

```html
<button class="btn btn-purple">Purple button</button>
```

but can't use them with bootstrap-vue.

I also can't solve it with class because BButton has default 'secondary' variant.

```vue
<b-button class="btn-purple">Purple button</b-button>
```
turns into this:
```html
<button class="btn btn-secondary btn-purple">Purple button</b-button>
```

Thereby, I think this issue is in the scope of bootstrap-vue.

aceofwings:
As @VividLemon lemon mentioned this will be considered an enhancement.. However I think its better to address this in the docs on how to extend our library typing namespace then to write generic code to address this issue and thus use your types as needed.

VividLemon:
> As @VividLemon lemon mentioned this will be considered an enhancement.. However I think its better to address this in the docs on how to extend our library typing namespace then to write generic code to address this issue and thus use your types as needed.

[...]

PR #791, opened 2022-11-04
Ability to extend sizes and variants (augmenting interfaces)

esl51:
BEGIN_COMMIT_OVERRIDE
feat: add the ability to extend certain built in types
docs: add info on how to extend types
END_COMMIT_OVERRIDE

PR for issue #790

Now we will be able to augment new interfaces:

**shims-bootstrap-vue-3.d.ts:**
```ts
import 'bootstrap-vue-3'

declare module 'bootstrap-vue-3/dist/types' {
  export interface BaseColorVariant {
    'purple': unknown
  }
  export interface BaseButtonVariant {
    'outline-purple': unknown
  }
  export interface BaseSize {
    'xl': unknown
  }
}
```
New values can be used:
```vue
<b-button variant="purple" size="xl">Extra large purple button</b-button>
<b-button variant="outline-purple">Outline purple button</b-button>
```


VividLemon:
I have two major questions on this.

First, how does `export interface BaseColorVariant {` syntax work? It would almost make more sense that the code you posted overwrites the bases... But I've never use it like that. 

Second, is it a possibility to use `type` instead of `interface`? Is it possible to augment the types in the same style of syntax you have made, without the weirdness of making the interface?

esl51:
> First, how does `export interface BaseColorVariant {` syntax work? It would almost make more sense that the code you posted overwrites the bases... But I've never use it like that.

First we define an interface with basic values (`properties`). It is only needed for later augmentation. We don't use the type of properties, so `unknown` type (perhaps there is a more elegant solution, but I can't think of one).
Then we define ButtonVariant type, which will be union type by interface properties (`keyof`).

> Second, is it a possibility to use `type` instead of `interface`? Is it possible to augment the types in the same style of syntax you have made, without the weirdness of making the interface?

No, augmentation of types is impossible. So I had to use interfaces to do it.

> Third, if the components themselves use ColorVariant, and you are modifying BaseColorVariant, how does that ultimately affect the components?

The BaseButtonVariant interface extends BaseColorVariant (and therefore includes all its properties). When BaseColorVariant is augmented, the properties will be added to BaseButtonVariant and then to ButtonVariant as well.

VividLemon:
Would you be willing to make an update to the docs, somewhere, speaking about how to go about using this? It doesn't need to be very good, just a good enough placeholder that I don't forget how to use it later.

esl51:
> Would you be willing to make an update to the docs, somewhere, speaking about how to go about using this? It doesn't need to be very good, just a good enough placeholder that I don't forget how to use it later.

Okay, I'll do it soon.

sfe-efficy:
If we try to use a vue typescript checker like "vue-tsc" this code is not working property.
For example the BButton.vue.d.ts is taken from: node_modules/bootstrap-vue-3/dist/components/BButton/BButton.vue.d.ts

And the the definition for variant has the following:

image

So, the vue-tsc is going to use the types inside the original module instead of using our custom definition.
Maybe is a lack of configuration in my side for vue-tsc.

In tsconfig.json I'm including types:

image

I have a file defined in types/bootstrap-vue3/shims-bootstrap-vue-3.d.ts

With the content:

```ts
import "bootstrap-vue-3";

declare module "bootstrap-vue-3" {
    export interface BaseButtonVariant {
        orange: unknown;
    }
}
```

This is the error reported by vue-tsc:

```

src/components/base-button/base-button.vue:13:15 - error TS2322: Type '"orange"' is not assignable to type 'keyof BaseButtonVariant | undefined'.

13               variant="orange"
                 ~~~~~~~

  node_modules/bootstrap-vue-3/dist/components/BButton/BButton.vue.d.ts:375:5
    375     variant: keyof import("../../types").BaseButtonVariant;
            ~~~~~~~
    The expected type comes from property 'variant' which is declared here on type 'ComponentProps<DefineComponent<{ active: { type: PropType<Booleanish>; default: boolean; }; disabled: { type: PropType<Booleanish>; default: bool
ean; }; href: { type: StringConstructor; required: false; }; ... 19 more ...; to: { ...; }; }, ... 10 more ..., { ...; }>>'


Found 1 error in src/components/base-button/base-button.vue:13
```

And this is the example:

image

also the IDE is complaining.

esl51:
> If we try to use a vue typescript checker like "vue-tsc" this code is not working property. 
Try to add this to your tsconfig:

```
files": [
    "./src/shims-bootstrap-vue-3.d.ts"
  ]
```

esl51:
And you have an error in shims file on this line:

```ts
declare module 'bootstrap-vue-3/dist/types' {
```

sfe-efficy:
> ```ts
>  export interface BaseButtonVariant {
>     'default'
>     'ghost-danger'
>   }
> ```

Thanks, the only problem was related with this.
I was trying to override a simple type (thinking if it was possible don't create an interface to avoid the redundant code), and I guess I was doing some invalid mix.

I don't really need to add the "file" section into tsconfig.json

PR #794, opened 2022-11-06
Review Aligment uses and definitions

sfe-efficy:
BEGIN_COMMIT_OVERRIDE
docs: Update definition of Alignment and json files.
docs: Add details about how to use directives, v-b-tooltip.
fix(BCard): Set correct align types.
fix(BNav): Set correct align types.
fix(BNavbarNav): Set correct align types.
feat(BPagination): Extend justify content types for between, around and evenly
fix(BPagination): Set correct align types.
fix(BTabs): Set correct align types.
fix(BRow): Set correct align types.
fix(BTooltip): Error toString if value is undefined.
END_COMMIT_OVERRIDE

# Describe the PR

In external projects, using an IDE like IntelliJ is taking problems to go directly to each definition of the Alignment namespaces, because we have a conflict naming for the type "Alignment" and the namespace "Alignment".

During the check also I see another problems like: definining a namespace as a type in props. No correct type definitions for text alignment, no correct definitions for justify content aligment, mixing "fill" type from Navigation as a justifify content alignment.

## PR checklist

- [x] Feature - `feat(...)`
- [ ] Other (please describe)


VividLemon:
How about instead we refactor this to remove the namespace. Essentially, just remove the namespace and append Alignment to the keys. VerticalAlignment, HorizontalAlignment, etc.

VividLemon:
> While we were trying to use the newest version, we saw that we have a problem using the directive tooltip (bad configuration with who is the root element with the assigned directive like v-b-tooltip.top.hover and the title is missing.
> 
> So, no value defined for the directive and no title, and we are trying to use an undefined value with the method "toString". Maybe we can create a warning in that case, that allows the developers to review when we try to use the directive, but it wasn't possible to retrieve the title.
> 
> With this change the error is silent, a warning is good, but no generating an error. what do you think?

Not sure what you mean

sfe-efficy:
> > While we were trying to use the newest version, we saw that we have a problem using the directive tooltip (bad configuration with who is the root element with the assigned directive like v-b-tooltip.top.hover and the title is missing.
> > So, no value defined for the directive and no title, and we are trying to use an undefined value with the method "toString". Maybe we can create a warning in that case, that allows the developers to review when we try to use the directive, but it wasn't possible to retrieve the title.
> > With this change the error is silent, a warning is good, but no generating an error. what do you think?
> 
> Not sure what you mean

I mean, since this PR https://github.com/cdmoro/bootstrap-vue-3/pull/647/files the directive related for Tooltip is error prune to having this error:

image

So, the code is not taking care of the possibility of a bad configuration, and it's going to break other projects. With this PR the error is gone, because I'm taking care of the possibility of returning an undefined value, but...

I guess to help other developers is good to display a warning message to say something like: "Please, review tooltip directive usage. Some uses are not defining a title in root component or a value like "v-b-tooltip=xxx"  to define a title".

And also, I guess there is a lack of documentation about using this directive.

VividLemon:
Ultimately I don't think it's a good use of time since eventually I would like to go through and rebuild the directives. But you're free to implement that change if you want. An error message, and perhaps a doc update if you'd like.

VividLemon:
Why did the tests need to change?

sfe-efficy:
> Why did the tests need to change? Specifically im referring to card.spec.ts

Because "fill" type wasn't a valid type for text horizontal alignment.
https://github.com/cdmoro/bootstrap-vue-3/pull/794/commits/14d0f1ea31edb11ebc8cef728f045e4da942b605

In Bootstrap 5 doesn't exist "text-fill" as a type alginment.

https://getbootstrap.com/docs/5.2/utilities/text/#text-alignment
https://getbootstrap.com/docs/5.2/components/card/#text-alignment

So, the text was trying to validate that we generate a string with the prefix "text-" with the given value, but "fill" in this MR is not a valid one.

Also, in bootstrap-vue (2) you can find the same:
https://bootstrap-vue.org/docs/components/card#horizontal-card-layout

PR #801, opened 2022-11-10
Avoid default values in BButton for size and variant

sfe-efficy:
BEGIN_COMMIT_OVERRIDE
docs: Update BButton to say no variant is apply by default.
fix(BButton): Allow using a BButton with no variant.
END_COMMIT_OVERRIDE

# Describe the PR

Setting as default values a variant as secondary sometimes could be annoying, generating other projects to set an invalid type variant like "empty string" to avoid applying by default the secondary variant. And if they use a checker code like "vue-tsc" that allow to check the templates with strict, it's going to report errors.

The default value for size is redundant and it can avoid.

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

- [X] Pull request title and all commits follow the [**Conventional Commits**](https://www.conventionalcommits.org/) convention or have an [**override**](https://github.com/googleapis/release-please#how-can-i-fix-release-notes) in this pull request body **This is very important, as the `CHANGELOG` is generated from these messages, and determines the next version type. Pull requests that do not follow conventional commits or do not have an override will be denied**


VividLemon:
This is a topic that I've wanted to get to for awhile. Probably like 4th of the 5 major steps required for a v1.0. If you would like to continue this work, extend this same thought to all of the other components. I knew when I was rewriting all of the components to script setup, that many of the components were going to be incorrect. However, the issue was that the components were originally written with https://eslint.vuejs.org/rules/require-default-prop.html turned off, which made the rewrite difficult.

VividLemon:
> Originally, the thought was that many components could keep their defaults, where "undefined" would give them their defaults. However, the idea was to introduce "null" as a possibility, to create the ability to remove the item outright. But again, BV was written a bit weirdly, along with Vue 2, in the regards of the two being pretty much identical, rather then having separate uses

It kind of dawned on me that part of the issue with trying to do this is there exists a Vue issue where if you try to do "complex types", Vue will break. For example, `prop?: string | number | boolean` is allowed. But `prop?: string | ColorVariant` is not allowed. I'd suspect the same is true for `prop?: ColorVariant | null` would equally break. Though, that should be fixed in the long awaited next minor version of Vue.

VividLemon:
Also, I'm not really sure why you're wanting to make the various types namespaced into `ColorVariant.TextColorVariant`. I prefer just importing `TextColorVariant`, `ColorVariant`, etc. The same goes for Alignment. Stylistically, I'm not a fan. But ultimately it's a style choice and if there's a good merit that you can think of, I'm not opposed as there isn't a "downside".

sfe-efficy:
> > Originally, the thought was that many components could keep their defaults, where "undefined" would give them their defaults. However, the idea was to introduce "null" as a possibility, to create the ability to remove the item outright. But again, BV was written a bit weirdly, along with Vue 2, in the regards of the two being pretty much identical, rather then having separate uses
> 
> It kind of dawned on me that part of the issue with trying to do this is there exists a Vue issue where if you try to do "complex types", Vue will break. For example, `prop?: string | number | boolean` is allowed. But `prop?: string | ColorVariant` is not allowed. I'd suspect the same is true for `prop?: ColorVariant | null` would equally break. Though, that should be fixed in the long awaited next minor version of Vue.

Hi! when you say that, is because are you getting errors in building time or in runtime?
I don't see problem of that in our project using an interface to define a closed list of strings with optinal (undefined) possibility.
I saw also that if we define as a type undefined value, the optinal value is discard if we set an "undefined" value from the parent component, at least in the declaration typescript code:

image

sfe-efficy:
> Also, I'm not really sure why you're wanting to make the various types namespaced into `ColorVariant.TextColorVariant`. I prefer just importing `TextColorVariant`, `ColorVariant`, etc. The same goes for Alignment. Stylistically, I'm not a fan. But ultimately it's a style choice and if there's a good merit that you can think of, I'm not opposed as there isn't a "downside".

Simplily it was because I was keeping the "as XX" import in the index.ts file, I've just removed.

sfe-efficy:
Sorry, I close this MR and redo it in another branch, with the rebase the commit was positioned in a wrong place

PR #804, opened 2022-11-12
Feature - Spin button

aceofwings:
# Describe the PR

First Functioning Spin Button Ported from the BootstrapVue Library.

Tests have also been ported and adjusted to meet library changes.

Finally some docs have been added(not in full)
## Small replication

## PR checklist

- [x] Feature - `feat(...)`
- [ ] Other (please describe)


apps/docs/docs/component-references/SpinButton.json
```diff
@@ -0,0 +1,19 @@
+{
+  "$schema": "./schema.json",
+  "name": "@bootstrap-vue-3/SpinButton",
```
VividLemon, line 3:
```suggestion
  "name": "@bootstrap-vue-3/spin-button",
```

apps/docs/docs/component-references/SpinButton.json
```diff
@@ -0,0 +1,19 @@
...
+  "version": "1.0.0",
+  "meta": {
+    "title": "SpinButton",
```
VividLemon, line 6:
```suggestion
    "title": "Spin Button",
```

apps/docs/docs/component-references/SpinButton.json
```diff
@@ -0,0 +1,19 @@
...
+        "component": "BFormSpinButton",
+        "props": [
+        ],
+        "emits": [],
+        "slots": [
+        ]
+      }
+    ]
+  }
+}
```
VividLemon, line 19:
```suggestion
}

```

packages/bootstrap-vue-3/src/utils/array.ts
```diff
@@ -0,0 +1 @@
+export const arrayIncludes = (array: Array<unknown>, value: unknown) => array.indexOf(value) !== -1
```
VividLemon, line 1: Replace instance of arrayIncludes with Array.includes, then remove this file

packages/bootstrap-vue-3/src/utils/dom.ts
```diff
@@ -1,17 +1,17 @@
 import {Comment, Slot, VNode} from 'vue'
-import {DOCUMENT, HAS_ELEMENT_SUPPORT} from '../constants/env'
+import {DOCUMENT} from '../constants/env'
 import {AnimationFrame} from '../types/safeTypes'
 import {HAS_WINDOW_SUPPORT} from './env'
 import {toString} from './stringUtils'
 
-const ELEMENT_PROTO = HAS_ELEMENT_SUPPORT ? Element.prototype : undefined
```
VividLemon, line 7: I recall there was a reason why this was added, it was added somewhat recently. Why has it been removed?

packages/bootstrap-vue-3/src/utils/inspect.ts
```diff
@@ -19,3 +19,11 @@ export const isNumeric = (value: unknown): boolean => /^[0-9]*\.?[0-9]+$/.test(S
...
 export const isPlainObject = (obj: unknown): obj is Record<PropertyKey, unknown> =>
   Object.prototype.toString.call(obj) === '[object Object]'
+
+/**
+ * Strict object type check. Only returns true for plain JavaScript objects
+ *
+ * @param obj
+ * @returns
+ */
+export const isNull = (value: unknown): Boolean => value === null
```
VividLemon, line 29: TypeScript magic will then tell you what the value is (type predicates)
```suggestion
export const isNull = (value: unknown): value is null => value === null
```

packages/bootstrap-vue-3/src/components/BFormSpinButton/BFormSpinButton.vue
```diff
@@ -0,0 +1,526 @@
...
+      event: (KeyboardEvent & {button: undefined}) | (MouseEvent & {code: undefined})
+    ) => {
+      const {code, altKey, ctrlKey, metaKey} = event
+
+      /* istanbul ignore if */
+      if (props.disabled || props.readonly || altKey || ctrlKey || metaKey) {
+        return
+      }
+
+      if (arrayIncludes(KEY_CODES, code)) {
```
VividLemon, line 260:
```suggestion
      if (KEY_CODES.includes(code)) {
```

packages/bootstrap-vue-3/src/components/BFormSpinButton/BFormSpinButton.vue
```diff
@@ -0,0 +1,526 @@
...
+        // https://w3c.github.io/aria-practices/#spinbutton
+        stopEvent(event, {propagation: false})
+        /* istanbul ignore if */
+        if ($_keyIsDown) {
+          // Keypress is already in progress
+          return
+        }
+
+        resetTimers()
+        if (arrayIncludes([CODE_UP, CODE_DOWN], code)) {
```
VividLemon, line 270:
```suggestion
        if ([CODE_UP, CODE_DOWN].includes(code)) {
```

packages/bootstrap-vue-3/src/components/BFormSpinButton/BFormSpinButton.vue
```diff
@@ -0,0 +1,526 @@
...
+    const onKeyup = (event: KeyboardEvent) => {
+      // Emit a change event when the keyup happens
+
+      const {code, altKey, ctrlKey, metaKey} = event
+
+      /* istanbul ignore if */
+      if (props.disabled || props.readonly || altKey || ctrlKey || metaKey) {
+        return
+      }
+      if (arrayIncludes(KEY_CODES, code)) {
```
VividLemon, line 302:
```suggestion
      if (KEY_CODES.includes(code)) {
```

packages/bootstrap-vue-3/src/utils/event.ts
```diff
@@ -0,0 +1,66 @@
...
+// Utility method to add/remove a event listener based on first argument (boolean)
+// It passes all other arguments to the `eventOn()` or `eventOff` method
+export const eventOnOff = (on: Boolean, eventParams: Parameters<typeof eventOff>) => {
+  const method = on ? eventOn : eventOff
+  method(...eventParams)
+}
+
+//checks to see if keyboard event
+export const isKeyBoardEvent = (event: Event): event is KeyboardEvent => {
+  return arrayIncludes(['keydown', 'keypress', 'keyup'], event.type)
```
VividLemon, line 50:
```suggestion
  return ['keydown', 'keypress', 'keyup'].includes(event.type)
```

packages/bootstrap-vue-3/src/utils/event.ts
```diff
@@ -0,0 +1,66 @@
...
+
+// Utility method to add/remove a event listener based on first argument (boolean)
+// It passes all other arguments to the `eventOn()` or `eventOff` method
+export const eventOnOff = (on: Boolean, eventParams: Parameters<typeof eventOff>) => {
+  const method = on ? eventOn : eventOff
+  method(...eventParams)
+}
+
+//checks to see if keyboard event
+export const isKeyBoardEvent = (event: Event): event is KeyboardEvent => {
```
VividLemon, line 49: This file should be linted. This should be warning for arrow body style, ex
```ts
// incorrect
const val = () => {
    return true
}
// correct
const val = () => true
```

packages/bootstrap-vue-3/src/utils/locale.ts
```diff
@@ -0,0 +1,30 @@
...
+  'yi',
+].map((locale) => locale.toLowerCase())
+
+// Returns true if the locale is RTL
+export const isLocaleRTL = (locale: string) => {
+  // Determines if the locale is RTL (only single locale supported)
+  const parts = toString(locale).toLowerCase().replace(RX_STRIP_LOCALE_MODS, '').split('-')
+  const locale1 = parts.slice(0, 2).join('-')
+  const locale2 = parts[0]
+  return arrayIncludes(RTL_LANGS, locale1) || arrayIncludes(RTL_LANGS, locale2)
```
VividLemon, line 29:
```suggestion
  return RTL_LANGS.includes(locale1) || RTL_LANGS.includes(locale2)
```

VividLemon:
The build steps been acting up lately. Wonder what's going on with it

aceofwings:
I've noticed that with the icons. Ill take a quick look but everything else looks good?

aceofwings:
something wrong with the playground...

aceofwings:
So apparently when I import SpinButton into components/index.ts, it  breaks something?

VividLemon:
> I've noticed that with the icons. Ill take a quick look but everything else looks good?

I forgot I wanted to inform you that I made some changes to the icons. Replaced the generate script with a ts version that used ts node. However, the script makes reference to an "icons" directory that doesn't exist. 

> So apparently when I import SpinButton into components/index.ts, it breaks something?

I'll take a further look myself eventually into why it's breaking. Just a note, I plan to do a mass update to the dependencies. There's some feature limitations that can be fixed by updating some dependencies

aceofwings:
Yeah I went through the build process and it has to do with importing the SpinButtons 
`import {BIconDash, BIconPlus} from '../../../../bootstrap-vue-3-icons/src/components/icons/icons'`
The types used in these components cannot be resolved

VividLemon:
> Yeah I went through the build process and it has to do with importing the SpinButtons `import {BIconDash, BIconPlus} from '../../../../bootstrap-vue-3-icons/src/components/icons/icons'` The types used in these components cannot be resolved

I was literally just about to send a message about that. Avoid importing items from other packages, especially when they're not in the package dependencies. I'd recommend instead to just make a small recreation of the icon. As far as i know, they can simply be just svgs

aceofwings:
Hmmm would it be ok go mark bootstrap-vue-3 as a dependency, externalize the bootstrap-vue-icon package and then import 
as
`
import {BIconDash, BIconPlus} from 'bootstrap-vue-3-icons'
`
further more remove bootstrap-vue-3 as a dependency from bootstrap-vue-icons to prevent circles?

VividLemon:
> Hmmm would it be ok go mark bootstrap-vue-3 as a dependency, externalize the bootstrap-vue-icon package and then import as `import {BIconDash, BIconPlus} from 'bootstrap-vue-3-icons'` further more remove bootstrap-vue-3 as a dependency from bootstrap-vue-icons to prevent circles?

It's probably best to avoid linking them at all. After all, the icons are removed for a reason

aceofwings:
> > Hmmm would it be ok go mark bootstrap-vue-3 as a dependency, externalize the bootstrap-vue-icon package and then import as `import {BIconDash, BIconPlus} from 'bootstrap-vue-3-icons'` further more remove bootstrap-vue-3 as a dependency from bootstrap-vue-icons to prevent circles?
> 
> It's probably best to avoid linking them at all. After all, the icons are removed for a reason. Linking them seems like a hazard. Plus, I don't really know how importing direct vue files in different packages will affect things

Good news that the setup will not bloat sizes unless we do not decide to externalize the library. Most libraries will use packages as a dependency? It will be similar to how we currently deal with bootstrap's source files. 

Also seems like the update to the icons broke the rendering and is now encapsulating the <path> element with in a string within the attribute. Is there a different way of using icons now since your update?

VividLemon:
> > > Hmmm would it be ok go mark bootstrap-vue-3 as a dependency, externalize the bootstrap-vue-icon package and then import as `import {BIconDash, BIconPlus} from 'bootstrap-vue-3-icons'` further more remove bootstrap-vue-3 as a dependency from bootstrap-vue-icons to prevent circles?
> > 
> > 
> > It's probably best to avoid linking them at all. After all, the icons are removed for a reason. Linking them seems like a hazard. Plus, I don't really know how importing direct vue files in different packages will affect things
> 
> Good news that the setup will not bloat sizes unless we do not decide to externalize the library. Most libraries will use packages as a dependency? It will be similar to how we currently deal with bootstrap's source files.
> 
> Also seems like the update to icons broke the rendering and is now encapsulating the element in a string assigned to the content prop. Seems like it is been handled as an attribute of the element as well. Is there a different way of using icons now since your update?

Right, another thing that I did to the icons was removing the content prop. I made it a slot instead. Pretty sure it's bad practice to be sending vnodes through props

aceofwings:
You will want to update make-icon helper function to reflect the changes made to utilize direct import of certain icons.

VividLemon:
> You will want to update make-icon helper function to reflect the changes made to utilize direct import of certain icons.

Regardless, as I noted,  generate.ts makes reference to a directory that doesn't exist. You mind looking into that and I'll look into fixing the helper tomorrow

aceofwings:
@VividLemon Killing me Lemon, anyways I pushed a fix. I think when refactoring you swapped naming conventions and used the wrong path variables.

VividLemon:
> @VividLemon Killing me Lemon, anyways I pushed a fix. I think when refactoring you swapped naming conventions and used the wrong path variables.

At least it's written in TS now. 

Screenshot from 2022-11-13 15-49-52

One day I will see this number to be 0 :pray:

VividLemon:
I still think it's better to just recreate the svg from bootstrap, that way we can avoid this circular dependency structure. But if you're adamant to get this working it's up to you. You should be able to just pull the svg direct from getbootstrap.com

aceofwings:
> I still think it's better to just recreate the svg from bootstrap, that way we can avoid this circular dependency structure. But if you're adamant to get this working it's up to you. You should be able to just pull the svg direct from getbootstrap.com

Like use their CDN or just copy and paste SVG content into the BoostrapVue Library?

aceofwings:
> I still think it's better to just recreate the svg from bootstrap, that way we can avoid this circular dependency structure. But if you're adamant to get this working it's up to you. You should be able to just pull the svg direct from getbootstrap.com

and would you know off hand why local packages are not being resolved? Maybe something to do with the cache?

VividLemon:
> > I still think it's better to just recreate the svg from bootstrap, that way we can avoid this circular dependency structure. But if you're adamant to get this working it's up to you. You should be able to just pull the svg direct from getbootstrap.com
> 
> Like use their CDN or just copy and paste SVG content into the BoostrapVue Library?

Copy and paste the svg content. I don't believe bootstrap-vue does much styling onto the buttons by default, or has any props specific to modifying the button state, so it should be fairly simple. You could possibly add in some props of your own, but that's up to you.

You'd just have to make sure that the user can still modify them with slots.

VividLemon:
> > I still think it's better to just recreate the svg from bootstrap, that way we can avoid this circular dependency structure. But if you're adamant to get this working it's up to you. You should be able to just pull the svg direct from getbootstrap.com
> 
> and would you know off hand why local packages are not being resolved? Maybe something to do with the cache?

I think this is probably another core issue on why importing from the workspace is probably not the best idea. The item is trying to import from bootstrap-vue-3, but it's trying to grab the item from the /dist, in the test-unit and test-coverage ci, the package is never built, thus no /dist.

aceofwings:
@VividLemon good call. I do want to point out that package structures like this do exist.  Some will use @XXX/Shared to represent common code used across the project. npm will not let you create circular dependencies and I think you mean from bootstrap-vue-3-icons instead of boostrap-vue-3? I am just point that out because the icons package does not depend on bootstrap-vue.

aceofwings:
Im going to merge this @VividLemon

VividLemon:
K

Issue #810, opened 2022-11-20
Alert countdown does not start


xmatthias:
**Describe the bug**
Using a dismissable alert with countdown does only work if the v-model ref is changed after the alert is initialized.
Having the variable already set at some value will have the timer never start

**To Reproduce**
https://stackblitz.com/edit/vue-wdw4g9?file=src%2Fcomponents%2FHelloWorld.vue

The stackblitz shows the sample from the documentation - except that `dismissCountDown` is now initialized to `dismissSecs`

`const dismissCountDown = ref(dismissSecs);`

**Expected behavior**
Timer should start the moment the alert is shown - hiding the alert afterwards

**Environment (please complete the following information):**
 - OS: [e.g. Windows 11] 
 - Version [e.g. 0.2.6] 0.4.8

**Additional context**

This is obviously a usecase that could be covered by an `onMounted` hook - however my aim is to show a list of alerts (which are created as part of a v-for loop) - so i don't really have the possibility to change the value of the ref after initialization.

aceofwings:
Hi , a simple fix is on the way.

PR #812, opened 2022-11-21
fix: on balert create, start counter

aceofwings:
Port over tests to verify count down on render.

Refs: #810 

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

### DIFF
--- a/packages/bootstrap-vue-3/src/BootstrapVue.ts
+++ b/packages/bootstrap-vue-3/src/BootstrapVue.ts
@@ -9,7 +9,7 @@ import './styles/styles.scss'
 // Keep this list in sync with /components/index.ts please
 import BAccordion from './components/BAccordion/BAccordion.vue'
 import BAccordionItem from './components/BAccordion/BAccordionItem.vue'
-import BAlert from './components/BAlert.vue'
+import BAlert from './components/BAlert/BAlert.vue'
 import BAvatar from './components/BAvatar/BAvatar.vue'
 import BAvatarGroup from './components/BAvatar/BAvatarGroup.vue'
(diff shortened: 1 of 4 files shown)


aceofwings:
@VividLemon will be merging this.

VividLemon:
I had started working on updating the BAlert a bit ago. https://github.com/cdmoro/bootstrap-vue-3/compare/main...VividLemon:bootstrap-vue-3:BAlert The next thing on my list was to get the counter functional, then the alert was more or less done. Your ability to get the system to work on BAlert is helpful for now. However, I think we could improve it. Vueuse has some composables that relate to time management, but reviewing all of them, they seem to need some updating for our specific purpose. 

For example, useIntervalFn would handle the bulk of the work, pausing, resuming, starting. Being an interval function, it could handle the emitting for the countdown for us easily as well. However, I'm not sure if it handles _restarting_ by default. I think the best option is to pull in that code, or some other that may work, and see if we can move this logic into a _useTimer_ composable. Some other pieces of code also use the same functionality, I believe Toast is one. So standardizing a composable for components to use would be great. @aceofwings If you don't mind, could you possibly try to make a rough draft of implementing this composable? It was something I was working on, but I'm going to be a bit busy for the next week or two, I could probably get to it next weekend if you don't want to. 

With some of the features of vueuse's composables, we could possibly add some additional features, like waiting to start the timer.

aceofwings:
> With some of the features of vueuse's composables, we could possibly add some additional features, like waiting to start the timer or having a prop preference to determine how often the timer updates

I agree. I am right now setting aside time for some general bug hunting and reintroducing tests to verify clean transfer from vue2 to vue3. So I think I can draft something over the next couple of days.

PR #816, opened 2022-11-23
feat(Popover): Allow html modifier in directive

bettysteger:
# Describe the PR

Allow html modifier in Popover directive

## Small replication

I am using `v-b-popover.html="'<p>text</p>'"` and it didn't render the HTML but showed the actual HTML tags. 

## PR checklist

- [x] Feature - `feat(...)`
- [ ] Other (please describe)


bettysteger:
@VividLemon unfortunately that fix was removed, and i now got this problem again:

image

Should i make another PR?

VividLemon:
Sadly no, we removed this modifier as well as many other html changing effects for security reasons. We want to make a single way, components, with explicit slots/v-html modifiers to do this. That way it is always the same as to when one is displaying html content

Basically, it was removed to force you to not shoot yourself in the foot

Simply update from using the directive into using the component variant and use v-html in the slot

PR #888, opened 2023-01-23
fix(BTabs): add role="presentation"

danielelkington:
# Describe the PR

Add `role="presentation"` to the `li` within the B-Tabs component for improved accessibility.

Without this, Microsoft's [Accessibility Insights for Web](https://accessibilityinsights.io/docs/web/overview/) Chrome extension flags an error, as the default role for `li` (listitem) should only be in `ul` or `ol` elements that have no explicit role, or `role="list"`.

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

### DIFF
--- a/packages/bootstrap-vue-next/src/components/BTabs/BTabs.vue
+++ b/packages/bootstrap-vue-next/src/components/BTabs/BTabs.vue
@@ -30,6 +30,7 @@
           :key="idx"
           class="nav-item"
           :class="tab.props['title-item-class']"
+          role="presentation"
         >
           <button
             :id="buttonId"


VividLemon:
Please provide some sort of documentation on where it would show that presentation is recommended. Or  place in bootstrap vue code if possible

danielelkington:
In the boostrap-vue-next rendered output, using a `b-tabs` component results in output similar to
```html
<ul role="tablist">
  <li>
    <button role="tab"></button>
  </li>
</ul>
```
The mdn docs state that ["Every tablist must have one or more tab children"](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tablist_role#associated_wai-aria_roles_states_and_properties). However, the `<li>` element [has an implicit role of `listitem`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/li#technical_summary), which means the children of tablist in the accessibility tree incorrectly have the role `listitem`.

To fix this, we can override the default role of `<li>` and give it a role of `presentation`, which [removes its "implicit ARIA semantics from being exposed to the accessibility tree"](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/presentation_role).

This is done in the current Bootstrap Vue code https://github.com/bootstrap-vue/bootstrap-vue/blob/dev/src/components/tabs/tabs.js#L181

VividLemon (approved):

PR #930, opened 2023-02-13
Update README.md

averichev:


VividLemon:
These aren't the same and won't work

VividLemon:
(Bootstrapvue3 resolver will attempt to import from bootstrap-vue-3. Which is not the package name)

PR #938, opened 2023-02-20
remove popperjs and convert to floating-ui

xvaara:
Many breaking changes!

Need to see if the logic with the component is valid in your opinion.
```html
 <b-popover>
  <template #title>
    Title
  </template>
  <template #content>
    Content
  </template>
  <b-button>Trigger</b-button>
</b-popover>
```

the first element in default slot is the trigger.
Added some features, will document later.


packages/bootstrap-vue-next/src/components/BPopover.vue
```diff
@@ -1,198 +1,368 @@
...
+  () => [props.target, props.reference],
+  () => {
+    unbind()
+    bind()
+  }
+)
+
+onMounted(() => {
+  bind()
+})
```
VividLemon, line 363:
```suggestion
onMounted(bind)
```

packages/bootstrap-vue-next/src/components/BPopover.vue
```diff
@@ -1,198 +1,368 @@
...
+
+onBeforeUnmount(() => {
+  unbind()
 })
```
VividLemon, line 367:
```suggestion
onBeforeUnmount(unbind)
```

packages/bootstrap-vue-next/src/components/popover.spec.ts
```diff
@@ -7,22 +7,29 @@ describe('popover', () => {
 
   it('has static class popover', () => {
     const wrapper = mount(BPopover)
-    expect(wrapper.classes()).toContain('popover')
+    const [$div] = wrapper.findAll('div.popover')
```
VividLemon, line 10: There's no point to using findAll if you're simply going to use the first item. You want to check the existence of the component first, then do your check
```ts
it('has div.poper', () => {
const $div = wrapper.findd('div.popover')
expect($div.exist()).toBe(true)
})
// Then 
it('has static class popover', () => {
const $div = wrapper.get('div.popover')
// rest of code
})
```
It's more writing, but more explicit and makes failing tests, logically make more sense

packages/bootstrap-vue-next/src/utils/floatingUi.ts
```diff
@@ -0,0 +1,206 @@
...
+  if (pos.middlewareData.hide) {
+    const {referenceHidden} = pos.middlewareData.hide
+    !referenceHidden && floating.classList.add('show')
+    referenceHidden && floating.classList.remove('show')
+  } else {
+    floating.classList.add('show')
+  }
+}
+
+export function resolveBootstrapPlacement(placement: Placement): string {
```
VividLemon, line 128: Prefer arrow functions

packages/bootstrap-vue-next/src/utils/floatingUi.ts
```diff
@@ -0,0 +1,206 @@
...
+
+// TODO this function doesn't currently resolve with RTL in mind. Once Bootstrap finalizes their RTL, we should make this change here
+/**
+ * Configures Bootstrap-like placement props to floating-ui Placement strings.
+ * Top drops up, bottom drops down, end drops right, start drops left, dropend will _align_ the drop to the 'end',
+ * dropstart will _align_ the drop to the 'start'. Bottom is default, so it is the last in the order. Bottom should essentially be the opposite of top
+ * @param {top: boolean; bottom: boolean; start: boolean; end: boolean; dropstart: boolean; dropend: boolean}
+ * @returns {Placement} Placement
+ */
+export function resolveFloatingPlacement({
```
VividLemon, line 25: Prefer arrow functions

packages/bootstrap-vue-next/src/utils/sanitizer.ts
```diff
@@ -0,0 +1,127 @@
...
+  'small': [],
+  'span': [],
+  'sub': [],
+  'sup': [],
+  'strong': [],
+  'u': [],
+  'ul': [],
+}
+
+export function sanitizeHtml(
```
VividLemon, line 90: Prefer arrow functions

packages/bootstrap-vue-next/src/directives/BTooltip.ts
```diff
@@ -1,109 +1,127 @@
...
-    trigger.push('click')
+  if (trigger.includes('focus')) {
+    el.addEventListener('focus', addTooltipElement)
+    el.addEventListener('blur', removeTooltipElement)
   }
 
-  if (modifiers.hover) {
-    trigger.push('hover')
+  if (trigger.includes('hover')) {
+    el.addEventListener('pointerenter', addTooltipElement)
```
VividLemon, line 88: I'm not exactly sure how well composables will work in this context. But you could use the `useEventListener` from Vueuse. As well as possibly using the useFocus. There's also one for hovering, which I used somewhere else, but I don't remember the name of it. -- I'm not sure how well this would work though.

packages/bootstrap-vue-next/src/directives/BTooltip.ts
```diff
@@ -1,109 +1,127 @@
...
+  resolveTitle,
+  resolveTrigger,
+  updateFloatingPosition,
+} from '../utils/floatingUi'
+import {useId} from '../composables'
+
+const state: Record<string, {el: HTMLElement; binding: DirectiveBinding; cleanup: () => void}> = {}
+
+const createTooltipElement = (id: string, content: string, customClass: string): HTMLDivElement => {
+  const tooltip = document.createElement('div')
```
VividLemon, line 14: Afaik, you may be able to use h() instead of all this manual dom manipulation. (I think you could, but I'm not 100%)

packages/bootstrap-vue-next/src/components/popover.spec.ts
```diff
@@ -60,42 +69,43 @@ describe('popover', () => {
...
+    const [$div] = wrapper.findAll('div.trigger')
     expect($div.text()).toBe('slotbar')
   })
 
   it('contains b-popover-{type} if prop variant', async () => {
     const wrapper = mount(BPopover, {
       props: {variant: 'primary'},
     })
-    expect(wrapper.classes()).toContain('b-popover-primary')
+    const [$div] = wrapper.findAll('div.popover')
```
VividLemon, line 104: The same issue mentioned before with the findAll usage should be applied universally.

packages/bootstrap-vue-next/src/components/BPopover.vue
```diff
@@ -1,198 +1,368 @@
...
-    class="popover b-popover"
-    :class="computedClasses"
-    role="tooltip"
-    tabindex="-1"
-  >
-    <div ref="titleRef">
-      <slot name="title">
-        {{ title }}
-      </slot>
+  <span ref="placeholder" />
```
VividLemon, line 2: You may want to consider where to set $attrs, since the multiple root nodes will not know where to place them.

packages/bootstrap-vue-next/src/components/BPopover.vue
```diff
@@ -1,198 +1,368 @@
...
+)
+const noShiftBoolean = useBooleanish(toRef(props, 'noShift'))
+const noFlipBoolean = useBooleanish(toRef(props, 'noFlip'))
+const noFadeBoolean = useBooleanish(toRef(props, 'noFade'))
+const noAutoCloseBoolean = useBooleanish(toRef(props, 'noAutoClose'))
+const hideBoolean = useBooleanish(toRef(props, 'hide'))
+const realtimeBoolean = useBooleanish(toRef(props, 'realtime'))
+const hidden = ref(false)
+
+const elementRef = ref<HTMLElement>()
```
VividLemon, line 150: No other component appends "Ref" to a template ref. It may be a good idea to bring this in line with other components.

VividLemon, line 150:
```suggestion
const elementRef = ref<HTMLElement | null>(null)
```
This should apply to the rest.
Undefined and null work the same in these situations, but null is more explicit

packages/bootstrap-vue-next/src/components/BPopover.vue
```diff
@@ -1,198 +1,368 @@
...
+  target:
+    | (() => HTMLElement | VNode)
+    | string
+    | ComponentPublicInstance<HTMLElement>
+    | HTMLElement
+    | undefined
+): HTMLElement | string | undefined => {
+  if (typeof target === 'string') {
+    return target
+  } else if (target instanceof HTMLElement) {
```
VividLemon, line 166: Why nest in an `else if` if you're going to return? I prefer the style of 
```ts
if(this){
 return 'a'
}
if(otherThis){
 return 'b'
}
// etc
```
It's style, but I subjectively feel that the styles should be similar and uniform throughout components. That way it's easier for future contributors to understand and be comfortable with a uniform style. Such non-uniformity is one of the issues I have with reviewing Bootstrap-vue code.

packages/bootstrap-vue-next/src/components/BPopover.vue
```diff
@@ -1,198 +1,368 @@
...
   nextTick,
   onBeforeUnmount,
   onMounted,
-  type PropType,
   ref,
   toRef,
+  type VNode,
   watch,
 } from 'vue'
-import {Popover} from 'bootstrap'
```
VividLemon, line 36: :partying_face:

packages/bootstrap-vue-next/src/components/BPopover.vue
```diff
@@ -1,198 +1,368 @@
...
+    | (() => HTMLElement | VNode)
+    | string
+    | ComponentPublicInstance<HTMLElement>
+    | HTMLElement
+    | null
+  content?: string
+  id?: string
+  title?: string
+  delay?: number | {show: number; hide: number}
+  triggers?: string
```
VividLemon, line 84: This should probably be strongly typed. That way it's a bit more straightforward to the user

packages/bootstrap-vue-next/src/components/BPopover.vue
```diff
@@ -1,198 +1,368 @@
...
+  noShift: false,
+  noFade: false,
+  hide: true,
+  realtime: false,
+  reference: null,
+  target: null,
+})
+
+interface BPopoverEmits {
+  (e: 'show'): void
```
VividLemon, line 119: Review: https://bootstrap-vue.org/docs/components/popover#comp-ref-b-popover-events

Please match up the events in BV to be ported over to BVN. Hide/show are cancellable in this context. There's an example in BModal on how to set up cancellable events. (I have yet to figure out how to make the "show" cancel correctly -- it's transition to cancel starts before preventDefault is possible to be called. But that's a later issue)

It's something that will need to be done eventually regardless.

packages/bootstrap-vue-next/src/components/BPopover.vue
```diff
@@ -1,198 +1,368 @@
...
 
-    const getElement = (element: HTMLElement | string | undefined): HTMLElement | undefined => {
-      if (!element) return undefined
-      if (typeof element === 'string') {
-        const idElement = document.getElementById(element)
-        return idElement ? idElement : undefined
-      }
-      return element
-    }
+const arrowStyle = ref({position: 'absolute'} as CSSProperties)
```
VividLemon, line 219: Why not `const arrowStyle = ref<CSSProperties>({...})`?

packages/bootstrap-vue-next/src/components/BPopover.vue
```diff
@@ -1,198 +1,368 @@
...
-        title: props.title || slots.title ? titleRef.value : '',
-        content: contentRef.value,
-        html: htmlBoolean.value,
-        delay: props.delay,
-        sanitize: sanitizeBoolean.value,
-        offset: props.offset,
+const {isOutside} = useMouseInElement(elementRef)
+const {isOutside: triggerIsOutside} = useMouseInElement(triggerRef)
+
+function toggle() {
```
VividLemon, line 252: Preffer arrow functions

packages/bootstrap-vue-next/src/components/BPopover.vue
```diff
@@ -1,198 +1,368 @@
...
+  showState.value ? dohide() : show()
+}
+
+function show() {
```
VividLemon, line 256: Prefer arrow functions

packages/bootstrap-vue-next/src/components/BPopover.vue
```diff
@@ -1,198 +1,368 @@
...
-        if (show) {
-          instance.value?.show()
-        } else {
-          instance.value?.hide()
-        }
-      }
-    })
-
-    onMounted(() => {
+function dohide() {
```
VividLemon, line 270: Prefer arrow functions

packages/bootstrap-vue-next/src/directives/BTooltip.ts
```diff
@@ -1,109 +1,127 @@
-import type {Directive, DirectiveBinding} from 'vue'
-import {Tooltip} from 'bootstrap'
+import {type Directive, type DirectiveBinding, toRef} from 'vue'
```
VividLemon, line 1: It would be nice if Tooltip was also a composable. Afaik, it is possible to do something with the h() fn in composables.

What I would imagine this to be like, is, extract some of the Tooltip stuff written here out into an actual .vue file. Then, (assuming you can use h() in directives), use h() here, render the .vue file. Then, in the composable you can also do the same. 

The composable `useTooltip` would essentially take in a target "template ref", (or selector if you want to be cool like that). Then the second param would be an options object. Which would have the same purpose as some of the other details here, things like delay, variant, etc.

Most of the code you have written here is still usable, just needs to be converted to a .vue file.

VividLemon: Bootstrap-vue does something similar, to where the BTooltip is an actual .vue file. How they get it to render is a bit over me. But the point is, we should make the .vue file, and figure out how to use it in a directive after... Or if you're feeling adventurous, you could make a fake mock up first, demonstrating the usage.

VividLemon: The same general idea in this thread should be extended to BPopover as well. They should both be true components, both with a Directive & Composable

VividLemon (changes requested):
Some changes, some thoughts.

packages/bootstrap-vue-next/src/directives/BTooltip.ts
xvaara: We can't use composables in directives as I understand it. Tried it, didn't work, googled it found out they don't work. Would have liked to use them for sure.

packages/bootstrap-vue-next/src/directives/BTooltip.ts
xvaara: didn't think about that, haven't seen that in any non render function context. Will check it out.

packages/bootstrap-vue-next/src/directives/BTooltip.ts
xvaara: It seemed more complicated that doing these few dom manipulations. You need to build a new vue instansce and mount it to dom: https://github.com/Akryum/floating-vue/blob/main/packages/floating-vue/src/directives/v-tooltip.ts#L57

want me to give it a go?

packages/bootstrap-vue-next/src/directives/BTooltip.ts
xvaara: to use composables we'd need to rebuild these as components.

packages/bootstrap-vue-next/src/directives/BTooltip.ts
VividLemon: Yes. This is preferable since if we use real .vue components, we can spread them to directives and composables. Its a shame the approach us a bit convoluted, but you could likely make a util called "mount component" or something to abstract the business logic out

packages/bootstrap-vue-next/src/components/BPopover.vue
xvaara: It was a else if to start with... will change them.

packages/bootstrap-vue-next/src/components/BPopover.vue
xvaara: There was some weirdness with this.  I think I just tested many things and left it like that 😄, will change.

packages/bootstrap-vue-next/src/components/BPopover.vue
VividLemon: I may have not seen the diff. Its whatever, it happens

packages/bootstrap-vue-next/src/components/popover.spec.ts
xvaara: I just did something that works and did a search&replace for them. 😅

packages/bootstrap-vue-next/src/directives/BTooltip.ts
xvaara: I was also thinking... Do we need to support html in tooltip directives? bootstrap does. We could get rid of sanitizer if not. I'd like not to open us for XSS vulnerability. 😁

VividLemon: I feel like we should.

packages/bootstrap-vue-next/src/components/BPopover.vue
VividLemon: sounds like a good idea. Makes it kore standard to other components

packages/bootstrap-vue-next/src/components/BDropdown/BDropdown.vue
```diff
@@ -83,7 +83,7 @@ interface BDropdownProps {
   dropup?: Booleanish
   dropend?: Booleanish
   dropstart?: Booleanish
-  alignStart?: Booleanish
+  alignCenter?: Booleanish
```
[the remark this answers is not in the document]
xvaara: You want me to remove them from this pr?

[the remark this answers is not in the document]
VividLemon: Yes

packages/bootstrap-vue-next/src/components/BPopover.vue
xvaara: I did this already. I added "manual" so that people can build their own interaction in the default slot.

packages/bootstrap-vue-next/src/directives/BTooltip.ts
xvaara: Checkout the component versions 😁

tooltip is just popover with different class, so I just created that option to popover and created a proxy component for tooltip.

packages/bootstrap-vue-next/src/utils/floatingUi.ts
```diff
@@ -0,0 +1,125 @@
...
+  $__app?: App
+  $__element?: HTMLElement
+}
+
+export const bind = (el: ElementWithPopper, binding: DirectiveBinding) => {
+  const div = document.createElement('span')
+  if (binding.modifiers.body) document.body.appendChild(div)
+  else if (binding.modifiers.child) el.appendChild(div)
+  else el.parentNode?.insertBefore(div, el.nextSibling)
+  el.$__app = createApp({render: () => h(BPopover, {...el.$__state?.value})})
```
xvaara, line 114: I was really surprised that this worked 😂

packages/bootstrap-vue-next/src/components/BPopover.vue
```diff
@@ -1,198 +1,432 @@
...
+  (e: 'hide-prevented'): void
+  (e: 'show-prevented'): void
+  (e: 'update:modelValue', value: boolean): void
+}
+
+const emit = defineEmits<BPopoverEmits>()
+
+// const id = useId('b-popover', props.id)
+
+// const showState = useVModel(props, 'modelValue', emit, {passive: true})
```
VividLemon, line 156: Why not use this?

packages/bootstrap-vue-next/src/components/BPopover.vue
```diff
@@ -1,198 +1,432 @@
...
+  (e: 'hide', value: BvTriggerableEvent): void
+  (e: 'hidden', value: BvTriggerableEvent): void
+  (e: 'hide-prevented'): void
+  (e: 'show-prevented'): void
+  (e: 'update:modelValue', value: boolean): void
+}
+
+const emit = defineEmits<BPopoverEmits>()
+
+// const id = useId('b-popover', props.id)
```
VividLemon, line 154: Dead and duplicated on line 169

packages/bootstrap-vue-next/src/components/BPopover.vue
```diff
@@ -1,198 +1,432 @@
...
+const realtimeBoolean = useBooleanish(toRef(props, 'realtime'))
+const inlineBoolean = useBooleanish(toRef(props, 'inline'))
+const tooltipBoolean = useBooleanish(toRef(props, 'tooltip'))
+const hidden = ref(false)
+
+const element = ref<HTMLElement | null>()
+const target = ref<HTMLElement | null>()
+const arrow = ref<HTMLElement | null>()
+const trigger = ref<HTMLElement | null>()
+const placeholder = ref<HTMLElement | null>()
```
VividLemon, line 187:
```suggestion
const element = ref<HTMLElement | null>(null)
const target = ref<HTMLElement | null>(null)
const arrow = ref<HTMLElement | null>(null)
const trigger = ref<HTMLElement | null>(null)
const placeholder = ref<HTMLElement | null>(null)
```

packages/bootstrap-vue-next/src/components/BPopover.vue
```diff
@@ -1,198 +1,432 @@
...
+const getElement = (element: HTMLElement | string | undefined): HTMLElement | undefined => {
+  if (!element) return undefined
+  if (typeof element === 'string') {
+    const idElement = document.getElementById(element)
+    return idElement ? idElement : undefined
+  }
+  return element
+}
+
+const ishtml = computed(() => props.html === 'thisIsDangerous')
```
VividLemon, line 222: This doesn't look right.

packages/bootstrap-vue-next/src/components/BPopover.vue
```diff
@@ -1,198 +1,432 @@
...
+      top: y ? `${y}px` : '',
+      left: x ? `${x}px` : '',
     }
+  }
+})
+
+const computedClasses = computed(() => {
+  const type = tooltipBoolean.value ? 'tooltip' : 'popover'
+  return {
+    [type]: true,
```
VividLemon, line 282: When you have values like this that are always rendered, you can do something more like this: 
```ts
const computedClasses = computed(() => [
    type, // These can be in the array because they're always rendered, ie `[type]: true`
    `b-${type}`,
    {  // These go into an object because they are only soometimes rendered
        [`b-${type}-${props.variant}`]: props.variant !== undefined
       // etc
    }
]
```
BOffcanvas has a quick and simple example

VividLemon (changes requested):

packages/bootstrap-vue-next/src/components/popover.spec.ts
```diff
@@ -5,24 +5,37 @@ import BPopover from './BPopover.vue'
 describe('popover', () => {
   enableAutoUnmount(afterEach)
 
+  it('has div.poper', () => {
```
VividLemon, line 8:
```suggestion
  it('has div.popover', () => {
```

VividLemon:
Sorry about the merge conflicts. I refactored the usage of ts `Array<>` syntax to be `[]` syntax

packages/bootstrap-vue-next/src/components/BPopover.vue
xvaara: Yeah, I relly don't want users to create xss bugs with this. Maybe I just force sanitize all the input if :html="true" ?

packages/bootstrap-vue-next/src/components/BPopover.vue
xvaara: the useVModel doesn't work with delay, we'd need another ref that works with that. So I just created one with the ref + watch that does both.

PR #946, opened 2023-02-24
align dropdown with start as default, useVModel passive to have local…

xvaara:
I think it's better to start from start and to not require v-model to be set.

### DIFF
--- a/packages/bootstrap-vue-next/src/components/BDropdown/BDropdown.vue
+++ b/packages/bootstrap-vue-next/src/components/BDropdown/BDropdown.vue
@@ -63,6 +63,7 @@ import {useBooleanish, useId} from '../../composables'
 import type {Booleanish, ButtonType, ButtonVariant, ClassValue, Size} from '../../types'
 import {BvEvent, resolveFloatingPlacement, stringToInteger} from '../../utils'
 import BButton from '../BButton/BButton.vue'
+import type {RouteLocationRaw} from 'vue-router'
 
 // TODO add navigation through keyboard events
 // TODO standardize keydown vs keyup events globally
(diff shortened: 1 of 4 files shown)


VividLemon:
So vue use vmodel "passive" option essentially means that you don't actually need to bind a model value?

VividLemon:
Interesting behavior when using passive. I extended the same to BCarousel. We should figure out the best way to communicate in the docs that in some components, v-model is optional

packages/bootstrap-vue-next/src/components/BDropdown/BDropdown.vue
```diff
@@ -169,7 +171,7 @@ const floatingPlacement = computed(() =>
     bottom: !dropupBoolean.value,
     start: dropstartBoolean.value,
     end: dropendBoolean.value,
-    dropstart: alignStartBoolean.value,
+    dropstart: !alignCenterBoolean.value,
     dropend: alignEndBoolean.value,
```
VividLemon, line 175: What happens if you do _alignCenter_ and _alignEnd_?

xvaara:
> So vue use vmodel "passive" option essentially means that you don't actually need to bind a model value?

It basically creates a ref in the background and has a watch on it with a emit in it.

packages/bootstrap-vue-next/src/components/BDropdown/BDropdown.vue
xvaara: I looked in to this and opened a can of worms... The alignments can be breakpoint dependant. Do we want to support that? I'll take a look at the bootstrap docs with a thought.

xvaara: I redid this. Now we can use end and center on dropstart and dropend also. So we have all the floating-ui directions.

PR #969, opened 2023-03-12
Collapse native vue, no more bs.js!

xvaara:
No more bootstrap.js!

SSR works! The docs can be build without <ClientOnly>! I belive nuxt works also!

tried to keep the changes to minimal on places.

changed the v-b-toggle to work with events, and now toggle, offcanvas and modal listen to "toggle" event.

As a test I remove bootstrap in playground and only added it's css in link tag in the index.html file.

In bootstrap-vue-next we only need few scss functions/variables/mixins from bootstrap npm package.

### DIFF
--- a/packages/bootstrap-vue-next/src/components/BAccordion/BAccordionItem.vue
+++ b/packages/bootstrap-vue-next/src/components/BAccordion/BAccordionItem.vue
@@ -1,26 +1,27 @@
 <template>
   <div class="accordion-item">
-    <component :is="headerTag" :id="`${computedId}heading`" class="accordion-header">
-      <button
-        v-b-toggle:[computedId]
-        class="accordion-button"
-        :class="{collapsed: !visibleBoolean}"
-        type="button"
-        :aria-expanded="visibleBoolean ? 'true' : 'false'"
-        :aria-controls="computedId"
-      >
-        <slot name="title">
-          {{ title }}
-        </slot>
-      </button>
-    </component>
     <b-collapse
       :id="computedId"
+      v-model="open"
       class="accordion-collapse"
       :visible="visible"
       :accordion="parentData ?? undefined"
       :aria-labelledby="`heading${computedId}`"
     >
+      <template #header="{visible: toggleVisible, toggle}">
+        <component :is="headerTag" :id="`heading${computedId}`" class="accordion-header">
+          <button
+            class="accordion-button"
+            :class="{collapsed: !toggleVisible}"
+            type="button"
+            :aria-expanded="toggleVisible ? 'true' : 'false'"
+            :aria-controls="computedId"
+            @click="toggle"
+          >
+            <slot name="title"> {{ title }} </slot>
+          </button>
+        </component>
+      </template>
       <div class="accordion-body">
         <slot />
       </div>
@@ -29,9 +30,9 @@
 </template>
 
 <script setup lang="ts">
-import {inject, toRef} from 'vue'
+import {inject, onMounted, toRef, watchEffect} from 'vue'
+import {useVModel} from '@vueuse/core'
 import BCollapse from '../BCollapse.vue'
-import {BToggle as vBToggle} from '../../directives'
 import {accordionInjectionKey} from '../../utils'
 import {useBooleanish, useId} from '../../composables'
 import type {Booleanish} from '../../types'
@@ -40,18 +41,29 @@ import type {Booleanish} from '../../types'
 interface BAccordionItemProps {
   id?: string
   title?: string
+  modelValue?: boolean
   visible?: Booleanish
   headerTag?: string
 }
 
 const props = withDefaults(defineProps<BAccordionItemProps>(), {
-  visible: false,
   headerTag: 'h2',
 })
+
+const emit = defineEmits<(e: 'update:modelValue', value: boolean) => void>()
+const open = useVModel(props, 'modelValue', emit, {passive: true})
+
+const visibleBoolean = useBooleanish(toRef(props, 'visible'))
+
+onMounted(() => {
+  if (visibleBoolean.value) {
+    open.value = true
+  }
+})
+
+watchEffect(() => (open.value = visibleBoolean.value))
 
 const parentData = inject(accordionInjectionKey, null)
 
 const computedId = useId(toRef(props, 'id'), 'accordion_item')
-
-const visibleBoolean = useBooleanish(toRef(props, 'visible'))
 </script>
--- a/packages/bootstrap-vue-next/src/components/BCollapse.vue
+++ b/packages/bootstrap-vue-next/src/components/BCollapse.vue
@@ -1,24 +1,25 @@
 <template>
+  <slot name="header" v-bind="{visible: modelValueBoolean, toggle, open, close, id: computedId}" />
   <component
     :is="tag"
     :id="computedId"
     ref="element"
     class="collapse"
     :class="computedClasses"
-    :data-bs-parent="accordion || null"
     :is-nav="isNavBoolean"
+    v-bind="$attrs"
   >
-    <slot :visible="modelValueBoolean" :close="close" />
+    <slot v-bind="{visible: modelValueBoolean, toggle, open, close}" />
   </component>
+  <slot name="footer" v-bind="{visible: modelValueBoolean, toggle, open, close, id: computedId}" />
 </template>
 
 <script setup lang="ts">
-// import type {BCollapseEmits, BCollapseProps} from '../types/components'
-import {computed, onMounted, ref, toRef, watch} from 'vue'
-import {Collapse} from 'bootstrap'
+import {computed, nextTick, onMounted, ref, toRef, watch, watchEffect} from 'vue'
 import {useBooleanish, useId} from '../composables'
 import {useEventListener, useVModel} from '@vueuse/core'
 import type {Booleanish} from '../types'
+import {BvTriggerableEvent} from '../utils'
 
 interface BCollapseProps {
   accordion?: string
@@ -27,6 +28,7 @@ interface BCollapseProps {
   modelValue?: Booleanish
   tag?: string
   toggle?: Booleanish
+  horizontal?: Booleanish
   visible?: Booleanish
   isNav?: Booleanish
 }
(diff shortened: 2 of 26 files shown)


packages/bootstrap-vue-next/src/components/BModal.vue
```diff
@@ -387,6 +387,10 @@ onMounted(() => {
     isActive.value = true
   }
 })
+
+useEventListener(element, 'toggle', () => {
```
VividLemon, line 391: BModal has it's own vBModal directive. It would be nice if that worked

packages/bootstrap-vue-next/src/components/BCollapse.vue
```diff
@@ -35,69 +37,165 @@ const props = withDefaults(defineProps<BCollapseProps>(), {
...
-useEventListener(element, 'show.bs.collapse', () => {
-  emit('show')
+if (visibleBoolean.value) {
   modelValue.value = true
+  show.value = true
+}
+
+watch(visibleBoolean, (visible) => {
+  visible ? open() : close()
 })
```
VividLemon, line 183:
```suggestion
watchEffect(() => {
  visibleBoolean.value ? open() : close()
})
```

packages/bootstrap-vue-next/src/components/BAccordion/BAccordionItem.vue
```diff
@@ -40,18 +41,29 @@ import type {Booleanish} from '../../types'
...
+
+const visibleBoolean = useBooleanish(toRef(props, 'visible'))
+
+if (visibleBoolean.value) {
+  open.value = true
+}
+
+watch(visibleBoolean, (visible) => {
+  open.value = visible
+})
```
VividLemon, line 64:
```suggestion
watchEffect(() => (open.value = visibleBoolean.value))
```

packages/bootstrap-vue-next/src/components/BAccordion/BAccordionItem.vue
```diff
@@ -40,18 +41,29 @@ import type {Booleanish} from '../../types'
...
-  visible: false,
   headerTag: 'h2',
 })
 
+const emit = defineEmits<(e: 'update:modelValue', value: boolean) => void>()
+const open = useVModel(props, 'modelValue', emit, {passive: true})
+
+const visibleBoolean = useBooleanish(toRef(props, 'visible'))
+
+if (visibleBoolean.value) {
```
VividLemon, line 58: This is probably code that runs onMounted

VividLemon: But it's possible it would be done automatically from the watchEffect

VividLemon (changes requested):

packages/bootstrap-vue-next/src/components/BModal.vue
xvaara: Commited

apps/nuxt-playground/.npmrc
```diff
@@ -0,0 +1,2 @@
+shamefully-hoist=true
```
VividLemon, line 1: This will not work
https://github.com/pnpm/pnpm/issues/5627#issuecomment-1312851603

VividLemon: I would be fine with moving this npmrc to the root of the project. Hopefully it would be a temporary solution until Nuxt gets their act together.

apps/nuxt-playground/components
```diff
@@ -0,0 +1 @@
+../playground/src/components
```
VividLemon, line 1: I don't know what this is

apps/playground/index.html
```diff
@@ -5,6 +5,7 @@
     <link rel="icon" type="image/svg+xml" href="/favicon.ico" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <title>Vite + Vue + TS</title>
+    <link href="https://cdn.jsdelivr.net/npm/bootstrap@5.3.0-alpha1/dist/css/bootstrap.min.css" rel="stylesheet" integrity="sha384-GLhlTQ8iRABdZLl6O3oVMWSktQOp6b7In1Zl3/Jr59b6EGGoI1aFkw7cmDA6j6gD" crossorigin="anonymous">
```
VividLemon, line 8: I've seen you do this a few times... It may be better to declare Bootstrap as a dependency, so it can be easily identifiable in the project -- ie so it can be updated in the future easier. (Rather than needing to go out and find the dist url, you just have to modify the package.json). It would make it a bit easier, you just have to import the css from the package.

apps/nuxt-playground/.npmrc
xvaara: nuxt did that by default when I just ran `npx nuxi@latest init nuxt-playground`. Shall I move it?

apps/nuxt-playground/.gitignore
```diff
@@ -0,0 +1,9 @@
+node_modules
```
VividLemon, line 1: This is a general note about the entire nuxt-playground...

It may be better instead to start working on an entire docs rewrite that is using Nuxt. My plan is to use Nuxt & Nuxt Content to write the docs.

That way the playground can stay how it is (relatively unused), and the docs rewritten in Nuxt can serve as a useful testing ground for nuxt support.

VividLemon (changes requested):

apps/nuxt-playground/components
xvaara: it is a symlink

apps/nuxt-playground/.npmrc
VividLemon: I'm surprised anything is working at all. It should be behaving as if shamefully-hoist is false

apps/nuxt-playground/.gitignore
VividLemon: The docs rewrite is a useful feature since VuePress is fairly unmaintained compared to VitePress & Nuxt. I hesitate to say they are "deprecated", but VitePress is the official recommendation over VuePress, but either way it would be better for the docs to bring forth Nuxt specific issues, since most of the user base that wants SSR will likely be using Nuxt

packages/bootstrap-vue-next/src/components/BOffcanvas/BOffcanvas.vue
```diff
@@ -239,6 +239,9 @@ watch(
   },
   {flush: 'post'}
 )
+useEventListener(element, 'toggle', () => {
```
VividLemon, line 242: I apologize that I didn't bring this up in the earlier critique about vBModal addition, but it may be useful to prepend the event with a specific Bootstrap vue event listener, as to not be conflicting with users. Perhaps `bv-toggle` would be more appropriate. I am not sure how Bootstrap vue does this. If they do what you did here, it can stay fine, but if they have a specific way they have been doing it, it may be best to follow their convention

packages/bootstrap-vue-next/src/components/BToast/plugin.ts
```diff
@@ -189,10 +189,8 @@ export class ToastController {
 }
 
 // default global inject key to fetch the controller
-const injectkey = Symbol()
-const fetchKey = Symbol()
-
-const rootkey = 'root' // TODO: I guess this variable is not used in any place...
+const injectkey = Symbol('toast')
```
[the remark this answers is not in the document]
VividLemon: I am aware that this isn't your doing. I'm not sure if you did anything to Toasts in this PR, but Toast needs to get an overhaul as well. afaik it's been broken for awhile.

packages/bootstrap-vue-next/src/directives/BModal.ts
```diff
@@ -1,19 +1,2 @@
...
-    if (Object.keys(binding.modifiers).length > 0) {
-      ;[target] = Object.keys(binding.modifiers)
-    }
-
-    el.setAttribute('data-bs-toggle', 'modal')
-    el.setAttribute('data-bs-target', `#${target}`)
-  },
-} as Directive<HTMLElement>
+import BToggle from './BToggle'
+export default BToggle
```
VividLemon, line 2: lol

VividLemon (changes requested):

apps/nuxt-playground/components
VividLemon: In that case it may be better to review https://github.com/bootstrap-vue/bootstrap-vue-next/pull/969#discussion_r1136320173

packages/bootstrap-vue-next/src/components/BToast/plugin.ts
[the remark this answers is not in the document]
xvaara: They aren't used outside this file. And I don't think that is in the scope of this pull request. I forgot I added names to Symbols when trying to figure out why they didn't work. I can reset the file to the original.

packages/bootstrap-vue-next/src/directives/BModal.ts
xvaara: Well... 😆 I could have also exported BToggle two times on different names, but that seemed like a possibility for confusion.

packages/bootstrap-vue-next/src/components/BOffcanvas/BOffcanvas.vue
xvaara: bootstrap-vue uses vue2 $on and $emit on the root node to create a eventBus. It's a hacky way and it's not supported in vue3 without compat build. I just figured since we look for css-selector or ref, we could just sen an html event to it to make it simple. Toggle official event for anything... and I think now you could add <div id="foo" @toggle="doSomething"> and trigger that with the directive.

xvaara: Well I did put a foot in my mouth: https://developer.mozilla.org/en-US/docs/Web/API/HTMLDetailsElement/toggle_event

but that's the only element that has that event :D

apps/nuxt-playground/.gitignore
xvaara: Should we add some ssr for the App.vue in the main package then, so that ssr errors are caught early?
It's not that hard, little express server with vite middleware.
Or is the ssr test going to be building the doc to html?

xvaara: maybe add https://test-utils.vuejs.org/guide/advanced/ssr.html to all the components?

VividLemon: > maybe add https://test-utils.vuejs.org/guide/advanced/ssr.html to all the components?

Such a test would be too brittle. Any slight change would break this test.

> Or is the ssr test going to be building the doc to html?

This should be sufficient.

packages/bootstrap-vue-next/src/directives/BToggle.ts
```diff
@@ -52,57 +35,41 @@ const getTargets = (binding: DirectiveBinding<string>, el: HTMLElement) => {
...
+}
 
-    //     el.setAttribute('data-bs-target', selector)
-    // }
+/**
+ * @external
+ */
+export default {
+  mounted(el, binding: DirectiveBinding<string>): void {
+    el.addEventListener('click', () => toggle(binding, el))
```
VividLemon, line 71: Not sure if event listeners are automatically destroyed. You may want to clean this up manually, but maybe I'm wrong and nothing needs to happen

xvaara: Yes, forgot to do that. Good catch.

packages/bootstrap-vue-next/src/components/BAccordion/BAccordionItem.vue
```diff
@@ -40,18 +41,29 @@ import type {Booleanish} from '../../types'
...
   headerTag?: string
 }
 
 const props = withDefaults(defineProps<BAccordionItemProps>(), {
-  visible: false,
   headerTag: 'h2',
 })
 
+const emit = defineEmits<(e: 'update:modelValue', value: boolean) => void>()
+const open = useVModel(props, 'modelValue', emit, {passive: true})
```
VividLemon, line 54:
```suggestion
const open = useVModel(props, 'modelValue', emit, {passive: true})
const openBoolean = useBooleanish(open)
```

VividLemon: Expand this to work, of course

packages/bootstrap-vue-next/src/components/BCollapse.vue
```diff
@@ -35,69 +37,166 @@ const props = withDefaults(defineProps<BCollapseProps>(), {
...
-const visibleBoolean = useBooleanish(toRef(props, 'visible'))
+const horizontalBoolean = useBooleanish(toRef(props, 'horizontal'))
 const isNavBoolean = useBooleanish(toRef(props, 'isNav'))
+const visibleBoolean = useBooleanish(toRef(props, 'visible'))
 
 const computedId = useId(toRef(props, 'id'), 'collapse')
 
-const element = ref<HTMLElement | null>(null)
-const instance = ref<Collapse>()
+const element = ref<HTMLElement>(null as unknown as HTMLElement)
```
VividLemon, line 80:
```suggestion
const element = ref<HTMLElement | null>(null)
```

packages/bootstrap-vue-next/src/components/BCollapse.vue
```diff
@@ -35,69 +37,166 @@ const props = withDefaults(defineProps<BCollapseProps>(), {
...
+    setTimeout(() => {
+      show.value = false
+      isCollapsing.value = false
+      emit('hidden', buildTriggerableEvent('hidden'))
+    }, getTransitionDelay(element.value))
+  })
+}
+
+watch(modelValue, (newValue, oldValue) => {
+  if (newValue === oldValue) return
```
VividLemon, line 148: Under what circumstance would this occur

VividLemon: You should probably watch modelValueBoolean instead for more accurate values

packages/bootstrap-vue-next/src/components/BCollapse.vue
```diff
@@ -35,69 +37,166 @@ const props = withDefaults(defineProps<BCollapseProps>(), {
...
+    reveal()
+  } else {
+    hide()
+  }
 })
 
-watch(visibleBoolean, (value) => {
-  modelValue.value = !!value
-  value ? instance.value?.show() : instance.value?.hide()
+function getTransitionDelay(element: HTMLElement) {
```
VividLemon, line 158: use arrow function

packages/bootstrap-vue-next/src/directives/BToggle.ts
```diff
@@ -52,57 +35,51 @@ const getTargets = (binding: DirectiveBinding<string>, el: HTMLElement) => {
...
-    //     }
+interface WithToggle extends HTMLElement {
+  __toggle: () => void
+}
 
-    //     el.setAttribute('data-bs-target', selector)
-    // }
+/**
+ * @external
+ */
```
VividLemon, line 72:
```suggestion
```

VividLemon: Removing this globally. It's kind of dumb

VividLemon (changes requested):

packages/bootstrap-vue-next/src/components/BCollapse.vue
VividLemon: & fix whatever issues "nullability" causes

VividLemon:
After above changes, one more full PR review from me, then if I don't find anything else, it is done

PR #990, opened 2023-03-30
fix(BLink): prefer passed active prop over isActive slot prop from RouterLink

tvaliasek:
# Describe the PR

BLink component ignores directly set active prop, when router link passes isActive prop, this causes invalid assignment of active class when nested routes is used.

## Small replication

See tests, or:

Create vue instance with vue-router and nested routes and two BLinks, one heading to default route target and with active prop set to false, and then second to its nested route. After navigation to second route, both BLinks has active class assigned, but the first BLink shouldn't have.

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

### DIFF
--- a/packages/bootstrap-vue-next/src/components/BLink/BLink.vue
+++ b/packages/bootstrap-vue-next/src/components/BLink/BLink.vue
@@ -10,10 +10,7 @@
       :is="routerTag"
       ref="link"
       :href="href"
-      :class="[
-        (isActive || activeBoolean) && activeClass,
-        (isExactActive || exactBoolean) && exactActiveClass,
-      ]"
+      :class="[(activeBoolean ?? isActive) && activeClass, isExactActive && exactActiveClass]"
       v-bind="$attrs"
       @click="navigate"
     >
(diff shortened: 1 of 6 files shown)


VividLemon:
https://router.vuejs.org/guide/migration/#removal-of-the-exact-prop-in-router-link

Remove this prop. It should not be doing anything

packages/bootstrap-vue-next/src/components/BLink/BLink.vue
```diff
@@ -70,6 +70,24 @@ export default defineComponent({
     const instance = getCurrentInstance()
     const link = ref<HTMLElement>(null as unknown as HTMLElement)
 
+    const isActivePassed = computed(() => {
+      if (props.active === undefined) {
```
VividLemon, line 74: `useBooleanish` has overloads that allow for passing undefined directly. Where the output would be boolean | undefined...

With this in mind, you likely don't need to do all this manipulation, you could instead just do 
```html
(activeBoolean ?? isActive) && activeClass,
```
I believe this would have the same affect, in a more concise syntax.

tvaliasek: Fixed as suggested. The tests passed without errors, so it looks fine.

tvaliasek:
> https://router.vuejs.org/guide/migration/#removal-of-the-exact-prop-in-router-link
> 
> Remove this prop. It should not be doing anything

Removed exact prop, also removed from dependent BButton component.

packages/bootstrap-vue-next/src/components/BLink/BLink.vue
```diff
@@ -39,12 +36,11 @@ import {computed, defineComponent, getCurrentInstance, type PropType, ref, toRef
...
 
 export const BLINK_PROPS = {
-  active: {type: [Boolean, String] as PropType<Booleanish>, default: false},
+  active: {type: [Boolean, String, undefined] as PropType<Booleanish>, default: undefined},
   activeClass: {type: String, default: 'router-link-active'},
   append: {type: [Boolean, String] as PropType<Booleanish>, default: false},
   disabled: {type: [Boolean, String] as PropType<Booleanish>, default: false},
   event: {type: [String, Array], default: 'click'},
-  exact: {type: [Boolean, String] as PropType<Booleanish>, default: false},
   exactActiveClass: {type: String, default: 'router-link-exact-active'},
```
VividLemon, line 44: Subsequently exactActiveClass should be removed. It serves no real purpose that couldn't be resolved from using $attrs

VividLemon (changes requested):

packages/bootstrap-vue-next/src/components/BLink/BLink.vue
tvaliasek: Ok, removed from both BLink and BButton.

VividLemon (approved):

PR #1000, opened 2023-04-04
AccordionItem closes when another is opened, add free boolean for opening multiple

### LINKED ISSUE #994: Collapse -> Accordion not working
Collapse works, but not accordion.
"Note that only one collapse in an accordion group can be open at a time." - that is not working 
it works in "bootstrap vue"

OS: [e.g. Windows 10]
Package Version: [0.7.3]
Npm Version: [npm@9.2.0]

xvaara:
As the title says.
fixes #994 
Adds v-modal to BAccordion that is the id of the open accordion. 
Add free booleanish to open multiple
Integrate more with BCollapse

### DIFF
--- a/packages/bootstrap-vue-next/src/components/BAccordion/BAccordion.vue
+++ b/packages/bootstrap-vue-next/src/components/BAccordion/BAccordion.vue
@@ -6,14 +6,16 @@
 
 <script setup lang="ts">
 import type {Booleanish} from '../../types'
-import {computed, provide, ref, toRef} from 'vue'
+import {computed, provide, toRef} from 'vue'
 import {accordionInjectionKey} from '../../utils'
 import {useBooleanish, useId} from '../../composables'
+import {useVModel} from '@vueuse/core'
 
 interface BAccordionProps {
   flush?: Booleanish
   free?: Booleanish
   id?: string
+  modelValue?: string
 }
 
 const props = withDefaults(defineProps<BAccordionProps>(), {
@@ -22,7 +24,8 @@ const props = withDefaults(defineProps<BAccordionProps>(), {
   id: undefined,
 })
 
-const openItem = ref('')
+const emit = defineEmits<(e: 'update:modelValue', value: string) => void>()
+const modelValue = useVModel(props, 'modelValue', emit, {passive: true})
 
 const computedId = useId(toRef(props, 'id'), 'accordion')
 
@@ -35,8 +38,7 @@ const computedClasses = computed(() => ({
 
 if (!freeBoolean.value) {
   provide(accordionInjectionKey, {
-    id: computedId,
-    openItem,
+    openItem: modelValue,
     free: freeBoolean,
   })
 }
--- a/packages/bootstrap-vue-next/src/components/BAccordion/BAccordion.vue
+++ b/packages/bootstrap-vue-next/src/components/BAccordion/BAccordion.vue
@@ -25,6 +25,7 @@ const props = withDefaults(defineProps<BAccordionProps>(), {
 })
 
 const emit = defineEmits<(e: 'update:modelValue', value: string) => void>()
+
 const modelValue = useVModel(props, 'modelValue', emit, {passive: true})
 
 const computedId = useId(toRef(props, 'id'), 'accordion')
--- a/packages/bootstrap-vue-next/src/components/BAccordion/BAccordion.vue
+++ b/packages/bootstrap-vue-next/src/components/BAccordion/BAccordion.vue
@@ -6,7 +6,7 @@
 
 <script setup lang="ts">
 import type {Booleanish} from '../../types'
-import {computed, provide, toRef} from 'vue'
+import {computed, provide, ref, toRef} from 'vue'
 import {accordionInjectionKey} from '../../utils'
 import {useBooleanish, useId} from '../../composables'
 
@@ -22,6 +22,8 @@ const props = withDefaults(defineProps<BAccordionProps>(), {
   id: undefined,
 })
 
+const openItem = ref('')
+
 const computedId = useId(toRef(props, 'id'), 'accordion')
 
 const flushBoolean = useBooleanish(toRef(props, 'flush'))
@@ -34,6 +36,8 @@ const computedClasses = computed(() => ({
 if (!freeBoolean.value) {
   provide(accordionInjectionKey, {
     id: computedId,
+    openItem,
+    free: freeBoolean,
   })
 }
 </script>
--- a/packages/bootstrap-vue-next/src/components/BAccordion/BAccordion.vue
+++ b/packages/bootstrap-vue-next/src/components/BAccordion/BAccordion.vue
@@ -6,7 +6,7 @@
 
 <script setup lang="ts">
 import type {Booleanish} from '../../types'
-import {computed, provide, toRef} from 'vue'
+import {computed, provide, readonly, toRef} from 'vue'
 import {accordionInjectionKey} from '../../utils'
 import {useBooleanish, useId} from '../../composables'
 import {useVModel} from '@vueuse/core'
@@ -38,7 +38,7 @@ const computedClasses = computed(() => ({
 }))
 
 provide(accordionInjectionKey, {
-  openItem: modelValue,
+  openItem: readonly(modelValue),
   free: freeBoolean,
   setOpenItem: (id: string) => {
     modelValue.value = id
--- a/packages/bootstrap-vue-next/src/components/BAccordion/BAccordion.vue
+++ b/packages/bootstrap-vue-next/src/components/BAccordion/BAccordion.vue
@@ -36,10 +36,8 @@ const computedClasses = computed(() => ({
   'accordion-flush': flushBoolean.value,
 }))
 
-if (!freeBoolean.value) {
-  provide(accordionInjectionKey, {
-    openItem: modelValue,
-    free: freeBoolean,
-  })
-}
+provide(accordionInjectionKey, {
+  openItem: modelValue,
+  free: freeBoolean,
+})
 </script>
--- a/packages/bootstrap-vue-next/src/components/BAccordion/BAccordion.vue
+++ b/packages/bootstrap-vue-next/src/components/BAccordion/BAccordion.vue
@@ -40,5 +40,8 @@ const computedClasses = computed(() => ({
 provide(accordionInjectionKey, {
   openItem: modelValue,
   free: freeBoolean,
+  setOpenItem: (id: string) => {
+    modelValue.value = id
+  },
 })
 </script>
--- a/packages/bootstrap-vue-next/src/components/BAccordion/BAccordionItem.vue
+++ b/packages/bootstrap-vue-next/src/components/BAccordion/BAccordionItem.vue
@@ -4,7 +4,7 @@
       :id="computedId"
       v-model="modelValue"
       class="accordion-collapse"
-      :visible="visible"
+      v-bind="$attrs"
       :aria-labelledby="`heading${computedId}`"
     >
       <template #header="{visible: toggleVisible, toggle}">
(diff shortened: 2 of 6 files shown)


packages/bootstrap-vue-next/src/components/BCollapse.vue
```diff
@@ -15,14 +15,13 @@
...
-import {computed, nextTick, onMounted, ref, toRef, watchEffect} from 'vue'
+import {computed, nextTick, onMounted, ref, toRef, watch, watchEffect} from 'vue'
 import {useBooleanish, useId} from '../composables'
 import {useEventListener, useVModel} from '@vueuse/core'
 import type {Booleanish} from '../types'
 import {BvTriggerableEvent} from '../utils'
 
-interface BCollapseProps {
-  accordion?: string
+export interface BCollapseProps {
```
VividLemon, line 24: I don't think you're allowed to export from script setup

xvaara: I think because it's only typescript functionality it's allowed. It is exported before the vue compiler munches over the setup part. Not sure, but some random stackoverflow post said you could do this and it works... so 😅🤷 up to you.

VividLemon: Okay, so this is possible. I thought it would first error on you because there is an eslint rule to disable named exports in script setup.

However this is a type and will not suffer from the limitation presented by (exporting a variable from a setup function that is inside of a export default object), just doesn't make sense.

However, there is still an issue with the usage of this, I will describe in a different comment

xvaara: it's used in vue docs on script setup: https://vuejs.org/api/sfc-script-setup.html#default-props-values-when-using-type-declaration

packages/bootstrap-vue-next/src/components/BAccordion/BAccordionItem.vue
```diff
@@ -29,41 +28,57 @@
...
-import BCollapse from '../BCollapse.vue'
+import {default as BCollapse, type BCollapseEmits, type BCollapseProps} from '../BCollapse.vue'
+// import  from '../BCollapse.vue'
 import {accordionInjectionKey} from '../../utils'
-import {useBooleanish, useId} from '../../composables'
+import {useId} from '../../composables'
 import type {Booleanish} from '../../types'
 
-interface BAccordionItemProps {
+interface BAccordionItemProps extends BCollapseProps {
```
VividLemon, line 45: This is **impossible** syntax. Extending prop types will break Vue as it's currently not supported by the Vue compiler. Your best bet is to just duplicate the props and leave a note. Alternatively, you could revert the component to <script> and use real objects, as shown in BLink & components that depend on its props. However, this is not the best option, I'd prefer to just duplicate the props and use a note.

packages/bootstrap-vue-next/src/components/BCollapse.vue
VividLemon: > it's used in vue docs on script setup: https://vuejs.org/api/sfc-script-setup.html#default-props-values-when-using-type-declaration

The problem is not with exporting a type, the problem is with extending a prop type, as the Vue compiler cannot crawl types:
https://vuejs.org/guide/typescript/composition-api.html#syntax-limitations 
> This is because Vue components are compiled in isolation and the compiler currently does not crawl imported files in order to analyze the source type. This limitation could be removed in a future release.

packages/bootstrap-vue-next/src/components/BAccordion/BAccordionItem.vue
xvaara: it works, You can't export an extended interface, that's what vue compiler doesn't support.
here's the bug for that: https://github.com/vuejs/core/issues/7794

xvaara: ah now I get what you mean:
https://sfc.vuejs.org/#eNqVkd1ugzAMhV/Fyk1biZ97RpH28wB7gNwwaihVcSLbtJ0Q774AnVZ106bdRIljfzrneDCP3ienHk1mcqm49QqC2ns4ltRsrVGxprDUdt6xgr57hAFe2Xl5KgVhhJpdB6skfXbdDFqFZlLkuqxwaQS8KNJObsYGqJ3LQJRbagIELFWORMHPA1s4t7p/wbrsjyrrHdYt4Tydz2ex3kQwjJsHS3m6qC5MZBaRcVf65CCOgqXBUmBfP4KTDObKVAtSp7c1e1UvWZpKXU36D5I4btJwS7gnbTtMULr4jd1ZkAPYmuiGkYbiCTnm4BAZ+TfmXes37oQdLY3Bymeaf6wFL/Na7gK/Rjwt6yvj/yU8IX5MefwAg0rGqA==

xvaara: we don't need them (BCollapseProps) in vue side anyway. But it works on the editor, I get the code completion from BCollapseProps for BAccordionItem. Every prop that BAccordionItem uses is defined in that interface. Everything else is forwarded to BCollapse.

xvaara: So if vue implements a fix for the extends interface, It'll break this. So maybe better to do this the hard way. 😅

packages/bootstrap-vue-next/src/components/BAccordion/BAccordion.vue
```diff
@@ -9,18 +9,23 @@ import type {Booleanish} from '../../types'
...
   id?: string
+  modelValue?: string
 }
 
 const props = withDefaults(defineProps<BAccordionProps>(), {
   flush: false,
   free: false,
 })
 
+const emit = defineEmits<(e: 'update:modelValue', value: string) => void>()
```
VividLemon, line 26:
```suggestion
const emit = defineEmits<(e: 'update:modelValue', value: string) => void>()

```

packages/bootstrap-vue-next/src/components/BCollapse.vue
```diff
@@ -180,7 +178,7 @@ if (visibleBoolean.value) {
   show.value = true
 }
 
-watchEffect(() => {
```
VividLemon, line 183: Why did you remove this watchEffect

xvaara: it was launching open() twice, don't know why. As general I've had so much problems with watchEffect in this project that it's the first thing I remove and usually it works.

VividLemon: > it was launching open() or close() everytime mounted, don't know why. As general I've had so much problems with watchEffect in this project that it's the first thing I remove and usually it works.

You should try to take a look at https://vuejs.org/api/reactivity-core.html#watcheffect and their 
```ts
interface WatchEffectOptions {
  flush?: 'pre' | 'post' | 'sync' // default: 'pre'
  onTrack?: (event: DebuggerEvent) => void
  onTrigger?: (event: DebuggerEvent) => void
}
```
I'd try messing around with the flush value. It's kind of a longshot, but who knows, maybe it will magically fix the issue?

packages/bootstrap-vue-next/src/components/BCollapse.vue
```diff
@@ -180,7 +178,7 @@ if (visibleBoolean.value) {
...
+watch(visibleBoolean, () => {
```
VividLemon, line 181: You will likely want 
```ts
watch(visibleBoolean, (newValue) => {
  newValue ? ...
}
```

packages/bootstrap-vue-next/src/components/BCollapse.vue
xvaara: It doesn't. watchEffect is run always, not only on change. It also watches all the reactive variables in the functions.

packages/bootstrap-vue-next/src/components/BAccordion/BAccordion.vue
```diff
@@ -30,9 +36,8 @@ const computedClasses = computed(() => ({
   'accordion-flush': flushBoolean.value,
 }))
 
-if (!freeBoolean.value) {
-  provide(accordionInjectionKey, {
-    id: computedId,
-  })
-}
+provide(accordionInjectionKey, {
+  openItem: modelValue,
```
VividLemon, line 40: I'd suggest wrapping this in readonly

packages/bootstrap-vue-next/src/components/BAccordion/BAccordionItem.vue
```diff
@@ -29,41 +34,77 @@
   </div>
 </template>
 
+<script lang="ts">
```
VividLemon, line 37: Move this block to the bottom of the file

packages/bootstrap-vue-next/src/components/BAccordion/BAccordionItem.vue
```diff
@@ -29,41 +34,77 @@
...
 <script setup lang="ts">
-import {inject, onMounted, toRef, watchEffect} from 'vue'
+import {inject, onMounted, ref, toRef, watch} from 'vue'
 import {useVModel} from '@vueuse/core'
 import BCollapse from '../BCollapse.vue'
-import {accordionInjectionKey} from '../../utils'
-import {useBooleanish, useId} from '../../composables'
+// import  from '../BCollapse.vue'
+import {accordionInjectionKey, BvTriggerableEvent} from '../../utils'
+
```
VividLemon, line 48:
```suggestion
```

packages/bootstrap-vue-next/src/components/BAccordion/BAccordionItem.vue
```diff
@@ -29,41 +34,77 @@
...
+  (e: 'show', value: BvTriggerableEvent): void
+  (e: 'shown', value: BvTriggerableEvent): void
+  (e: 'hide', value: BvTriggerableEvent): void
+  (e: 'hidden', value: BvTriggerableEvent): void
+  (e: 'hide-prevented'): void
+  (e: 'show-prevented'): void
+  (e: 'update:modelValue', value: boolean): void
+}
+
+const emit = defineEmits<BAccordionItemEmits>()
```
VividLemon, line 78:
```suggestion
const emit = defineEmits<BAccordionItemEmits>()

```

packages/bootstrap-vue-next/src/components/BAccordion/BAccordionItem.vue
```diff
@@ -29,41 +34,77 @@
...
+  openItem: ref(''),
+  free: ref(false),
+})
+
+const computedId = useId(toRef(props, 'id'), 'accordion_item')
 
 onMounted(() => {
-  if (visibleBoolean.value) {
+  if (modelValue.value && !free.value) {
+    openItem.value = computedId.value
```
VividLemon, line 99: I am not certain, but I'm wondering if you can assign this in the provide default
```ts
const {openItem, free} = inject(accordionInjectionKey, {
  openItem: ref(computedId.value),
  free: ref(false),
})
```
There was that issue with some values not being in the correct state in between the finishing of mounting, and the finishing of the execution of the onMounted callback. I assume this would suffer from the same issue.

packages/bootstrap-vue-next/src/components/BCollapse.vue
```diff
@@ -52,6 +41,15 @@ interface BCollapseEmits {
   (e: 'update:modelValue', value: boolean): void
 }
 
+const props = withDefaults(defineProps<BCollapseProps>(), {
+  modelValue: false,
```
VividLemon, line 45: Ensure you have fetched the main branch. There may be eslint rules that may be missing. (vue/require-default-prop)

packages/bootstrap-vue-next/src/utils/keys.ts
```diff
@@ -45,7 +45,8 @@ export const avatarGroupInjectionKey: InjectionKey<{
 
 // BAccordion
 export const accordionInjectionKey: InjectionKey<{
-  id: Ref<string>
+  openItem: Ref<string>
```
VividLemon, line 48: Wrap these in Readonly<> & use the equivalent runtime vue readonly() if needed

VividLemon (changes requested):

packages/bootstrap-vue-next/src/components/BAccordion/BAccordionItem.vue
xvaara: The parent doesn't know what might be open, if accordionItem is opened with visible or toggle Booleanish. If multiple is open the "last one wins".

packages/bootstrap-vue-next/src/components/BAccordion/BAccordion.vue
VividLemon: You created a method to set this, but didn't wrap it in readonly

PR #1035, opened 2023-04-12
enable fade on tabs, redo tabs only when slot is changed.

### LINKED ISSUE #251: BTabs fade effect is not working
I'm using it in a private repo and it's not doing the fade effect despite everything seemingly looking good.

Note: it also does not work in the doc examples
https://cdmoro.github.io/bootstrap-vue-3/components/Tabs.html

Looking at https://getbootstrap.com/docs/5.0/components/navs-tabs/#javascript-behavior and inspecting elements, it seems everything is similar except for the aria. I don't know if that is the cause of it.

xvaara:
fixes #251 

split the tabs to two, watch for changes in slots.default and recreate the tabs only then. Otherwise we're creating the tabs object on each change and the ID:s change.
Then have computed for the local changes.

### DIFF
--- a/packages/bootstrap-vue-next/src/components/BTabs/BTab.vue
+++ b/packages/bootstrap-vue-next/src/components/BTabs/BTab.vue
@@ -65,10 +65,15 @@ const showSlot = computed<boolean>(() => {
     computedLazy.value && computedLazyOnce.value && lazyRenderCompleted.value
   return computedActive.value || !computedLazy.value || hasLazyRenderedOnce
 })
-
+const show = ref(activeBoolean.value)
+watch(activeBoolean, (active) => {
+  setTimeout(() => {
+    show.value = active
+  }, 0)
+})
 const computedClasses = computed(() => ({
   'active': activeBoolean.value,
-  'show': activeBoolean.value,
+  'show': show.value,
   'card-body': parentData?.card.value && props.noBody === false,
 }))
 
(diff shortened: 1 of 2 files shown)


packages/bootstrap-vue-next/src/components/BTabs/BTab.vue
```diff
@@ -65,10 +65,15 @@ const showSlot = computed<boolean>(() => {
     computedLazy.value && computedLazyOnce.value && lazyRenderCompleted.value
   return computedActive.value || !computedLazy.value || hasLazyRenderedOnce
 })
-
+const show = ref(activeBoolean.value)
+watch(activeBoolean, (active) => {
+  setTimeout(() => {
```
VividLemon, line 70: I saw this same 'setTimeout' hack in a previous PR.

What is the purpose of this?

If it's to force nextTick, perhaps you should use `{ flush: 'post' }` instead

VividLemon (changes requested):

packages/bootstrap-vue-next/src/components/BTabs/BTab.vue
xvaara: The show class needs to be in the dom before we add the active class for the animation to launch. It's just to delay the show. `nextTick` / `{ flush: 'post' }` for some resaon didn't work, the dom isn't updated in time. Probably they have some optimizations.

xvaara: Probably. It would require some bigger changes, and this was a quick fix. If there are any problems with this we can revisit this.

PR #1053, opened 2023-04-16
fix(BDropdownItem): add attrs when no to props is present

### LINKED ISSUE #1048: [0.8.4] data-attribute would be nice
in bootstrap vue, it was possible (on almost components) to set the data-attribute, for example:

`  <b-dropdown-item @click="onclick_searchtype"  data-searchtype="lastname">Lastname</b-dropdown-item>`

I use it a lot, and it's a handy way to assign data to an element.

**what do you thing, is it possible to add this feature as in bootstrap-vue ?**

xvaara:
fixes #1048


VividLemon:
I thought it was a bit suspicious that an attribute wouldn't just fall through :thinking:

PR #1054, opened 2023-04-16
fix(BPopover): remove hidden element from dom

### LINKED ISSUE #1044: [0.8.4]  b-popover works different than in 0.7.3
a) prop "placement" - value "auto" is missing - It's a shame because it automatically calculated the position, and that can change depending on where the popover is installed

b) i use this: :delay="{ show: 1000, hide: 10 }"
    on my gui there are more elements with a popover functionality (fullcalendar events)
    now, it happens, that more than one popup is open, it does not close, when i move away with the mouse

all that worked in Version 0.7.3

xvaara:
works with overflow: scroll + container="body"

fixes #1044 problem that @aki77 has.

also use the RenderOrSkip component for popover.

### DIFF
--- a/packages/bootstrap-vue-next/src/components/RenderComponentOrSkip.vue
+++ b/packages/bootstrap-vue-next/src/components/RenderComponentOrSkip.vue
@@ -1,5 +1,5 @@
 <script lang="ts">
-import {defineComponent, h, Teleport} from 'vue'
+import {defineComponent, h, type PropType, type RendererElement, Teleport} from 'vue'
 
 export default defineComponent({
   name: 'ComponentOrEmpty',
@@ -9,7 +9,7 @@ export default defineComponent({
       default: 'div',
     },
     to: {
-      type: [String, HTMLElement],
+      type: [String, Object] as PropType<string | RendererElement | null | undefined>,
       required: false,
       default: null,
     },
--- a/packages/bootstrap-vue-next/src/components/RenderComponentOrSkip.vue
+++ b/packages/bootstrap-vue-next/src/components/RenderComponentOrSkip.vue
@@ -1,5 +1,5 @@
 <script lang="ts">
-import {defineComponent, h} from 'vue'
+import {defineComponent, h, Teleport} from 'vue'
 
 export default defineComponent({
   name: 'ComponentOrEmpty',
@@ -8,13 +8,23 @@ export default defineComponent({
       type: String,
       default: 'div',
     },
+    to: {
+      type: [String, HTMLElement],
+      required: false,
+      default: null,
+    },
     skip: {
       type: Boolean,
       default: false,
     },
   },
   setup(props, {slots, attrs}) {
-    return () => (props.skip ? slots.default?.() : h(props.tag, {...attrs}, [slots.default?.()]))
+    return () =>
+      props.skip
+        ? slots.default?.()
+        : props.tag === 'Teleport'
+        ? h(Teleport, {to: props.to}, [slots.default?.()])
+        : h(props.tag, {...attrs}, [slots.default?.()])
   },
 })
 </script>
--- a/packages/bootstrap-vue-next/src/components/RenderComponentOrSkip.vue
+++ b/packages/bootstrap-vue-next/src/components/RenderComponentOrSkip.vue
@@ -10,7 +10,6 @@ export default defineComponent({
     },
     to: {
       type: [String, Object] as PropType<string | RendererElement | null | undefined>,
-      required: false,
       default: null,
     },
     skip: {
--- a/apps/playground/src/components/Comps/TPopover.vue
+++ b/apps/playground/src/components/Comps/TPopover.vue
@@ -185,6 +185,18 @@
           in body. {{ textValue }}
         </b-popover>
       </b-col>
+      <b-col>
+        <div style="height: 50vh; width: 400px; overflow-y: scroll; padding: 5em;">
+          <div v-for="(_, i) in Array(20)" :key="i" style="height: 100px">
+            <b-popover v-bind="vari">
+              jee
+              <template #target>
+                <b-button>hover / focus</b-button>
+              </template>
+            </b-popover>
+          </div>
+        </div>
+      </b-col>
     </b-row>
   </b-container>
 </template>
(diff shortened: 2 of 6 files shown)


xvaara:
@VividLemon is there some problem with this pr? My other stuff is waiting for this to be committed so there isn't so many conflicts.

VividLemon:
I thought you had noticed that it doesn't build

packages/bootstrap-vue-next/src/components/RenderComponentOrSkip.vue
```diff
@@ -8,13 +8,23 @@ export default defineComponent({
       type: String,
       default: 'div',
     },
+    to: {
+      type: [String, HTMLElement],
```
VividLemon, line 12: I think you likely are looking for RendererElement from Vue
image

xvaara: So what is that in javascript land? Object? What do I put inside type, since it's javascript not typescript part?

xvaara:
> I thought you had noticed that it doesn't build

I just thought that nuxt build was broken in main since it hasn't build on my machine in few other branches.

I get errors like: 
```
@bootstrap-vue-next/nuxt:build: [nitro]  ERROR  Error: ENOENT: no such file or directory, realpath '--path--/bootstrap-vue-3/node_modules/.pnpm/vue@3.2.47/node_modules/@vue/server-renderer'
@bootstrap-vue-next/nuxt:build: 
@bootstrap-vue-next/nuxt:build: 
@bootstrap-vue-next/nuxt:build: undefined
@bootstrap-vue-next/nuxt:build: 
@bootstrap-vue-next/nuxt:build: 
@bootstrap-vue-next/nuxt:build:  ERROR  ENOENT: no such file or directory, realpath '--path--/bootstrap-vue-3/node_modules/.pnpm/vue@3.2.47/node_modules/@vue/server-renderer'
@bootstrap-vue-next/nuxt:build: 
@bootstrap-vue-next/nuxt:build: 
@bootstrap-vue-next/nuxt:build: 
@bootstrap-vue-next/nuxt:build:  ELIFECYCLE  Command failed with exit code 1.
@bootstrap-vue-next/nuxt:build:  ELIFECYCLE  Command failed with exit code 1.
@bootstrap-vue-next/nuxt:build: ERROR: command finished with error: command (--path--/bootstrap-vue-3/packages/nuxt) pnpm run build exited (1)
```
I've tried to reinstall node_modules etc...

packages/bootstrap-vue-next/src/components/RenderComponentOrSkip.vue
VividLemon: HTMLElement is, as expected, a JS class. Which can be used and interpreted by the Vue compiler. Typically props are constructor functions, String, Number, Array, Object, etc. However, you can use a class as one of these just fine. Though, it is not typical to use pure classes in Vue due to the unpopularity of JS classes. So "HTMLElement" throws an issue in the Nuxt app, as according the Nuxt, HTMLElement doesn't exist. So it doesn't know what to do.

RendererElement is a TS type. The Vue compiler doesn't really know what to do with any of them, most of the time Vue will compile them out altogether and just leave, what really is just a blank placeholder (When using Composition api, in options api they have the String, Array, Number constructors). The only thing that matters in the blank placeholder is TS holding you back, not necessarily any limitation by JS or what Vue could do. 

So the solution, in these cases -- specifically when you're using options api is to:

```ts
import type {RendererElement, PropType} from 'vue'

export default {
  props: {
    to: {
      [String, Object] as PropType<RendererElement | string | null | undefined>
    }
  }
}
```
`import _type_` is important in these cases. Otherwise Vite will try to bundle them. But this is a non-issue as TS will throw errors if it's not included.

xvaara: That didn't work. I've commited a working version with HTMLElement from safeTypes.ts

xvaara:
I upgraded pnpm, nuked the pnpm cache, nuked node_modules again, and now build works... oh boy! 😅

VividLemon:
> I upgraded pnpm, nuked the pnpm cache, nuked node_modules again, and now build works... oh boy! 😅

https://github.com/bootstrap-vue/bootstrap-vue-next/pull/1054/commits/5ac58b8296e23c53566592954009c44b905e86fa did already pass though. What was the issue?

xvaara:
for some reason my local node_modules/pnpm cache was broken in some weird way that it could not build nuxt module.

packages/bootstrap-vue-next/src/components/RenderComponentOrSkip.vue
VividLemon: https://play.vuejs.org/#eNp9UMtqwzAQ/JVBF6Vg4rtRA21/oNDeqh6URg4GWxKSElpc/XtXUmIcCL1I+5rZ2ZnZk3Pb80mzjonw5QcXMSpzfJQsBskQdDy5nTTD5KyPmOF1j4Te2wmcYHxpvdjJXerbNieZldqANAfdD0a/euuCmHtrO4ToB3NMu82DNKKti2kNJVFPblRRUwaIwtoRhARVJN8rzxNJa+vEYTiXAKB2qbS1JNqFijXsquj+nasLq9Y8bo02sUH8cRpZ+zsFt6fn2/R3wRFMncbyr+GbOU+5fHmHEhedS4xC3+HjrRjS4NnaUSvzCRWWpYJHT/vwC96rMZRoX+coujW0ufJeBJFhRSiQ8kdP+s9yGVeGzpk6FdwdU6Vh6Q9DCcSe

Script setup compiles roughly to the same end code as options api does. However, when using TS & script setup, the actual compiled code does not include said constructors. It's my belief that the only thing prop constructors actually do is runtime prop type checking. However, since it would be included by TS, the actual use for constructors with TS is null as it will be caught beforehand. 

The constructors in options api that have TS aren't much use. So in this case, as long as the macro constructors include all possibilities in the PropType definition, then it will work as expected.

VividLemon:
Sometimes you just need to start anew

packages/bootstrap-vue-next/src/components/RenderComponentOrSkip.vue
```diff
@@ -8,13 +8,23 @@ export default defineComponent({
       type: String,
       default: 'div',
     },
+    to: {
+      type: [String, Object] as PropType<string | RendererElement | null | undefined>,
+      required: false,
```
VividLemon, line 13:
```suggestion
```

PR #1070, opened 2023-04-28
fix: vueuse may cause CSP error by adding style tag to DOM head

ndavenia:
# Describe the PR

In order to add 'modal-open' to body, useDark from vueuse is used. useDark uses useColorMode (https://vueuse.org/core/useColorMode/)

However useColorMode will add <style>...</style> to <head> dom which may cause CSP error.

## Small replication

Just open a BModal and in index.html set this :
<meta http-equiv="Content-Security-Policy"
          content="style-src 'self'; ">

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

### DIFF
--- a/packages/bootstrap-vue-next/src/components/BModal.vue
+++ b/packages/bootstrap-vue-next/src/components/BModal.vue
@@ -251,6 +251,7 @@ const modalOpen = useSetAttr({
   selector: 'body',
   valueDark: 'modal-open',
   valueLight: '',
+  disableTransition: false,
 })
 
 const element = ref<HTMLElement | null>(null)


VividLemon:
What affect does this have to fix the CSP error? I am curious.

aki77:
@VividLemon  

With the update to vueuse@v10, using useDark is also causing performance issues.
I think using useDark to give the modal-open class to the body element is excessive, don't you?

https://github.com/vueuse/vueuse/blob/790bcbca3a3410f0e862f890a4a91fd62ac8f3e4/packages/core/useColorMode/index.ts#L135-L185

minimal reproduction: https://stackblitz.com/edit/vue-mjnfaa
Closing the modal is very slow.

ndavenia:
> What affect does this have to fix the CSP error? I am curious. Why does this cause a CSP error in the first place? What is a CSP error 🤔

Hi,

If you go there : https://github.com/vueuse/vueuse/blob/main/packages/core/useColorMode/index.ts you will see : 
`if (disableTransition) {
        style = window!.document.createElement('style')
        style.appendChild(document.createTextNode('*{-webkit-transition:none!important;-moz-transition:none!important;-o-transition:none!important;-ms-transition:none!important;transition:none!important}'))
        window!.document.head.appendChild(style)
      }
`
So with disableTransition=true (default value) it will add a style node to the html page.
However, for security reason, some company policies add CSP (Content Security Policy) on meta or response header to prevent hot loading of style/src/assets... and with the most restrictive one "style-src 'self'" it will raise console error.

Also, I am agree with @aki77, since the use of useDark, there is performance issue with modal.

VividLemon:
> @VividLemon
> 
> With the update to vueuse@v10, using useDark is also causing performance issues. I think using useDark to give the modal-open class to the body element is excessive, don't you?
> 
> https://github.com/vueuse/vueuse/blob/790bcbca3a3410f0e862f890a4a91fd62ac8f3e4/packages/core/useColorMode/index.ts#L135-L185
> 
> minimal reproduction: https://stackblitz.com/edit/vue-mjnfaa Closing the modal is very slow.

It's not an ideal solution. I did it because it made it simple. That being said, without transition from this, any implementation would essentially be the same as whatever anyone else would produce.

aki77:
@VividLemon 
I sent a PR that fixes two issues. (#1073)

Issue #1030, opened 2023-04-12
BModal class modal-open on body disappears if two modals are open, and one closes -- leaving one open still


VividLemon:
```html
<template>
  <b-button @click="show = !show">Click me</b-button>
  <b-modal v-model="show">
    <b-button @click="show2 = !show2">Click me</b-button>
    <b-modal v-model="show2" />
  </b-modal>
</template>

<script setup lang="ts">
import {ref} from 'vue'

const show = ref(true)
const show2 = ref(false)
</script>
```

If you open the second modal, then click to close the second modal, the body class `modal-open` will disappear. This should still be active as at least one modal is open.

PR #1073, opened 2023-04-29
fix(BModal): multiple open modals and performance issues

aki77:
# Describe the PR

fix #1070 and #1030

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

### DIFF
--- a/packages/bootstrap-vue-next/src/components/BModal.vue
+++ b/packages/bootstrap-vue-next/src/components/BModal.vue
@@ -90,9 +90,9 @@
 </template>
 
 <script setup lang="ts">
-import {computed, ref, toRef, useSlots, watch} from 'vue'
-import {useBooleanish, useId} from '../composables'
-import {useEventListener, useFocus, useDark as useSetAttr, useVModel} from '@vueuse/core'
+import {computed, ref, toRef, useSlots} from 'vue'
+import {useBooleanish, useId, useModalManager} from '../composables'
+import {useEventListener, useFocus, useVModel} from '@vueuse/core'
 import type {Booleanish, ButtonVariant, ClassValue, ColorVariant, Size} from '../types'
 import {BvTriggerableEvent, isEmptySlot} from '../utils'
 import BButton from './BButton/BButton.vue'
@@ -245,13 +245,6 @@ const okOnlyBoolean = useBooleanish(toRef(props, 'okOnly'))
 const scrollableBoolean = useBooleanish(toRef(props, 'scrollable'))
 const titleSrOnlyBoolean = useBooleanish(toRef(props, 'titleSrOnly'))
 const staticBoolean = useBooleanish(toRef(props, 'static'))
-
-const modalOpen = useSetAttr({
-  attribute: 'class',
-  selector: 'body',
-  valueDark: 'modal-open',
-  valueLight: '',
-})
 
 const element = ref<HTMLElement | null>(null)
 const okButton = ref<HTMLElement | null>(null)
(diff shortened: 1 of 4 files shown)


packages/bootstrap-vue-next/src/components/BModal.vue
```diff
@@ -428,6 +429,32 @@ useEventListener(element, 'bv-toggle', () => {
 export default {
   inheritAttrs: false,
 }
+
+const useSharedModalManager = createSharedComposable(() => {
```
VividLemon, line 433: Extract all this logic out into a separate composable. 

I say create a single file. `useSharedModalManager`. Then put all the relevant logic from there into it.

VividLemon (changes requested):

packages/bootstrap-vue-next/src/components/BModal.vue
```diff
@@ -428,6 +429,32 @@ useEventListener(element, 'bv-toggle', () => {
...
+  const {count, inc, dec} = useCounter()
+  watch(count, (newValue) => {
+    if (document) {
```
VividLemon, line 436: I'm not sure if this watch would be ssr friendly. Specifically: 

1. You start a modal open
2. Ssr renders without document -- does not add classList to body
3. Sends over to client,
4. Hydration mismatch.

packages/bootstrap-vue-next/src/components/BModal.vue
```diff
@@ -428,6 +429,32 @@ useEventListener(element, 'bv-toggle', () => {
...
+      document.body.classList.toggle('modal-open', newValue > 0)
+    }
+  })
+
+  return {inc, dec}
+})
+
+function useModal() {
+  const {inc, dec} = useSharedModalManager()
+  const modalOpen = ref(false)
```
VividLemon, line 446: It doesn't appear that this value ever actually will change. I'm not sure the intended effect here

packages/bootstrap-vue-next/src/components/BModal.vue
aki77: I understand.
Do I just separate them in the `bootstrap-vue-next/src/composables` directory?

packages/bootstrap-vue-next/src/components/BModal.vue
aki77: I think the original implementation also caused hydration mismatch, am I mistaken?

packages/bootstrap-vue-next/src/components/BModal.vue
aki77: I tried to change the original implementation as little as possible.

https://github.com/bootstrap-vue/bootstrap-vue-next/blob/e2860b7080721aa811a557640115dd6484e180e0/packages/bootstrap-vue-next/src/components/BModal.vue#L418

packages/bootstrap-vue-next/src/components/BModal.vue
VividLemon: Keep them together. One can be a private function since it is only used in one place

packages/bootstrap-vue-next/src/components/BModal.vue
VividLemon: It should not have. Vueuse handles ssr afaik

packages/bootstrap-vue-next/src/components/BModal.vue
VividLemon: The actual return value of modalOpen didn't have any affect. It simply placed the class value on the body

packages/bootstrap-vue-next/src/composables/useModalManager.ts
```diff
@@ -0,0 +1,44 @@
...
+  })
+
+  watch(count, (newValue) => {
+    updateHTMLAttrs('body', 'class', newValue > 0 ? MODAL_OPEN_CLASS_NAME : '')
+  })
+
+  return {inc, dec}
+})
+
+export default function useModalManager(modalOpen: Ref<boolean>): void {
```
VividLemon, line 30:
```suggestion
export default (modalOpen: Ref<boolean>): void => {
```

aki77:
Thanks!

VividLemon:
I'm surprised I didn't think of this solution during my first implementation. Simple, yet creative

PR #1123, opened 2023-05-26
Doc formatting & vue-html card removal

StirStudios:
I have gone ahead and updated the doc .md files to fix ```html-vue and <b-card> inconsistencies. This includes:

- tabbing
- adding <template>
- removing <b-card> from ```html-vue
- adding <b-card> to code block examples
- for spinners I removed non breaking spaces and used BS margin classes. for the moment I removed any dark mode variants to make sure they render in the current mode


VividLemon:
> for spinners I removed non breaking spaces and used BS margin classes. for the moment I removed any dark mode variants to make sure they render in the current mode

`variant="dark"` is deprecated. They will all need to be removed at some point https://getbootstrap.com/docs/5.3/migration/#color-modes

PR #1125, opened 2023-05-26
Tabbing and added cards to tabs

StirStudios:
Very small updates and formatting:

- Read me updates, mainly adding <b-cards> to the tabs for better visibility.

### DIFF
--- a/apps/docs/src/getting-started.md
+++ b/apps/docs/src/getting-started.md
@@ -1,10 +1,18 @@
 # Introduction
 
+<div class="lead">
+
 Get started with BootstrapVueNext and Bootstrap `v5`, the world’s most popular framework for building responsive, mobile-first sites
 
-::: danger NOT PRODUCTION READY
+</div>
+
+<b-alert variant="danger" :model-value="true" class="my-5">
+
+NOT PRODUCTION READY
+
 This project is still in **alpha version**. There is a lot of work to do and the documentation may be out of date, if you want to contribute you can use it and submit an [issue](https://github.com/bootstrap-vue-next/bootstrap-vue-next/issues) or even better, a [pull request](https://github.com/bootstrap-vue-next/bootstrap-vue-next/pulls) 😄
-:::
+
+</b-alert>
 
 ## Why BootstrapVueNext?
 
@@ -19,8 +27,6 @@ This project is still in **alpha version** so there is a lot of work to do. If y
 
 Read our [Contribution Guide](https://github.com/bootstrap-vue-next/bootstrap-vue-next/blob/main/CONTRIBUTING.md) on how to start helping
 
----
-
 ## Install
 
 ### Installation - Vue.js
@@ -33,10 +39,10 @@ Read our [Contribution Guide](https://github.com/bootstrap-vue-next/bootstrap-vu
 Install the necessary packages for `BootstrapVueNext`:
 
 <client-only>
-<b-tabs v-model="codePreference" class="mb-3">
+<b-tabs v-model="codePreference">
   <b-tab title="PNPM">
 
-  <b-card class="border-top-0 rounded-top-0">
+  <b-card>
 
 ```bash
 pnpm add bootstrap bootstrap-vue-next
@@ -49,7 +55,7 @@ pnpm add unplugin-vue-components -D
   </b-tab>
   <b-tab title="YARN">
 
-  <b-card class="border-top-0 rounded-top-0">
+  <b-card>
 
 ```bash
 yarn add bootstrap bootstrap-vue-next
@@ -62,7 +68,7 @@ yarn add unplugin-vue-components -D
   </b-tab>
   <b-tab title="NPM">
 
-  <b-card class="border-top-0 rounded-top-0">
+  <b-card>
 
 ```bash
 npm i bootstrap bootstrap-vue-next
@@ -77,6 +83,8 @@ npm i unplugin-vue-components -D
 </client-only>
 
 The following is an example of a basic `vite.config.js/ts`. All you need to do is add **Components** to the Vite **plugins** option, with the additional imports:
+
+<b-card>
 
 ```ts
 // vite.config.js/ts
@@ -95,7 +103,11 @@ export default defineConfig({
 })
 ```
 
+</b-card>
+
 Finally, in your `main.js/ts` import the CSS:
+
+<b-card>
 
 ```ts
 // main.js/ts
@@ -103,9 +115,13 @@ import 'bootstrap/dist/css/bootstrap.css'
 import 'bootstrap-vue-next/dist/bootstrap-vue-next.css'
 ```
 
+</b-card>
+
 ##### Typescript Features
 
 If using TypeScript you will want to add `components.d.ts` to the `include` array in your tsconfig.json:
+
+<b-card>
 
 ```json
 // tsconfig.json
@@ -114,7 +130,7 @@ If using TypeScript you will want to add `components.d.ts` to the `include` arra
 }
 ```
 
----
+</b-card>
 
 #### Legacy Installation
 
@@ -123,10 +139,10 @@ If using TypeScript you will want to add `components.d.ts` to the `include` arra
 Install the necessary packages for `bootstrap-vue-next`:
 
 <client-only>
-<b-tabs v-model="codePreference" class="mb-3">
+<b-tabs v-model="codePreference">
   <b-tab title="PNPM">
 
-  <b-card class="border-top-0 rounded-top-0">
+  <b-card>
 
 ```bash
 pnpm add bootstrap bootstrap-vue-next
@@ -137,7 +153,7 @@ pnpm add bootstrap bootstrap-vue-next
   </b-tab>
   <b-tab title="YARN">
 
-  <b-card class="border-top-0 rounded-top-0">
+  <b-card>
 
 ```bash
 yarn add bootstrap bootstrap-vue-next
@@ -162,6 +178,8 @@ npm i bootstrap bootstrap-vue-next
 
 Then, add to your `main.js/ts`:
 
+<b-card>
+
 ```typescript
 // main.js/ts
 import App from './App.vue' // Can be a different place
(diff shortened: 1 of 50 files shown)


apps/docs/src/getting-started.md
```diff
@@ -1,11 +1,19 @@
...
 
+<div class="lead">
+
 Get started with BootstrapVueNext and Bootstrap `v5`, the world’s most popular framework for building responsive, mobile-first sites
 
+</div>
+
+<b-alert variant="danger" :model-value="true">
+
 ::: danger NOT PRODUCTION READY
```
VividLemon, line 11: Previously the `:::` syntax did this automatically. However, the `:::` syntax was automatically generated by the _theme_. I had reached out to Vitepress in hopes to seeing how and if we could essentially make `:::` _compile to_ a custom component, but have received no response.

apps/docs/src/reference/directives/BTooltip.md
```diff
@@ -16,21 +16,21 @@ So, here we are defining to use the directive "b-tooltip", with different option
 
 We can define when we want to trigger a tooltip with the following modifiers.
 
-* click
-* hover
-* focus
-* manual
+- click
```
VividLemon, line 19: If we're going to pick `-` over `*`, it should really be done _globally_... Just to stay consistent.

VividLemon (commented):
Other than that, I don't have anything else. But even those two things aren't dealbreakers. I could see arguments for doing it either way.

apps/docs/src/getting-started.md
VividLemon: ("Doing it automatically" was the _themes_ custom implementation of this, not necessarily a bootstrap implementation. Alert is the closest component that matched what `:::` did)

StirStudios:
Thanks for the feedback. I have gone ahead and removed the BS alert. However the * vs - seems more like a linting prepare issue. This happens on commit staging. I will see if i can revert to avoid.

apps/docs/src/reference/directives/BTooltip.md
StirStudios: However the * vs - seems more like a linting prepare issue. This happens on commit staging. I will see if i can revert to avoid.

StirStudios: Reviewing the code it seems in most cases - it used over * so I went ahead and converted the remaining * versions.

VividLemon: Interesting, if it happens again, perhaps we will use the other syntax than.

apps/docs/src/getting-started.md
VividLemon: I've recently gotten a response from a Vitepress core team member about a few of these issues. This particular issue can be fixed by modifying the behavior of markdown-it. However, this particular fix isn't quite extendable to other syntaxes (say `> desc`) type syntax (afaik). So, it's my super executive decision, that moving forward, the usage should be as you have done previously, with using true html tags to provide this functionality. 

If you don't mind, could this PR address fixes for `> desc` syntax and `:::` syntax? 

* When I say `> desc` syntax, I am referring to those top level descriptions that are present throughout things like the component references

StirStudios: Okay to be clear that is what is needed:

Update `:::`
```
::: tip
The animation effect of this component is dependent on the prefers-reduced-motion media query. See the [reduced motion section of our accessibility documentation](https://getbootstrap.com/docs/5.0/getting-started/accessibility/#reduced-motion).
:::
```

to
```
<b-alert variant="info" :model-value="true">
::: tip
The animation effect of this component is dependent on the prefers-reduced-motion media query. See the [reduced motion section of our accessibility documentation](https://getbootstrap.com/docs/5.0/getting-started/accessibility/#reduced-motion).
:::
<b-alert>
```

---

Top level descriptions:
```
# Accordion

Build vertically collapsing accordions in combination with `<b-collapse>` component.
```

to

```
<p class="lead">
# Accordion

Build vertically collapsing accordions in combination with `<b-collapse>` component.
</p>
```

Or possibly a blockquote?

VividLemon: Only change is 
```
<b-alert variant="info" :model-value="true">
::: tip
The animation effect of this component is dependent on the prefers-reduced-motion media query. See the [reduced motion section of our accessibility documentation](https://getbootstrap.com/docs/5.0/getting-started/accessibility/#reduced-motion).
:::
<b-alert>
```
to 
```
<b-alert variant="info" :model-value="true">
The animation effect of this component is dependent on the prefers-reduced-motion media query. See the [reduced motion section of our accessibility documentation](https://getbootstrap.com/docs/5.0/getting-started/accessibility/#reduced-motion).
<b-alert>
```

StirStudios: Awesome on it now then!

StirStudios: I have gone ahead and made a few more adjustments for design and UI, makes things easier to read: https://github.com/bootstrap-vue-next/bootstrap-vue-next/pull/1125/commits/28297b50b7d4d94db45682975c2e327380cf5c1e

If this works I can move through and update the rest.

apps/docs/src/reference/components/form-group.md
```diff
@@ -393,28 +393,26 @@ You can make a floating label by setting the property `floating` to true and spe
...
-  <b-card>
-    <div>
-      <b-form-group
-        id="fieldset-1"
-        description="Let us know your name."
-        label="Name"
-        label-for="input-floating-1"
-        valid-feedback="Thank you!"
-        :invalid-feedback="floatingInvalidFeedback"
+  <div>
```
VividLemon, line 396: In addition to removals of <b-cards> around code blocks, it would also be helpful to remove excess div tags as well. 

Also, occasionally I have made some code like
```html
<b-button>do one thing</b-button>
<b-button>do something else</b-button>
```
When it could be wrapped in a button group
```html
<b-button-group>
<b-button>do one thing</b-button>
<b-button>do something else</b-button>
</b-button-group>
```
I'm not sure where some of these instances are, but if you ever find one, it would look nicer if they were grouped up... I think I did one on the useBreadcrumb page that I changed today.

apps/docs/src/reference/directives/BTooltip.md
```diff
@@ -79,40 +81,40 @@ When we are using a directive, we have two ways to define the title to use in th
 
 ### Incorrect use
 
-````vue-html
+```vue-html
```
VividLemon, line 84: Nice catch

apps/docs/src/reference/icons.md
```diff
@@ -17,27 +17,39 @@ The preferred installation makes use of [unplugin-vue-components](https://github
 To start, install the necessary packages:
 
 <client-only>
-<b-tabs v-model="codePreference">
+<b-tabs v-model="codePreference" class="mb-3">
   <b-tab title="PNPM">
 
+  <b-card class="border-top-0 rounded-top-0">
```
VividLemon, line 23: If you have any other suggestions on making the docs look "nicer", I am all for seeing them. It could be a large or small change, I'm open to anything.

packages/bootstrap-vue-next/CHANGELOG.md
```diff
@@ -1,749 +1,699 @@
...
 
-
 ## [0.8.15](https://github.com/bootstrap-vue-next/bootstrap-vue-next/compare/bootstrap-vue-next-v0.8.14...bootstrap-vue-next-v0.8.15) (2023-05-26)
 
-
 ### Bug Fixes
 
-* **BDropdown:** using manual control to display/hide the popover it should allow the event update:modelValue when manualBoolean is true ([ba490c9](https://github.com/bootstrap-vue-next/bootstrap-vue-next/commit/ba490c952a1f2fc16cd1a2beca2bc1b663d6ebc5))
-* **BPopover:** using manual control to display/hide the popover it should allow the event update:modelValue when manualBoolean is true ([d169360](https://github.com/bootstrap-vue-next/bootstrap-vue-next/commit/d16936008c40fc22a7a2a5ccc268d27011171d2b))
+- **BDropdown:** using manual control to display/hide the popover it should allow the event update:modelValue when manualBoolean is true ([ba490c9](https://github.com/bootstrap-vue-next/bootstrap-vue-next/commit/ba490c952a1f2fc16cd1a2beca2bc1b663d6ebc5))
```
VividLemon, line 7: You're changelog is out of sync. During the fixing of the merge conflict, just copy the existing one on the repo, then paste it into yours. 

Basically, the * to - fixes are fine, but you can leave the changelog alone

packages/nuxt/CHANGELOG.md
```diff
@@ -2,29 +2,26 @@
 
 ## [0.0.7](https://github.com/bootstrap-vue-next/bootstrap-vue-next/compare/nuxt-v0.0.6...nuxt-v0.0.7) (2023-05-17)
 
-
 ### ⚠ BREAKING CHANGES
 
-* **nuxt:** remove nuxt.ts from main package & @nuxt/kit from dependencies -- use package @bootstrap-vue-next/nuxt (readme coming soon)
-* **nuxt:** create explicit nuxt module
-* **BDropdown:** do not use internal state to modify dropdown instead expose vmodel
+- **nuxt:** remove nuxt.ts from main package & @nuxt/kit from dependencies -- use package @bootstrap-vue-next/nuxt (readme coming soon)
```
VividLemon, line 7: Same as above, overall, just leave the changelog alone. I figure it was a global find/replace and just an oversight.

VividLemon (changes requested):

apps/docs/src/reference/components/form-group.md
VividLemon: I could also just do this one tomorrow.

apps/docs/src/reference/icons.md
StirStudios: I think I have done as much as possible to this point. I would love to get your feedback on the TOC offcanvas idea I put together. Let's see if we can get a merge happening here.

VividLemon: There's still a merge conflict

StirStudios: Ah, did a quick copy and paste and seems to be passing now. Will add the tiny theming tweaks for the useBreadcrumb.md later. Fingers crossed this time!

StirStudios:
@VividLemon seems there is a build failure not sure that is something I can fix?

VividLemon:
> @VividLemon seems there is a build failure not sure that is something I can fix?

A component in BTooltip.md is missing an end tag

StirStudios:
Always something!

VividLemon:
> Always something! It's fixed but is still failing it seems!

image

PR #1135, opened 2023-05-28
Added in clipboard icon

StirStudios:
Added a clipboard copy icon. It is a simple SVG.  You might prefer using css svg from the icons you use in the app I was just not sure how to find the css encode. Regardless the icon works in both color modes and has a slight animation and color change once it has been copied.


VividLemon:
You could use bootstrap-icons copy icon to have it be "in house"

PR #1138, opened 2023-05-28
feat(BDropdown): allow define aria-label to b-button

sfe-efficy:
# Describe the PR

allow define aria-label to b-button (and same using b-dropdown)

## Small replication

## PR checklist

- [x] Feature - `feat(...)`
- [ ] Other (please describe)

BEGIN_COMMIT_OVERRIDE
feat(BDropdown): allow define aria-label to b-button
END_COMMIT_OVERRIDE

### DIFF
--- a/packages/bootstrap-vue-next/src/components/BButton/BButton.vue
+++ b/packages/bootstrap-vue-next/src/components/BButton/BButton.vue
@@ -93,8 +93,8 @@ export default defineComponent({
     ])
 
     const computedAttrs = computed(() => ({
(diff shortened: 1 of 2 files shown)


VividLemon:
image
Aria label is treated as an attr fallthrough

sfe-efficy:
Ok, it's true for a BButton, maybe for the BDropdown, we should allow define the aria-label for the deeper button generated, at this moment it's defined in the div, but this is not an interactic element, right?

PR #1146, opened 2023-05-29
Code Copy Theming Improvements

StirStudios:
- Small update to optimize the copy code experience, css.
- You can see what type of code snippet it is: ts, vue, etc
- Fixed card body missing import.


apps/docs/src/reference/icons.md
```diff
@@ -1,7 +1,11 @@
 # Icons
 
+<div class="lead mb-5">
```
VividLemon, line 3: Make this an alert

apps/docs/src/components/HighlightCard.vue
```diff
@@ -13,3 +13,67 @@
 <script setup lang="ts">
 import {BCard, BCardBody} from 'bootstrap-vue-next'
 </script>
+
+<style lang="scss">
```
VividLemon, line 17: Move this back to the layout file

VividLemon (changes requested):

apps/docs/src/components/HighlightCard.vue
StirStudios: Done!

apps/docs/src/reference/icons.md
StirStudios: Done!

PR #1147, opened 2023-05-29
Offcanvas TOC nav

StirStudios:
Thought this was ready for a PR:

1. Uses offcanvas and toggle
2. On collapse it adds extra nav links
3. Add a nav link component to create all the links
4. Works as designed for desktop and small screen sizes

Note: not sure the best way to hide offcanvas on router nav click. Right now you can click through to everything but you need to manually close the offcanvas.

I think this adds a much improved UI experience.

Looking forward to your feedback.

### DIFF
--- a/apps/docs/.vitepress/theme/Layout.vue
+++ b/apps/docs/.vitepress/theme/Layout.vue
@@ -3,8 +3,8 @@
     <b-navbar-toggle @click="toggler" />
     <b-collapse is-nav>
       <b-navbar-nav>
-        <b-navbar-brand :to="withBase('/')"
-          ><svg
+        <b-navbar-brand :to="withBase('/')">
+          <svg
             xmlns="http://www.w3.org/2000/svg"
             viewBox="0 0 953 953"
             width="32"
@@ -21,8 +21,9 @@
             <path
               fill="currentColor"
               d="M92 0h769c50 0 92 42 92 92v769c0 50-42 92-92 92H92c-50 0-92-42-92-92V92C0 42 42 0 92 0zm216 710c100 0 160-50 160-133 0-62-44-107-108-113v-3c48-8 86-52 86-102 0-71-55-117-140-117H111v468h197zM195 307h90c50 0 78 23 78 64 0 44-33 68-91 68h-77V307zm0 338V499h90c64 0 98 25 98 73s-33 73-94 73h-94zm503 65l163-468h-90L652 621h-2L531 242h-92l163 468h96z"
-            ></path></svg
-        ></b-navbar-brand>
+            ></path>
+          </svg>
+        </b-navbar-brand>
         <b-nav>
           <b-nav-item :to="withBase('/getting-started')">Getting Started</b-nav-item>
           <b-nav-item :to="withBase('/reference/icons')">Icons</b-nav-item>
--- a/apps/docs/.vitepress/theme/Layout.vue
+++ b/apps/docs/.vitepress/theme/Layout.vue
@@ -3,7 +3,26 @@
     <b-navbar-toggle @click="toggler" />
     <b-collapse is-nav>
       <b-navbar-nav>
-        <b-navbar-brand :to="withBase('/')">Home</b-navbar-brand>
+        <b-navbar-brand :to="withBase('/')"
+          ><svg
+            xmlns="http://www.w3.org/2000/svg"
+            viewBox="0 0 953 953"
+            width="32"
+            height="32"
+            version="1.1"
+            shape-rendering="geometricPrecision"
+            fill-rule="evenodd"
+            clip-rule="evenodd"
+            focusable="false"
+            role="img"
+            class="d-block"
+          >
+            <title>BootstrapVue Next</title>
+            <path
+              fill="currentColor"
+              d="M92 0h769c50 0 92 42 92 92v769c0 50-42 92-92 92H92c-50 0-92-42-92-92V92C0 42 42 0 92 0zm216 710c100 0 160-50 160-133 0-62-44-107-108-113v-3c48-8 86-52 86-102 0-71-55-117-140-117H111v468h197zM195 307h90c50 0 78 23 78 64 0 44-33 68-91 68h-77V307zm0 338V499h90c64 0 98 25 98 73s-33 73-94 73h-94zm503 65l163-468h-90L652 621h-2L531 242h-92l163 468h96z"
+            ></path></svg
+        ></b-navbar-brand>
         <b-nav>
           <b-nav-item :to="withBase('/getting-started')">Getting Started</b-nav-item>
           <b-nav-item :to="withBase('/reference/icons')">Icons</b-nav-item>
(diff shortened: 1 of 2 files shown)


apps/docs/src/components/TableOfContentsNav.vue
```diff
@@ -0,0 +1,133 @@
...
+  <b-list-group>
+    <strong class="bd-links-heading">
+      <b-link :to="withBase('/reference/directives')">Directives</b-link>
+    </strong>
+    <b-list-group-item v-for="component in directivesComputedList" :key="component.name">
+      <b-link :to="component.route">{{ component.name }}</b-link>
+    </b-list-group-item>
+  </b-list-group>
+  <b-list-group v-if="!isLargeScreen">
+    <strong class="bd-links-heading">General</strong>
```
VividLemon, line 32: Move this section to the top

apps/docs/.vitepress/theme/Layout.vue
```diff
@@ -1,95 +1,104 @@
...
+        </b-dropdown>
+      </client-only>
+    </b-nav>
+  </b-navbar>
+  <b-container fluid class="container-lg mt-3 my-md-4 bd-layout">
+    <aside class="bd-sidebar">
+      <b-offcanvas
+        v-model="sidebar"
+        static="true"
+        :backdrop="isLargeScreen ? 'static' : ''"
```
VividLemon, line 64: I'm not sure the intended purpose of 'static' is. BOffcanvas prop backdrop is a boolean type

apps/docs/.vitepress/theme/Layout.vue
```diff
@@ -1,95 +1,104 @@
...
+  <b-navbar variant="primary" sticky="top" toggleable="lg">
+    <b-navbar-toggle @click="toggler" />
+    <b-collapse is-nav>
+      <b-navbar-nav>
+        <b-navbar-brand :to="withBase('/')">Home</b-navbar-brand>
+        <b-nav>
+          <b-nav-item :to="withBase('/getting-started')">Getting Started</b-nav-item>
+          <!-- <b-nav-item :to="withBase('/reference/components')">Components</b-nav-item>
+          <b-nav-item :to="withBase('/reference/composables')">Composables</b-nav-item>
+          <b-nav-item :to="withBase('/reference/directives')">Directives</b-nav-item> -->
```
VividLemon, line 11:
```suggestion
```

VividLemon (changes requested):

apps/docs/src/components/TableOfContentsNav.vue
StirStudios: Done!

apps/docs/.vitepress/theme/Layout.vue
StirStudios: We need this otherwise the offcanvas on large devices closes when clicked outside the area. Not sure how it is working but it is.

apps/docs/.vitepress/theme/Layout.vue
StirStudios: Done!

StirStudios:
The last part is how do we hide the offcanvas on menu click. Could we add this to the watcher?

// Hide sidebar on route change.
watch(route, () => {
  sidebar.value = false
})

apps/docs/.vitepress/theme/Layout.vue
VividLemon: `:backdrop="isLargeScreen ? 'static' : ''"` 
static is not a valid value.
`:backdrop="isLargeScreen ? true : false"`

VividLemon:
> The last part is how do we hide the offcanvas on menu click. Could we add this to the watcher?
> 
> // Hide sidebar on route change. watch(route, () => { sidebar.value = false })

I'd do something like 

```ts
import {useRoute} from 'vitepress'

const {path} = useRoute()

watch(path, () => {
  sidebar.value = false
})
```

apps/docs/.vitepress/theme/Layout.vue
StirStudios: Perhaps I am making something work accidentally:

https://getbootstrap.com/docs/5.3/components/offcanvas/#static-backdrop

We need this to allow no clicking outside the area. Somehow I guess I have it working. As soon as I switch to true/false clicking outside closes the offcanvas. Not sure how to deal with this one.

StirStudios:
I tried your suggestion:

```
import {useRoute} from 'vitepress'

const {path} = useRoute()

watch(path, () => {
  sidebar.value = false
})
```

However, I get the following issue

```
No overload matches this call.
  The last overload gave the following error.ts(2769)
runtime-core.d.ts(1226, 25): The last overload is declared here.
const path: string
```

apps/docs/.vitepress/theme/Layout.vue
VividLemon: Vue will consolidate truthy values into what makes sense "HTML"-wise. 

Accidentally you are causing this code:

```ts
export const resolveBooleanish = (input: Booleanish): boolean =>
  typeof input === 'boolean' ? input : input === '' ? true : input === 'true' ? true : false
```
Where, since static is not, boolean, 'true' or 'false', it just uses the last in the chain -- false. It might get a little more complicated than that, with all the moving parts, it might be lost.

Static backdrop is achieved through the `noCloseOnBackdrop`. I'm not exactly sure how I feel about the naming of this prop. On one hand, it aligns with the other prop -- noCloseOnEsc, which doesn't have a getbootstrap specific counterpart. In Bootstrap, these two props are combined into the term "static". Our version gives a bit more flexibility. But in addition, the bootstrap "static" term doesn't really make much sense, as our "static" prop refers to rendering teleports in place, rather than attaching to the body element... So, by modifying in any direction, we'd have to adjust for breaking changes in other places -- it's a case of, is it better to keep flexibility and stick to bootstrap-vue, or is it better to stay strictly to bootstrap... In my view, generally sticking to bootstrap is preferred, but when in this case, the better developer experience with the flexibility pushes it over the edge to have it stay how it is.

VividLemon:
> I tried your suggestion:
> 
> ```
> import {useRoute} from 'vitepress'
> 
> const {path} = useRoute()
> 
> watch(path, () => {
>   sidebar.value = false
> })
> ```
> 
> However, I get the following issue
> 
> ```
> No overload matches this call.
>   The last overload gave the following error.ts(2769)
> runtime-core.d.ts(1226, 25): The last overload is declared here.
> const path: string
> ```

```ts
const route = useRoute()

watch(() => route.path, () => {
  console.log('foo')
})
```

This should work. I guess path isn't a ref

apps/docs/.vitepress/theme/Layout.vue
StirStudios: Right makes complete sense. I reviewed what I was doing and decided to just set backdrop to false as it solves the problem regardless.

StirStudios:
@VividLemon I think we are good. I added in the BV svg instead of home. I  thought it was a good call, but was not totally sure.

PR #1152, opened 2023-05-30
docs: minor fixes and improvements

VividLemon:
# Describe the PR

## Small replication

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

### DIFF
--- a/apps/docs/.vitepress/theme/Layout.vue
+++ b/apps/docs/.vitepress/theme/Layout.vue
@@ -21,7 +21,7 @@
             <path
               fill="currentColor"
               d="M92 0h769c50 0 92 42 92 92v769c0 50-42 92-92 92H92c-50 0-92-42-92-92V92C0 42 42 0 92 0zm216 710c100 0 160-50 160-133 0-62-44-107-108-113v-3c48-8 86-52 86-102 0-71-55-117-140-117H111v468h197zM195 307h90c50 0 78 23 78 64 0 44-33 68-91 68h-77V307zm0 338V499h90c64 0 98 25 98 73s-33 73-94 73h-94zm503 65l163-468h-90L652 621h-2L531 242h-92l163 468h96z"
-            ></path>
+            />
           </svg>
         </b-navbar-brand>
         <b-nav>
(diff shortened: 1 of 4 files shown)


VividLemon:
@StirStudios 

For a small description of some of the things that I changed in this PR. So you can learn from the problems they cause

First:
image
This code was removed in favor of an alternative... When you do `onMounted(() => {someValue.value = someOtherValue}`
This is problematic code because it will "blink".
In between the creation of the target, and the termination of the onMounted there exist a point where sidebar's value is 'true', and the finishing of the onMounted callback, so very briefly, it is true, then proceeds to be false. A common issue for some, which is not often noticed, is that onMounted does not wait for the callback to finish, before proceeding to render the element. A common issue that occurs is some expect async onMounted to wait for the Promise to resolve, when in reality, the element being mounted doesn't wait for onMounted to finish. The same issue occurs here, as for a split second, (more noticable on a mobile screen), the offcanvas appears, then proceeds to disappear quickly... Thus causing a "blink". 

The second issue was the watching multiple targets, but treating them slightly differently... While this code didn't necessarily have any issues, these elements likely didn't need to be coupled together. A much more effective way to use mutliple watch targets is when they really do the same thing, like so:
image
These are in the same watcher, as they both treat and do the same things. So, it's best to keep them uncoupled. 

The issue that you raised was a hydration issue, I'm 99% certain that it's caused because you're rendering things differently based on `useMediaQuery('(min-width: 992px)')` -- This is not an issue in plain Vue, but is an issue when things are SSR or SSG created. When the documentation builds, it creates and stores an finished html file, but how does the media query behave in an SSG environment when there is no viewport? It is likely 0px... So, if the input device has the ref from useMediaQuery be true, and the built html file is false, well, then there's a hydration mismatch... Sometimes they can be harmless, but should always be fixed, and sometimes they make it so you can't interact with a button.

This PR doesn't fix the hydration issue just yet. That one may be tough, as I have to weight the benefits of just making that particular code as client-only... The only problem I have is rendering "nav" elements as client only... As with anything, if you're going to use SSR to achieve better SEO, the one thing you really don't want to make client only is the nav elements. So, I'm not sure the best fix yet, or if I should really not care about SEO all too much for the documentation.

StirStudios:
Dam you are good at explaining everything. Incredibly helpful. And you did mention the SSR thing. I mainly deal with Nuxt and Nuxt 3 helps even more in SEO department.

I apologize for adding another issue into the mix...

VividLemon:
> Dam you are good at explaining everything. Incredibly helpful. And you did mention the SSR thing. I mainly deal with Nuxt and Nuxt 3 helps even more in SEO department.
> 
> I apologize for adding another issue into the mix...

I don't mind the issues. Its better than going at it alone.

Vitepress I'm pretty sure is SSG. It does generate true HTML files after all. The difference only being that one renders HTML at build time, and one renders on the fly... So whatever consequences that occur from that (say if you had a login state, ssg would not be a good choice as it would certainly be a hydration issue)

PR #1155, opened 2023-05-30
Check offcanvas to remain open on desktop

StirStudios:
We need this extra check in place to help keep the menu open on desktops, otherwise, it closes and we only want that behavior when offvanvas is activated (collapse mode).

### DIFF
--- a/apps/docs/.vitepress/theme/Layout.vue
+++ b/apps/docs/.vitepress/theme/Layout.vue
@@ -287,6 +287,7 @@ const globalData = inject(appInfoKey, {
   // Short term fix for navbar modes.
   .navbar {
     color: #fff;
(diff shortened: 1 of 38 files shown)


apps/docs/.vitepress/theme/Layout.vue
```diff
@@ -1,29 +1,29 @@
...
+        height="32"
+        version="1.1"
+        shape-rendering="geometricPrecision"
+        fill-rule="evenodd"
+        clip-rule="evenodd"
+        focusable="false"
+        role="img"
+        class="d-block"
+      >
+        <title>BootstrapVue Next</title>
```
VividLemon, line 18:
```suggestion
        <title>BootstrapVueNext</title>
```

PR #1161, opened 2023-05-31
SCSS theme implimentation

StirStudios:
Introduced custom scss setup for better dev theming for the docs.

- Moved app scss to _general.scss
- Commented out BS dist.css
- Imported our own compiled styles.css
- Introduced purple color based on BS (nothing set in stone) to make sure all was working

If this works, then smooth sailing!

Let me know if you have any feedback.


VividLemon:
This doesn't seem like the most straightforward implementation. I more prefer the overriding solution. Perhaps you could move the overrides into a separate scss file rather than pulling it onto the layout file, without all the extra files?

StirStudios:
It is possible, I guess I would just do the SCSS compile locally and push the compiled css file to git. That way we only have 1 css file. If someone wants to make any css updates that would be the process they need to follow. I guess we can cross that bridge when the time comes!

Does that work?

StirStudios:
I have proposed another idea to help keep things lean and still tracked. Let me know if this could work.

VividLemon:
I'm thinking the best solution is to:
1. Import the bootstrap styles again
2. Write overrides in a scss file
3. Install sass
It's essentially the same at what exists now, but moved into it's own file.

StirStudios:
Got it, the issue will be if you want to perform things based on colors for example. So if we want to set a new primary color it would have to be done in various places as apposed to the variables file, and done - 1 line. Well that is how we do it, so things like --var-etc work without overrides in various places.

I guess the doc part is so small it does not matter, I was just trying to be cleaner with our approach.

Either way, your call.

StirStudios:
Reverted updates.

VividLemon:
So, what is the issue that this seeked to solve?

StirStudios:
The ultimate idea was to use our own compiled css file. The current setup works but we have to override instead of utilizing bs variables to set our theme and then compile. However, right now I am only seeing this for color at the moment so it may not be worth it.

PR #1164, opened 2023-06-02
Small theming update for purple navbar

StirStudios:
Very small update to include a purple navbar like BS with some extra vars and commenting.


VividLemon:
I'm not opposed to using a different color

StirStudios:
@VividLemon what about this type of thing?

SCR-20230603-piqo

VividLemon:
> @VividLemon what about this type of thing?
> 
> SCR-20230603-piqo

I've never been a huge fan of large gradient changes. I've always found it to be a bit gimmicky. I'm not opposed to a slight gradient, if you'd like to lower the percentage.

StirStudios:
Absolutely agree, as it is a gimmic! So can we get away from both of these colors? Does it matter?

Some Options:

Option 1
SCR-20230604-mzea

Option 2 (total mistake but thought it was cool in some way)
SCR-20230604-mzzk

Option 3
SCR-20230604-naoc

Option 4
SCR-20230604-nbop

Option 5
SCR-20230604-ncns

Option 6
SCR-20230604-ndch

Option 7
SCR-20230604-ndyb

Option 8
SCR-20230604-ndyb

Option 9
SCR-20230604-ngcr

I think a subtle gradient adds a little flavour. That is me, thoughts?

VividLemon:
I think the first one look nice

StirStudios:
Done!

Issue #1178, opened 2023-06-09
Ability to se "to" property for components using Teleport.


phlegx:
**Is your feature request related to a problem? Please describe.**

Like described in the official [documentation of vue](https://vuejs.org/guide/scaling-up/ssr.html#teleports), SSR deployments should consider to use an own DOM node for teleports:

_Avoid targeting body when using Teleports and SSR together - usually, `<body>` will contain other server-rendered content which makes it impossible for Teleports to determine the correct starting location for hydration. Instead, prefer a dedicated container, e.g. `<div id="teleported"></div>` which contains only teleported content._

Every bootstrap-vue component, that uses teleport, cannot be configured with `:to=` property. Default is `body` and SSR deployments need to set teleport `:to=` to e.g. `#body-teleports`.

We have two possible solutions:

1. Teleport components are rendered only on client side (no hydration node mismatch).

```html
<ClientOnly>
  <BOffcanvas>
    ...
  </BOffcanvas>
</ClientOnly> 
```

2. Required SSR rendering of teleport components, by using a unique DOM node outside app node (solves hydration node mismatch).

```html
<b-offcanvas to="#body-teleports">
  ...
</b-offcanvas>
```

with `index.html` like:

```html
...
<body>
  <div id="body-teleports"></div>
  <div id="app"></div>
</body>
```

SSR rendering does something like:

```js
teleported = context.teleports['#body-teleports']
...
// Replace <div id="body-teleports"> with <div id="body-teleports" data-server-rendered="true">${teleported}
```

**Describe the solution you would like.**

Ability to config `to` property for components using `Teleport` (e.g. `BOffcanvas`).

**Describe alternatives you have considered.**

An alternative is to wrap bootstrap-vue components using `Teleport` with `ClientOnly`. But that eliminates a huge bonus from using SSR, as web crawlers not able to traverse JS won't be able to index website correctly.

**Additional context.**

Trying to add teleports in body (SSR side) causes twice teleports in client body. (Vue docs: _`<body>` will contain other server-rendered content which makes it impossible for Teleports to determine the correct starting location for hydration._)

```log
[Vue warn]: Hydration children mismatch in <body>: server rendered element contains fewer child nodes than client vdom. 
  at <BOffcanvas class="lb-header-oc" id="lb-header-oc" modelValue=true  ... >
  ...
```

Vite-SSR PR: frandiox/vite-ssr/pull/207

PR #1188, opened 2023-06-14
feat(BModal): adds teleport to property.

phlegx:
# Describe the PR

Like described in the official [documentation of vue](https://vuejs.org/guide/scaling-up/ssr.html#teleports), SSR deployments should consider to use an own DOM node for teleports:

_Avoid targeting body when using Teleports and SSR together - usually, `<body>` will contain other server-rendered content which makes it impossible for Teleports to determine the correct starting location for hydration. Instead, prefer a dedicated container, e.g. `<div id="teleported"></div>` which contains only teleported content._

Every bootstrap-vue component, that uses teleport, cannot be configured with `:to=` property. Default is `body` and SSR deployments need to set teleport `:to=` to e.g. `#body-teleports`.

**This PR introduces `to` property for `BModal`.**

Solves issue #1178 for `BModal`.

Teleport Interface: [src/components/Teleport.ts#L19](https://github.com/vuejs/core/blob/60cd23c7520f9098bf31fc2d0c09d58ded75f173/packages/runtime-core/src/components/Teleport.ts#L19)

## Small replication

Required SSR rendering of teleport components, by using a unique DOM node outside app node (solves hydration node mismatch).

```html
<b-modal to="#body-teleports">
  ...
</b-modal>
```

with `index.html` like:

```html
...
<body>
  <div id="body-teleports"></div>
  <div id="app"></div>
</body>
```

## PR checklist

- [x] Feature - `feat(...)`
- [ ] Other (please describe)

### DIFF
--- a/apps/playground/src/Views/HomePage.vue
+++ b/apps/playground/src/Views/HomePage.vue
@@ -79,6 +79,7 @@ import TListGroup from '../components/Comps/TListGroup.vue'
 import TModal from '../components/Comps/TModal.vue'
 import TNav from '../components/Comps/TNav.vue'
 import TNavbar from '../components/Comps/TNavbar.vue'
+import TOffcanvas from '../components/Comps/TOffcanvas.vue'
 import TOverlay from '../components/Comps/TOverlay.vue'
 import TPagination from '../components/Comps/TPagination.vue'
 import TPopover from '../components/Comps/TPopover.vue'
@@ -118,6 +119,7 @@ const comps: {name: string; is: unknown; disabled?: true}[] = [
   {name: 'Modal', is: TModal},
   {name: 'Nav', is: TNav},
   {name: 'Navbar', is: TNavbar},
+  {name: 'Offcanvas', is: TOffcanvas},
   {name: 'Overlay', is: TOverlay},
   {name: 'Pagination', is: TPagination},
   {name: 'Popover', is: TPopover},
--- a/apps/playground/src/components/Comps/TOffcanvas.vue
+++ b/apps/playground/src/components/Comps/TOffcanvas.vue
@@ -1,0 +1,15 @@
+<template>
+  <div>
+    <b-offcanvas v-model="showOffcanvas" title="Title">
+      content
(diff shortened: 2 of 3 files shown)


packages/bootstrap-vue-next/src/components/BModal.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <teleport to="body" :disabled="staticBoolean">
+  <teleport :to="props.to" :disabled="staticBoolean">
```
VividLemon, line 2: Rename prop `static` to prop `teleportDisabled` 

Declare this a breaking change

Rename prop `to` to prop `teleportTo`

> Rename prop `static` to prop `teleportDisabled` 

Do this to all instances of teleports. (It's only this and BOffcanvas)

packages/bootstrap-vue-next/src/components/BModal.vue
```diff
@@ -158,6 +158,7 @@ interface BModalProps {
   titleTag?: string
   static?: Booleanish
   autoFocusButton?: 'ok' | 'cancel' | 'close'
+  to?: string | RendererElement | null | undefined
```
VividLemon, line 161: I'm not sure where you're getting these prop definitions from. The Api documentation is located here
https://vuejs.org/api/built-in-components.html#teleport
You may prefer `to?: string | HTMLElement`

VividLemon (changes requested):

packages/bootstrap-vue-next/src/components/BModal.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <teleport to="body" :disabled="staticBoolean">
+  <teleport :to="props.teleportTo" :disabled="teleportDisabledBoolean">
```
VividLemon, line 2:
```suggestion
  <teleport :to="teleportTo" :disabled="teleportDisabledBoolean">
```

VividLemon (changes requested):

PR #1210, opened 2023-06-23
feat(BToast): Simplify toast plugin

aceofwings:
# Describe the PR

Simplify toast plugin.  The plugin had a lot of misc code that were ideas and features that could possibly be developed further. None the less it was clutter and to start from the beginning makes more sense. 

Add documentation and client-side guard. I do not know if it is necessary though I remember the hassle of deploying docs in the olden days to due compilation errors with  vuepress. 

## PR checklist

- [x] Feature - `feat(...)`
- [ ] Other (please describe)


VividLemon:
> Add documentation and client-side guard. I do not know if it is necessary though I remember the hassle of deploying docs in the olden days to due compilation errors with vuepress.

It is not necessary

The simplification of the plugin is a welcome change. However, I would think about not writing tests for it. I'd like to develop this experience further. I would like to decouple toasts from b-container. It would be nice if the `useToast` function would dynamically mount/unmount the toaster container, but I'm not sure how I could achieve this.

Hypothetically, if you could render a VNode in a composable, this would fix our issues:

```ts
export default () => {
  const toasts = []
  const show = () => {
    toasts.push(h(BToast))
  }
toasts.forEach((el) => document.body.appendChild(el))
}
```
However, I don't know if this is possible.

An alternative solution could be to use provide/inject. Where
```ts
export default () => {
  const parent = inject(BToasterPluginKey)
  const show = () => {
    parent.addToast()
  }
  return {show}
}
```
Then have some component render that. Similar to BContainer rendering it, but not BContainer... A plugin would be preferred as opposed to a real component.

Vuetify solves this by using real components -- VApp. But I'd like to avoid a component requirement for this. If possible, I'd prefer a plugin option that you can call once. You can mount elements in plugins. For example, the BTooltip directive does this. So if we combined rendering BToaster through a plugin, then had BToaster using a provide/inject system, we could accomplish a more streamlined system. 

Though, in this situation, BToast component would need some modifications as if someone did https://bootstrap-vue.org/docs/components/toast and created a toast component, it wouldn't really be usable. I don't know how one would make this usable... But perhaps the simplicity of just calling a method would mean that the exact usage from bootstrap-vue would justify it being deprecated. 

Bootstrap-vue had it "easy", as they relied on portal-vue, as opposed to Vue 3's teleport. In portal-vue, you can check for a targets existence. In their situation, they can check for the existence of BToaster, if it doesn't exist, they can make it exist. If it does exist, they can just teleport to it. This is not something you can do in Vue 3. First off, Vue 3's teleport lacks the ability to wait for a component to exist. The second issue, is you can't functionally determine if the target exists. So, we need to get creative, and likely will need deprecation & removals.

VividLemon:
>  It would be nice if the useToast function would dynamically mount/unmount the toaster container, but I'm not sure how I could achieve this.

"Maybe" we could do something like
```ts
const toaster = createSharedComposable(() => {
  const toaster = createApp(BToaster) // append to body
})

const useToast = () => {
  toaster()
  // Create and append children toasts
}
```
But I don't know if this would work.

aceofwings:
> > It would be nice if the useToast function would dynamically mount/unmount the toaster container, but I'm not sure how I could achieve this.
> 
> "Maybe" we could do something like
> 
> ```ts
> const toaster = createSharedComposable(() => {
>   const toaster = createApp(BToaster) // append to body
> })
> 
> const useToast = () => {
>   toaster()
>   // Create and append children toasts
> }
> ```
> 
> But I don't know if this would work.

I see where you are going with this and I have reviewed the tool tip code. The toast code can be readily written to mount another app to a appended container. We can readily pass down the simplified code (ToastController) to be shared between the instances (I think).  Just to confirm it is a totally acceptable practice due to library limitation to spawn multiple app instances instead of using  real components?

VividLemon:
>  Just to confirm it is a totally acceptable practice due to library limitation to spawn multiple app instances instead of using real components? 

idfk. But I say "just do it".

VividLemon:
Would you prefer me to merge this now, or are you planning on trying to implement what has been discussed and would prefer me to wait?

VividLemon:
I apologize, but there was an issue with the scss code for BToast. In Bootstrap v5.3, they changed some things that affected the 'alpha' function for BToast. As a temporary measure, I have removed the scss from BToast, as ultimately the component didn't really work to begin with. I somewhat forgot about this PR that changed some of how it functioned, but even if I did not forget about it, I would likely have still made the temporary change.

If you would like, it would be a large help to rebuild the scss (perhaps some of it is unneeded and can be removed... Perhaps it all could have been removed and there is no issue) But some additional information can be found here https://github.com/twbs/bootstrap/issues/38687 as it is not a bug to Bootstrap and will require some fixing on our end

aceofwings:
No worries , I already did a very small adjustment to get things to compile properly. I will overhaul it. The scss was orignally written as bootstrap v4 did not support toasts. Since the release of 5.3 they have released subtle color pallets. This will give us the same lightly color background. let me see what i can do

aceofwings:
So accidently left out the minor changes (the small adjustment I mentioned earlier) removing the compilation error you were talking about above.

VividLemon:
https://github.com/bootstrap-vue-next/bootstrap-vue-next/pull/1210#issuecomment-1605800203

aceofwings:
I looked into using bg-opacity and bg-${variant} classes and styling assignments. Sadly, the default coloring looks a little to harsh and is not as elegant as I hoped. None the less the css I pushed 2 weeks ago covers the compilation error you referenced above. I think the best thing to do is keep bootstrap vue's original toast color pallet and in the near future address opacity and solid settings in the toast component.

VividLemon:
> I looked into using bg-opacity and bg-${variant} classes and styling assignments. Sadly, the default coloring looks a little to harsh and is not as elegant as I hoped. None the less the css I pushed 2 weeks ago covers the compilation error you referenced above. I think the best thing to do is keep bootstrap vue's original toast color pallet and in the near future address opacity and solid settings in the toast component.

Why is any of the custom css required? What parts of the component are not already handled by Bootstrap?

aceofwings:
Handling toast varieties is done by bg-color and bg-opacities. However, the contrast for doing them is poor. I suggest we maintain color pallet as implemented in the bootstrap vue and later when I implement the solid prop, remove opacity styling and have it computed in the component. Attached are pictures of bootstraps' color pallet vs bootstrap vues. 
 
toast 1
toast 2
toast 3

VividLemon:
> Handling toast varieties is done by bg-color and bg-opacities. However, the contrast for doing them is poor. I suggest we maintain color pallet as implemented in the bootstrap vue and later when I implement the solid prop, remove opacity styling and have it computed in the component. Attached are pictures of bootstraps' color pallet vs bootstrap vues.
> toast 1 toast 2 toast 3

You will also need to keep in mind that Bootstrap v5.3 implements color modes. So the existing code will need to change to compensate for this -- I don't mind about custom css to make up for some shortcomings of bootstrap, but color modes complicates things.

sceee:
Shouldn't this just reflect the default styling of bootstrap?
If someones wants to customize it, it is probably best just use the [color schemes starting with Bootstrap 5.3](https://getbootstrap.com/docs/5.3/components/toasts/#color-schemes).
This uses `text-*` and `text-bg-*` classes [for styling](https://getbootstrap.com/docs/5.3/helpers/color-background/).

So if it would be possible to set these classes as prop (or via a `variant` prop for both (text and bg) which then translates into those classes), it would reflect the default bootstrap 5.3+ behavior.

Someone could probably still use [CSS Variables to override](https://getbootstrap.com/docs/5.3/components/toasts/#variables) some general styling.

apps/docs/.vitepress/theme/index.ts
```diff
@@ -14,5 +14,9 @@ export default {
       discordUrl: 'https://discord.gg/j2Mtcny',
       opencollectiveUrl: 'https://opencollective.com/bootstrap-vue-next',
     })
+
+    if (!import.meta.env.SSR) {
```
VividLemon, line 18: Remove this. -- The SSR portion

apps/docs/src/docs/components/toast.md
```diff
@@ -24,12 +24,147 @@ To encourage extensible and predictable toasts, we recommend providing a header
 
 Toasts are slightly translucent, too, so they blend over whatever they might appear over. For browsers that support the `backdrop-filter` CSS property, they also attempt to blur the elements under the toast.
 
-## Docs to be completed
+## Plugin Usage
+
+The toast plugin allows for global access and handling of toasts for a single `<b-container>` instance.
```
VividLemon, line 29: `<BContainer>` > `<b-container>` use PascalCase

apps/docs/src/docs/components/toast.md
```diff
@@ -24,12 +24,147 @@ To encourage extensible and predictable toasts, we recommend providing a header
...
+
+To use the plugin make sure to include it in your app.
+
+```js
```
VividLemon, line 33:
```suggestion
```ts
```
(Though it technically makes no difference

apps/docs/src/docs/components/toast.md
```diff
@@ -24,12 +24,147 @@ To encourage extensible and predictable toasts, we recommend providing a header
...
+import {BToastPlugin} from 'bootstrap-vue-next'
+app.use(BToastPlugin)
+```
+
+Then designate a `<b-container>` component using the toast prop.
```
VividLemon, line 38: Review above comment

apps/docs/src/docs/components/toast.md
```diff
@@ -24,12 +24,147 @@ To encourage extensible and predictable toasts, we recommend providing a header
...
+
+```vue-html
+<b-container :toast="true" fluid="sm" position="position-fixed"></b-container>
```
VividLemon, line 41: Review above comment

Also, use <BContainer /> syntax over <BContainer></BContainer>

apps/docs/src/docs/components/toast.md
```diff
@@ -24,12 +24,147 @@ To encourage extensible and predictable toasts, we recommend providing a header
...
+```
+
+We then use the toast composable to grab our root toast model for displaying and handling toasts.
+
+```js
```
VividLemon, line 46:
```suggestion
```ts
```

apps/docs/src/docs/components/toast.md
```diff
@@ -24,12 +24,147 @@ To encourage extensible and predictable toasts, we recommend providing a header
...
+//setup function
```
VividLemon, line 47:
```suggestion
```
This is a given

apps/docs/src/docs/components/toast.md
```diff
@@ -24,12 +24,147 @@ To encourage extensible and predictable toasts, we recommend providing a header
...
+import {useToast} from 'bootstrap-vue-next'
+let toast = useToast()
```
VividLemon, line 49:
```suggestion

const toast = useToast()
```

apps/docs/src/docs/components/toast.md
```diff
@@ -24,12 +24,147 @@ To encourage extensible and predictable toasts, we recommend providing a header
...
+  - `body` : string or an array of `VNodes`
+- `options`: an optional options object for changing position and delay hidden
+  - `autoHide` (optional) : true/false - by default is true
+  - `noCloseButton` (optional) : true/false - by default is false
+  - `pos`: 'top-left', 'top-center', 'top-right', 'middle-left', 'middle-center', 'middle-right', 'bottom-left', 'bottom-center', 'bottom-right'
+  - `delay`: number
+  - `variant` : string ColorVariant
+
+<HighlightCard>
+<b-button class="" @click="show1">Show Toast</b-button>
```
VividLemon, line 66: Pascal case. Also, fix your indentation

apps/docs/src/docs/components/toast.md
```diff
@@ -24,12 +24,147 @@ To encourage extensible and predictable toasts, we recommend providing a header
...
+<b-button class="ms-2" @click="show2">Show Toast with Body</b-button>
+  <template #html>
+
+```vue
```
VividLemon, line 70: Review other similar documentation examples. There is a difference between vue-html and vue. One is for pure template stuff, the other is for both template and scripts.

Put a line break between the two.

Code blocks indentations restart (if that makes sense)

apps/docs/src/docs/components/toast.md
```diff
@@ -24,12 +24,147 @@ To encourage extensible and predictable toasts, we recommend providing a header
...
+<b-button class="" @click="show1">Show Toast</b-button>
+<b-button class="ms-2" @click="show2">Show Toast with Body</b-button>
+<script setup lang="ts">
+let show1 = () => {
```
VividLemon, line 74: Use const

apps/docs/src/docs/components/toast.md
```diff
@@ -24,12 +24,147 @@ To encourage extensible and predictable toasts, we recommend providing a header
...
+  toast.show({body: 'example title'}, {pos: 'bottom-right'})
+}
+let show2 = () => {
```
VividLemon, line 77: Use const

apps/docs/src/docs/components/toast.md
```diff
@@ -24,12 +24,147 @@ To encourage extensible and predictable toasts, we recommend providing a header
...
+
+<HighlightCard>
+  <b-button class="" variant="danger" @click="variantshow1">Show Danger</b-button>
+  <b-button class="ms-2" variant="info" @click="variantshow2">Show Info</b-button>
+  <b-button class="ms-2" variant="warning" @click="variantshow3">Show Warning</b-button>
+  <b-button class="ms-2" variant="success" @click="variantshow4">Show Sucess</b-button>
+  <template #html>
+
+```vue
+<b-button class="" variant="danger" @click="variantshow1">Show Danger</b-button>
```
VividLemon, line 101: Pascal case

VividLemon: Wrap in` <template>`

apps/docs/src/docs/components/toast.md
```diff
@@ -24,12 +24,147 @@ To encourage extensible and predictable toasts, we recommend providing a header
...
+<b-button class="ms-2" variant="info" @click="variantshow2">Show Info</b-button>
+<b-button class="ms-2" variant="warning" @click="variantshow3">Show Warning</b-button>
+<b-button class="ms-2" variant="success" @click="variantshow4">Show Sucess</b-button>
```
VividLemon, line 104:
```suggestion
<b-button class="ms-2" variant="success" @click="variantshow4">Show Sucess</b-button>

```

apps/docs/src/docs/components/toast.md
```diff
@@ -24,12 +24,147 @@ To encourage extensible and predictable toasts, we recommend providing a header
...
+<script setup lang="ts">
+import {useToast} from 'bootstrap-vue-next'
+let toast = useToast()
```
VividLemon, line 107: const
(this goes for everything below)

packages/bootstrap-vue-next/src/components/BToast/plugin.ts
```diff
@@ -145,86 +137,45 @@ export class ToastController {
...
-
-export function useToast(): ToastInstance | undefined
-export function useToast(vm: {id: symbol}, key?: symbol): ToastInstance | undefined
-export function useToast(
-  vm: {container: Ref<ComponentPublicInstance>; root: boolean},
-  key?: symbol
-): ToastInstance | undefined
-
-export function useToast(vm?: any, key: symbol = injectkey): ToastInstance | undefined {
+export function useToast(): ToastInstance | undefined {
```
VividLemon, line 166: Use arrow function syntax over functions (assuming the function does not access `this`)

VividLemon (changes requested):

apps/docs/.vitepress/theme/index.ts
aceofwings: so if I remove this then the docs do not work with the toast implementation.

aceofwings: nvm

aceofwings:
I do recommend for future we revisit the suggestions above. Perhaps come to terms with have 2 implementations. One driven by B-Container as seen in the Veutify eco system. We are trying to create a bootstrap eco system so I would not shy away from real component solution. As well as maintain a separate app for spawning toasts. As mentioned in another issue multiple apps pollute the debugging tool Vue has to offer.

VividLemon:
@aceofwings Why did you close this?

PR #1212, opened 2023-06-24
Added sidebar right for 'on this page'

StirStudios:
Moved 'on this page' to an off canvas for better usability.

Also removed the toggle to use v-b-toggle instead.

### DIFF
--- a/apps/docs/.vitepress/theme/Layout.vue
+++ b/apps/docs/.vitepress/theme/Layout.vue
@@ -74,7 +74,7 @@
       </client-only>
     </b-nav>
     <b-navbar-toggle v-b-toggle.otp-menu class="otp-menu-toggle">
-      On the page <ChevronRight :variant="null" aria-label="On this page menu" />
+      On the page <ChevronRight :variant="null" aria-hidden />
     </b-navbar-toggle>
   </b-navbar>
   <b-container fluid class="container-lg mt-3 my-md-4 bd-layout">
--- a/apps/docs/.vitepress/theme/Layout.vue
+++ b/apps/docs/.vitepress/theme/Layout.vue
@@ -1,6 +1,6 @@
 <template>
   <b-navbar variant="primary" sticky="top" toggleable="lg">
-    <b-navbar-toggle @click="toggler" />
+    <b-navbar-toggle v-b-toggle.sidebar-menu />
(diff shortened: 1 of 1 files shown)


apps/docs/.vitepress/theme/Layout.vue
```diff
@@ -73,11 +73,15 @@
         </b-dropdown>
       </client-only>
     </b-nav>
+    <b-navbar-toggle v-b-toggle.otp-menu class="otp-menu-toggle">
+      On the page <ChevronRight :variant="null" aria-label="On this page menu" />
```
VividLemon, line 77: A screen reader will announce this as 
`On the page On this page menu`
Which doesn't quite make sense.

You may prefer to simply use aria-hidden on this icon, as I think the text is sufficient.

VividLemon (changes requested):

apps/docs/.vitepress/theme/Layout.vue
StirStudios: Done, good point, silly me!

apps/docs/.vitepress/theme/Layout.vue
```diff
@@ -73,11 +73,15 @@
         </b-dropdown>
       </client-only>
     </b-nav>
+    <b-navbar-toggle v-b-toggle.otp-menu class="otp-menu-toggle">
+      On the page <ChevronRight :variant="null" aria-hidden />
```
VividLemon, line 77: You put :variant="null" on icons every now and again, but I don't think they have any affect. 

If you're trying to replicate what I did in https://github.com/bootstrap-vue-next/bootstrap-vue-next/blob/080b4f8e02eace6b593d374916c2e976bdd85774/apps/docs/.vitepress/theme/Layout.vue#L38 , the reason for the variant null on that is because that is a button, and since it's an icon, we don't really want the background colors for the button.

Notice that in that case, it's on the button, and not the actual icon, so there is a difference.

VividLemon (changes requested):

apps/docs/.vitepress/theme/Layout.vue
StirStudios: Yup, silly me again. I also adjusted the case for consistency as well.

StirStudios:
I think this PR should give us yet another good boost for the docs usability.

PR #1231, opened 2023-06-30
feat: defineExpose reset on BFormFile input

sceee:
# Describe the PR

Add `reset` method to the `BFormFile` component similar to [bootstrap-vue's reset method](https://bootstrap-vue.org/docs/components/form-file#clearing-the-file-selection)
* Allows to reset the input on the BFormFile component through a template ref
* Add documentation for previously added focus, blur exposed methods

## PR checklist

- [x] Feature - `feat(...)`
- [ ] Other (please describe)


VividLemon:
The modelValue will still contain a value despite setting the input value to an empty string

sceee:
> The modelValue will still contain a value despite setting the input value to an empty string

@VividLemon I'm unsure whether you mean that's a bug in the solution now merged that still needs to be adressed?
Do you mean we should additionally emit a `change` event after the reset?

VividLemon:
I'm saying that the solution you provided only changes the value of the input field. It does not affect the modelValue. This is a better solution: 
image
I have it in my own branch. You don't need to do anything

Issue #1261, opened 2023-07-08
Library discovery difficulty


devhus:
I've noticed that the library is not very popular especially after merging with BV2 and being independent again. This issue is to discuss a few matters about the library info that should improve the package's SEO on NPM and other aspects.

_**All the given solutions are based on my point of view and might be wrong but the problems do exist. I will publish a PR that includes a few updates to improve the library discovery and this issue is just to clarify the PR  porpuse.**_

### The problem
As most of this library users know the library name used to be "bootstrap-vue-3" before merging with BV2 and jumping to bootstrap-vue-next repo. 
The problem with the current name "bootstrap-vue-next" is that it's not a very clear name to junior devs or the search bots as you can see in the image below you can't find the library when searching for something like "Vue 3 Bootstrap" 
![image](https://github.com/bootstrap-vue/bootstrap-vue/assets/58523735/9b811060-016a-4ca5-9898-bec3e4a4194f)

- We can see BV2 is getting the lead but is understandable since their website has a higher ranking..etc.
- The next search result is the NPM URL of the old package, which doesn't have any direct link to the new NPM package. The package is just deprecated 
- We can see the GitHub URL of the package but that not going to appear to most devs, it appears to me because of how many times visited this repo

### The solution
**Change the library name to "Bootstrap Vue 3".**
The library name (at least the display name) should include "vue 3" words, That will impact the search engine bots to be able to give better results when searching for direct libraries for vue 3. Keep in mind that the word "next" doesn't specify that the library is for vue 3 or 2, "next" can be used as an upgraded version of the library but not the main name or display_name.

### The problem
As we can all notice that all of the library read-me, description, and keywords are missing a lot of directions and clarifications about the library to be well noticed by both devs and search bots.

### The solution
Simply we will need to update the library's read-me, description, and keywords to include more directions and clarifications about the library such as using the word "vue 3", and "nuxt 3" more.

## Documentations
Nothing to say about this part since the docs are getting updated slowly to include all of the library aspects. We will just need to spend more time improving the documentation and the content's SEO

### The problem
I have no idea about what is going on behind the scenes but it's a very bad idea to keep changing the main repository and npm package of the library that will impact the library's popularity.

### The solution
Up to the library's owners to stop these kind of updates

VividLemon:
This should help https://github.com/bootstrap-vue-next/bootstrap-vue-next/pull/1263

VividLemon:
[...]

Issue #1253, opened 2023-07-06
Sponsor a bit development on the BTable-component


JanBolmeson:
Hi!

We love this project and used the old bootstrap-vue extensively. Instead of putting in the time myself in choosing another component or doing the work in our own project, would it be possible for me to sponsor development for some BTable-functionality by some of the core-members? 

Thanks,
Jan

devhus:
I am coming back soon to continue developing the BTable component, hopefully.

VividLemon:
Any sponsorships should be done solely through the open collective or an individuals github sponsor links. I'm not going to tell individuals on how they can accept payments, but to protect all parties involved, it would be best to use official channels.

JanBolmeson:
I am sorry, didn't mean to break any rules. Just tell me how to do it correctly and I will comply. Sorry. 🙈

VividLemon:
There is no written rule. The issue is, hypothetically, if you were to convince someone to send you confidential financial details for the purposes of extorting or scamming them. This is what I seek to prevent by this statement.

devhus:
@VividLemon Just to make it clear, I am not planning to get any money from anyone my collaboration will be for free.
@JanBolmeson I am just making a statement that i am gonna be back soon, so you don't really need to pay anyone.
However, if you need a specific feature/fix just submit an issue with it.

VividLemon:
This ain't bootstrap-vue where people donate and the money just sits in an account for years. Anyone that has contributed and feel like their work has made an impact is free to submit an expense.

devhus:
@JanBolmeson The feature has been added in this https://github.com/bootstrap-vue-next/bootstrap-vue-next/pull/1262 PR

VividLemon:
@devhus it would be really neat if the table component could utilize Vue 3.3s generic components. I could see a lot of things on that component utilizing generics

devhus:
@VividLemon i am not up to date with vues updates, would you like to send me docs or article of the component generic you mentioned?

VividLemon:
> @VividLemon i am not up to date with vues updates, would you like to send me docs or article of the component generic you mentioned?

https://blog.vuejs.org/posts/vue-3-3#generic-components
https://vuejs.org/api/sfc-script-setup.html#generics

VividLemon:
https://github.com/bootstrap-vue-next/bootstrap-vue-next/pull/1533 considering this closed in the meantime

Issue #1251, opened 2023-07-05
Row select support in tables is missing features


stevemandl:
The bootstrap-vue feature for row select support is partly documented here: https://bootstrap-vue.org/docs/components/table#row-select-support

- When a row is selected, the 'selected' field should display a check iff the row is selected. Currently the check is always displayed, and only the class of the check changes (text-primary is added) when the row is selected.

- The checkbox field should only display if the fields array contains a 'selected' field, and it should display in the column location corresponding to the position in the fields array. Currently the select cell always displays in the first column, unless the select-head prop is false in which case it does not display.

devhus:
In this https://github.com/bootstrap-vue-next/bootstrap-vue-next/pull/1262 PR, You will be able to apply your own selection styling using the mentioned css selectors.
About the "selected" field, I didn't like the way BV2 doing it and how its limiting you from using "selected" named field. Keep in mind that it is very rare to find someone who wants the selection box in the center of the table and i don't think it worth the effort

stevemandl:
@devhus 
Thanks for the thoughtful reply.
I agree a checkbox in the center of the table is rare. I have a need to put the selection box at the end of the table, and tried to follow the BV2 docs to see if it would work. I ended up coding my own selection box in a column at the end, and if I need to keep my custom table so be it.
   Is there is an easy way in css to make the checkbox appear at the end of the selected row using the selection styling you mentioned? From what I saw it looks like we will be able to change the way the checkbox looks, or how the row is styled, but adding a checkbox to the last column seems tricky or impossible in css.

VividLemon:
So, based on my changes there are a couple of things to note. 

The major note is that the "row-selected" event is different than the BV version. Namely that it does not emit the array. 

The second major note is that the functionality for row selection has changed. Now instead you can (in addition to the other methods, exposed and otherwise) use a v-model. This allows for more flexible selection. 

The documentation on this is yet to be written, and its in an unreleased state. I'm going to keep "row-selected" event as emitting only its row, as I believe it makes more logical sense for a "row-selected" to emit only that "selected row", given all the new options you have for selection (simply `v-model:selectedItems` should give you the same result, with some minor modification [you won't need the event])

In the meantime, I'm going to close this, docs to come

Issue #1061, opened 2023-04-22
B-Table: Pre-Sorting


susanne99:
in bootstrap vue I used the props: "sortby" and "sortdesc" to presort the table. 
All columns had "sortable: false" set. 
Although the documentation says: "Setting sort-by to a column that is not defined in the fields as sortable will result in the table not being sorted." it worked anyway. 
It no longer works in Bootstrap vue next. 
you really have to set the sorting column to "sortable: true". 
Unfortunately, you can no longer pre-sort the table. That would be great though, because then I can pre-sort the table without making a column sortable. 
Now I have to pre-sort the table myself. 

**Suggested solution: 
the two props "sortby" and "sortdesc" should be used for pre-sorting without making columns sortable. 
This should be relatively quick and easy to install. thanks in advance**

VividLemon:
Given that this is a _bug_ according to the documentation, I am obligated to fix it. You will need to sort it before it reaches BTable. Will be fixed in 0.15

PR #1262, opened 2023-07-08
Package SEO improves, BTable updates, and Docs updates

devhus:
Package SEO improves, Please read this issue to understand why the package info needs to be updated.
https://github.com/bootstrap-vue-next/bootstrap-vue-next/issues/1261

BTable updates:
- feat(BTable): updated table field formatter to a reusable type
- feat(BTable): corrected field object formatter type
- feat(BTable): corrected item rendering logic, allowing easier updates to the renderItem function which can now include more logic for rendering besides formatting
- BREAKING CHANGE(BTable): Updated the items processing logic, Now items are fully reactive and can be updated directly through the item's cell slot.
- feat(BTable): removed items cloning when emitting to the filtered handler to deliver reactive items
- feat(BTable): sortBy & sortDesc now can be used to pre-sort the table  items without making any fields sortable
- feat(BTable): the selection box is now can be modified through CSS uisng the ``.b-table-selection-icon`` and ``.selected`` selectors
- feat(BTable): added tbodyTrClass prop to update the tbody > rows styles based on function results
- feat(BTable): the sort-by prop now has a shell value to be more flexible and can be used as a v-model, read-only prop, or don't be passed at all

Note that items cloning is partly removed and that should improve the component's performance when displaying a large number of rows that need to be sorted or filtered.

This PR should solve these issues:
- https://github.com/bootstrap-vue-next/bootstrap-vue-next/issues/1261
- https://github.com/bootstrap-vue-next/bootstrap-vue-next/issues/958
- https://github.com/bootstrap-vue-next/bootstrap-vue-next/issues/1061
- https://github.com/bootstrap-vue-next/bootstrap-vue-next/issues/1251 (partly)
- https://github.com/bootstrap-vue-next/bootstrap-vue-next/issues/1253
- https://github.com/bootstrap-vue-next/bootstrap-vue-next/issues/1269
- https://github.com/bootstrap-vue-next/bootstrap-vue-next/issues/1270
- https://github.com/bootstrap-vue-next/bootstrap-vue-next/issues/995


packages/bootstrap-vue-next/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "bootstrap-vue-next",
-  "displayName": "BootstrapVueNext",
-  "description": "Early (but lovely) implementation of Vue 3, Bootstrap 5 and Typescript",
+  "displayName": "BootstrapVue3",
```
VividLemon, line 3: Do not change the display name. It will stay BootstrapVueNext

packages/bootstrap-vue-next/package.json
```diff
@@ -1,7 +1,7 @@
...
+  "description": "Bootstrap Vue 3 library is an early and lovely for Vue 3 & Nuxt 3 based on Bootstrap 5 and Typescript.",
```
VividLemon, line 4: Modification of the description is "fine", but review above.

packages/bootstrap-vue-next/src/components/BTable/table-items.ts
```diff
@@ -0,0 +1,171 @@
+import {computed, ref, type Ref} from 'vue'
```
VividLemon, line 1: This file name should be camelCase

packages/bootstrap-vue-next/src/components/BTable/table-items.ts
```diff
@@ -0,0 +1,171 @@
...
+import type {BTableProps} from './table'
+import type {BTableSortCompare, TableField, TableFieldObject, TableItem} from 'src/types'
```
VividLemon, line 3: You cannot import from src/ use relative file paths

packages/bootstrap-vue-next/src/components/BTable/table-items.ts
```diff
@@ -0,0 +1,171 @@
...
+import type {TableFieldObjectFormatter} from 'src/types/TableFieldObject'
+
+export function useTableItems(
```
VividLemon, line 6: arrow functions  > functions

VividLemon, line 6: I'd prefer this code to simply be in the actual BTable file. It may look bloated, but I think it outweighs the drawback of code being hidden from first view. I similarly don't like how BFormInput has a special composable, but in that case, it is used in multiple places. This is not.

packages/bootstrap-vue-next/src/components/BTable/table-items.ts
```diff
@@ -0,0 +1,171 @@
...
+  tableProps: BTableProps,
```
VividLemon, line 7: In addition to the above, I'd prefer BTableProps to not be declared in a separate file. (Review below)

packages/bootstrap-vue-next/src/components/BTable/table-items.ts
```diff
@@ -0,0 +1,171 @@
...
+  flags: Record<string, Ref<boolean>>,
```
VividLemon, line 8: It may be better to use `Record<string, MaybeRefOrGetter<boolean>>`, but you will need to handle it differently... However this may be a non issue

packages/bootstrap-vue-next/src/components/BTable/table.d.ts
```diff
@@ -0,0 +1,51 @@
...
+  Booleanish,
+  Breakpoint,
+  BTableProvider,
+  BTableSortCompare,
+  ColorVariant,
+  VerticalAlign,
+} from '../../types'
+
+export interface BTableProps {
+  align?: VerticalAlign
```
VividLemon, line 11: As mentioned above, I'd prefer the composable to be moved back into the BTable file alone. Along with this, this type can be moved back into the position it was before. I am preferring to not have different declaration files anymore.

packages/bootstrap-vue-next/src/components/BTable/table.d.ts
```diff
@@ -0,0 +1,51 @@
+import type {
```
VividLemon, line 1: In addition to this, even though this file should be removed, you are not supposed to write .d.ts files. Review https://www.youtube.com/watch?v=zu-EgnbmcLY

VividLemon (changes requested):

packages/bootstrap-vue-next/src/components/BTable/BTable.vue
```diff
@@ -362,8 +361,7 @@ filteredHandler.value = async (items) => {
     await callItemsProvider()
     return
   }
-  const clone = await cloneDeepAsync(items)
```
VividLemon: I agree, the clone deep didn't serve much purpose. You can likely remove the cloneDeep function and the subsequent cloneDeepAsync. But you'd want to check first.

packages/bootstrap-vue-next/src/components/BTable/table-items.ts
devhus: Yeah, long time not working on the library i forgot that part xD

packages/bootstrap-vue-next/src/components/BTable/table-items.ts
devhus: I wanted to have a clean, more readable code. Having the logic of the items handling as composable would be more maintainable if we need to change anything related to the items processing but not to the interface. Think about it and let me know if you still wanna merge interface and items logic into one large file.

VividLemon: In my opinion, moving one monstrous ugly code from one place to another, doesn't fix the problem of it being monstrous and ugly. Not to mention, I'm not a fan of the idea of having the parameter be "the entire props object", things get messy, reactivity wise as it doesn't quite fit the expected flow (at least in my mind)

devhus: Hmmm, Idk it does feel better to have the items processing code in a standalone file. Keep in mind that most of the table-items.ts code was already existing in itemHelper.ts file which is also a standalone file.

packages/bootstrap-vue-next/package.json
devhus: Np

packages/bootstrap-vue-next/src/components/BTable/table-items.ts
devhus: It will not be an issue since all of the given flags are Ref-like anyway, if there are getters added to the flags we can change the types for sure. But for now, there is no need to add any extra validation for the type.

packages/bootstrap-vue-next/src/components/BTable/table.d.ts
devhus: Good to know, thanks for the tip

packages/bootstrap-vue-next/src/components/BTable/table.d.ts
devhus: The logic used within the composable has been always outside the BTable file. So do you want to remove the idea of having external files for logic processing, or you just don't like to have the props declaration outside the .vue file?

packages/bootstrap-vue-next/src/components/BTable/BTable.vue
devhus: I have checked with a few manual tests, Everything seems to be working as it was except now items are reactive everywhere.

packages/bootstrap-vue-next/src/components/BTable/table-items.ts
devhus: So, this file is simply a reactive alternative for itemHelper.ts file.  https://github.com/bootstrap-vue-next/bootstrap-vue-next/pull/1262/commits/c6e004fc41b5678730db0b18eaf7b5e55880dc4c#diff-ef682efae412cee917e8dc66838252a8c9d3189da77868f1aeab58e8b078f4df

If the problem with moving the props type outside BTable.vue file let me know and i will handle it.

devhus: I've submitted a few updates to the PR, let me know if this is enough or not to continue working on other issues

packages/bootstrap-vue-next/src/components/BTable/BTable.vue
VividLemon: Typically the setter getters and proxies are destroyed during the serialization process, which breaks reactivity.

packages/bootstrap-vue-next/src/components/BTable/table.d.ts
VividLemon: I suppose I'm not opposed to placing this file in the composables folder. However, if its moved, it should likely receive a more specific name. The explicit use of the entire props object is an issue for me, I find it very messy. If possible, perhaps it would be better to instead split them into many smaller (and more readable) composables. (Assuming that's doable and the current composable can be 'decomposed')

VividLemon:
You're going to need to either clean up some of these props or define what they're actually used for. A few weeks ago, I had done some cleanup as your code mimicked a "passive vmodel". If you define what some of the props you made during your last dive into the table, perhaps we can discuss if they're actually needed, or better methods that can be done

VividLemon:
Also you should review https://github.com/orgs/bootstrap-vue-next/projects/1/views/1?pane=issue&itemId=29332903 as that is the direction that I'd like to push BTable to be in v1

devhus:
> You're going to need to either clean up some of these props or define what they're actually used for. A few weeks ago, I had done some cleanup as your code mimicked a "passive vmodel". If you define what some of the props you made during your last dive into the table, perhaps we can discuss if they're actually needed, or better methods that can be done.
> 
> In this instance, at first glance, I'm curious about the effect of sort internal.

Almost all props have a very good use, i don't think there is any props that can be removed tbh.
For the ``sort-internal`` prop it is useful when you are sorting the table from a paginated server-request.
when ``sort-internal`` is true, the table will sort the displayed items 
when ``sort-internal`` is false, the table will not sort the displayed items but instead will only emit sort props to the given ``sorted`` event so you can request an updated item list from your server and update the ``:items`` ref based on your response.
Think about ``sort-internal`` = false as a simpler alternative to the ``provider`` prop, it's quite useful and i am using it in 2 production projects.

Anyway, I was planning to document the props, emits, and most of the BTable code so it can be understood better. Maybe i will make the component's Doc page when i get some more free time.

devhus:
> Also you should review https://github.com/orgs/bootstrap-vue-next/projects/1/views/1?pane=issue&itemId=29332903 as that is the direction that I'd like to push BTable to be in v1

Should be easy to do but not in this PR because it is already 5 AM in my country. I will consider these changes with the next PR with the rest of the bug fixes.
For now, this PR is wrapped and will not have any more updates, But feel free to tell me if any of the previous commits need to be updated to accept the PR.

devhus:
@VividLemon all tests now are passed, goodnight

devhus:
@VividLemon BTableLite is created and, BTable is now a wrapper for BTableLite.

VividLemon:
https://github.com/bootstrap-vue-next/bootstrap-vue-next/pull/1262/commits/480ac8b09e56a27aa360aca4b1909d09a101a109

Review https://vueuse.org/core/useVModel/ , specifically, the type definitions section. To clarify, the "passive" option, by using a "passive" vmodel, you aren't required to pass in a prop in order to have it work. You can modify the value like it is bound, and use it like it is bound, but not require it. For instance, in your old code you used a variable called "internalBusyFlag". This value mimiced the idea of a passive vmodel, in that it is used as if a value is bound, and changes the behavior internally as if exists.

So, long story short, use a passive vmodel. 

However, this leads to a bigger question on what the behavior of sorting is if you don't specify a sortBy

devhus:
@VividLemon very good to know.

when using a shell value or the ``passive`` option on useVModel they will behave the same way.
Simply the sorting will still work in all causes but the given ``ref`` will not updated unless used with ``v-model:sort-by`` and in the video below you can see how the table behaves when using ``v-model:sort-by``, `:sort-by`, and removing the prop.

https://www.loom.com/share/6f22865acfc6480ba04d738863362163?sid=c6a0de79-f683-440b-a588-61bc10a83fca

VividLemon:
> @VividLemon very good to know, updated the code to use `passive` mode.
> 
> when using a shell value or the `passive` option on useVModel they will behave the same way. Simply the sorting will still work in all causes but the given `ref` will not updated unless used with `v-model:sort-by` and in the video below you can see how the table behaves when using `v-model:sort-by`, `:sort-by`, and removing the prop.
> 
> https://www.loom.com/share/6f22865acfc6480ba04d738863362163?sid=c6a0de79-f683-440b-a588-61bc10a83fca

Passive behavior is something that I didn't know about either, but comes in great handy when working with libraries. In addition to this, when you had initially developed the internalBusyFlag and other improvements, we did not use vueuse and couldn't take advantage of this feature.

devhus:
@VividLemon btw these are the features that currently exist in BTableLite:
- basic controlling like (foot, sticky, stacked.. etc)
- selection
- busy
- row details
- pagination

And for BTable it includes all the features of BTable ofc with extra advanced features:
- item custom providing
- sorting
- filtering,

Let me know if there are any features that need to be moved from BTableLite to BTable so i can wrap up this PR and start working on the tables docs.

VividLemon:
> @VividLemon btw these are the features that currently exist in BTableLite:
> 
>     * basic controlling like (foot, sticky, stacked.. etc)
> 
>     * selection
> 
>     * busy
> 
>     * row details
> 
>     * pagination
> 
> 
> And for BTable it includes all the features of BTableLite ofc with extra advanced features:
> 
>     * item custom providing
> 
>     * sorting
> 
>     * filtering,
> 
> 
> Let me know if there are any features that need to be moved from BTableLite to BTable so i can wrap up this PR and start working on the tables docs.

Pagination and busy should be moved out of BTableLite. It's debatable if selection is needed on BTableLite -- although I think it would be difficult to offload that issue onto the BTable wrapper component. I'm not sure what row details does but it seems fine to include it (I believe it's in the bootstrap-vue component so that's good)

packages/bootstrap-vue-next/src/components/BTable/BTable.vue
```diff
@@ -319,198 +163,89 @@ const emit = defineEmits<{
...
-  emit('headClicked', fieldKey, field, event, isFooter)
-
+const onFieldHeadClick = (
+  fieldKey: LiteralUnion<string>,
+  field: TableField,
+  event: MouseEvent,
+  isFooter = false
+) => {
   handleFieldSorting(field)
+  emit('headClicked', fieldKey, field, event, isFooter)
```
VividLemon, line 233: This wasn't directly from this PR, but all events should be kebab-case, rather than camelCase. I'm surprised I didn't notice this before. It may be worth double checking if this issue exists anywhere else (of course, `update:` prefix & modelValue for models is an exception)

packages/bootstrap-vue-next/src/components/BTable/BTableLite.vue
```diff
@@ -0,0 +1,522 @@
...
+)
+
+const emit = defineEmits<{
+  'headClicked': [
+    key: TableFieldObject['key'],
+    field: TableField,
+    event: MouseEvent,
+    isFooter: boolean
+  ]
+  'rowClicked': [item: TableItem, index: number, event: MouseEvent]
```
VividLemon, line 268: As mentioned in a previous comment, events should be kebab case... This is likely v 0.10 worthy and is a breaking change

packages/bootstrap-vue-next/src/components/BTable/BTableLite.vue
```diff
@@ -0,0 +1,522 @@
+<template>
+  <BTableSimple v-bind="containerAttrs">
+    <thead>
+      <slot v-if="$slots['thead-top']" name="thead-top" />
+      <tr>
```
VividLemon, line 5: Under the hood, tr, th, thead, etc etc should all use the internal components - BTr, BTh, BThead, etc etc

This will likely modify how classes in the parent are derived... As I think the only thing the wrapper components are used for are for variant classes.

VividLemon (changes requested):
There's likely one more round of review to go, considering the size of the PR. But everything considered, thank you for the PR.

packages/bootstrap-vue-next/src/components/BTable/table-lite.spec.ts
```diff
@@ -0,0 +1,555 @@
+import {enableAutoUnmount, flushPromises, mount} from '@vue/test-utils'
+import {afterEach, describe, expect, it} from 'vitest'
+import {ref} from 'vue'
```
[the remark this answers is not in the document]
devhus: I don't remember, must been a mistake or testing something out xD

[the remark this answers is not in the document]
VividLemon: I think it may be an unused variable. So perhaps you may want to give a quick test lint

packages/bootstrap-vue-next/src/components/BTable/BTableLite.vue
devhus: I will leave this update to the next PR because it's 6 AM here and gotta go, And will be too busy to update this PR in the upcoming days. So let's wrap it up to let you be free to publish it.

devhus:
@VividLemon I did all that can be done for now, Feel free to publish the PR or ask for more reviews to be handled if my time allows it.
Goodnight.

VividLemon:
> @VividLemon I did all that can be done for now, Feel free to publish the PR or ask for more reviews to be handled if my time allows it. Goodnight.

👋 I'll review it tomorrow, and assuming its good, it will be published. I'll make an issue tomorrow about the event names so it can be tracked.

devhus:
@VividLemon check out the Docs layout updates.
It does feel more like Bootstrap V5 styling to me now, let me know what you think.

Before:
image

After:
image

VividLemon:
Navbar dark is deprecated

devhus:
@VividLemon will update it.

i've changed a few things:
- navbar from fluid-container to container
- removed borders of the content sidebars 
- adjusted a few white spaces
- add active state to the left sidebar 
- added active state to the navbar links as well
image
- fixed sidebar toggler color on mobile view, it's now working correctly in the dark and light mode unlike before
- fixed spaces of nav elements on mobile view 

Before
image

After
image

apps/docs/.vitepress/theme/Layout.vue
```diff
@@ -204,6 +204,7 @@ import {useData, useRoute, withBase} from 'vitepress'
 import {appInfoKey} from './keys'
 import {useMediaQuery} from '@vueuse/core'
 import TableOfContentsNav from '../../src/components/TableOfContentsNav.vue'
+import {vBColorMode} from 'bootstrap-vue-next'
```
VividLemon, line 207: You duplicated this import

devhus: It's not working without importing it.

apps/docs/.vitepress/theme/Layout.vue
```diff
@@ -1,79 +1,118 @@
...
-        <title>BootstrapVueNext</title>
-        <path
-          fill="currentColor"
-          d="M92 0h769c50 0 92 42 92 92v769c0 50-42 92-92 92H92c-50 0-92-42-92-92V92C0 42 42 0 92 0zm216 710c100 0 160-50 160-133 0-62-44-107-108-113v-3c48-8 86-52 86-102 0-71-55-117-140-117H111v468h197zM195 307h90c50 0 78 23 78 64 0 44-33 68-91 68h-77V307zm0 338V499h90c64 0 98 25 98 73s-33 73-94 73h-94zm503 65l163-468h-90L652 621h-2L531 242h-92l163 468h96z"
-        />
-      </svg>
-    </BNavbarBrand>
+  <BNavbar variant="primary" sticky="top" toggleable="lg" :container="true" v-b-color-mode="'dark'">
+    <div class="d-flex gap-2 align-items-center">
+      <BNavbarToggle v-b-toggle.sidebar-menu class="text-light" />
```
VividLemon, line 4: Are you sure you need text-light if the color mode of the parent is dark?

apps/docs/.vitepress/theme/Layout.vue
```diff
@@ -1,79 +1,118 @@
...
+    </div>
     <BCollapse is-nav>
       <BNavbarNav>
         <BNav>
-          <BNavItem :to="withBase('/docs')">Getting Started</BNavItem>
-          <BNavItem :to="withBase('/docs/icons')">Icons</BNavItem>
-          <BNavItem :to="withBase('/docs/types')">Types</BNavItem>
-          <BNavItem :to="withBase('/docs/reference')">Reference</BNavItem>
-          <BNavItem :to="withBase('/docs/migration-guide')">Migrate</BNavItem>
+          <BNavItem
```
VividLemon, line 30: It wasn't necessarily built this way, but you may prefer to create a variable in the script and use a v-for. Rather than duplicate these similar components a few times.

apps/docs/.vitepress/theme/Layout.vue
```diff
@@ -1,79 +1,118 @@
...
-          </BDropdownItem>
-        </BNavItemDropdown>
-      </ClientOnly>
-    </BNav>
-    <BNavbarToggle v-b-toggle.otp-menu class="otp-menu-toggle">
-      On the page <ChevronRight aria-hidden />
-    </BNavbarToggle>
+    <div class="d-flex gap-2 flex-wrap">
+      <BNav class="d-flex gap-4">
+        <BNavItem
```
VividLemon, line 70: This could probably also do the above comment

apps/docs/.vitepress/theme/Layout.vue
```diff
@@ -1,79 +1,118 @@
...
+              <component :is="currentIcon" :aria-label="`Toggle theme (${dark})`" />
+            </template>
+            <BDropdownItem v-for="el in options" :key="el" :active="dark === el" @click="set(el)">
+              <component :is="map[el]" /> {{ el }}
+            </BDropdownItem>
+          </BNavItemDropdown>
+        </ClientOnly>
+      </BNav>
+      <BNavbarToggle v-b-toggle.otp-menu class="otp-menu-toggle">
+        On the page
```
VividLemon, line 112:
```suggestion
        On this page
```
Is this correct?

apps/docs/.vitepress/theme/Layout.vue
VividLemon: Move this 
image

apps/docs/src/components/TableOfContentsNav.vue
```diff
@@ -1,50 +1,68 @@
 <template>
   <div class="bd-links-nav">
     <BListGroup v-if="!isLargeScreen">
-      <strong class="bd-links-heading"><GearIcon aria-hidden /> General</strong>
+      <strong class="bd-links-heading"> <GearIcon aria-hidden /> General </strong>
       <BListGroupItem>
-        <BLink :to="withBase('/docs')">Getting Started</BLink>
+        <BLink
```
VividLemon, line 6: This series of components could probably do the above suggestion

apps/docs/src/components/TableOfContentsNav.vue
```diff
@@ -151,4 +171,25 @@ const directivesComputedList = computed(() =>
     }))
     .sort((a, b) => a.name.localeCompare(b.name))
 )
+
+const groupComputedList = computed(() => [
+  {
+    label: 'Components',
+    uri: '/docs/components',
+    icon: () => IntersectIcon,
+    children: componentsComputedList.value,
```
VividLemon, line 180: I originally made these things computeds, but they don't seem to be reactive. You may be able to declare all of the computeds here as regular, non-reactive variables. If you do that, you likely don't need to make the icon property a function as there shouldn't be any reactivity overhead (perf)

packages/bootstrap-vue-next/src/components/BLink/BLink.vue
```diff
@@ -153,6 +153,7 @@ const routerAttr = computed(() => ({
 
 const computedLinkClasses = computed(() => ({
   active: activeBoolean.value,
+  [props.activeClass ?? '']: activeBoolean.value,
```
VividLemon, line 156: activeClass is used above. Which indicates that there was some issue somewhere. This may warrant some addition investigation on how Bootstrap-vue did this. Do you mind doing some investigation on this for me?

packages/bootstrap-vue-next/src/components/BNav/BNavItem.vue
```diff
@@ -4,7 +4,7 @@
       class="nav-link"
       :class="linkClasses"
       v-bind="{...linkProps, ...linkAttrs}"
-      active-class="active"
+      :active-class="activeClass ?? 'active'"
```
VividLemon, line 7: props.activeClass will never be undefined by definition

packages/bootstrap-vue-next/src/components/BLink/BLink.vue
```diff
@@ -38,7 +38,7 @@ defineSlots<{
 
 const props = withDefaults(defineProps<BLinkProps>(), {
   active: undefined,
-  activeClass: 'router-link-active',
+  activeClass: undefined,
```
VividLemon, line 41: Why was this changed

packages/bootstrap-vue-next/src/composables/useTableItems.ts
```diff
@@ -0,0 +1,158 @@
...
+  perPage?: number
+  currentPage?: number
+  filter?: string
+  filterable?: string[]
+  sortBy?: string
+  sortDesc?: Booleanish
+  sortCompare?: BTableSortCompare
+}
+
+export const useTableItems = (
```
VividLemon, line 16: Only export this as default. Using a named export here in addition to a default export is redundant

packages/bootstrap-vue-next/src/composables/useTableItems.ts
```diff
@@ -0,0 +1,158 @@
...
+    internalItems,
+    computedItems,
+    updateInternalItems,
+    filteredHandler,
+    notifyFilteredItems,
+    computedDisplayItems,
+  }
+}
+
+const mapItems = (
```
VividLemon, line 81: These should probably go above the exported item. I'm surprised it didn't error out with 'no use before define', but the `export` likely solves that issue.

packages/bootstrap-vue-next/src/components/BTable/BTable.vue
```diff
@@ -556,75 +357,126 @@ const callItemsProvider = async () => {
...
-    : undefined,
-  {
-    [`table-${field.variant}`]: field.variant !== null,
-    'b-table-sticky-column': field.stickyColumn,
-  },
-]
+const getRowClasses = (
+  item: TableItem | null,
+  type: string
+): string | Array<any> | null | undefined => {
```
VividLemon, line 369:
```suggestion
): string | any[] | null | undefined => {
```
Use [] syntax over Array syntax

packages/bootstrap-vue-next/src/components/BTable/BTable.vue
```diff
@@ -698,3 +548,4 @@ defineExpose({
   refresh: callItemsProvider,
 })
 </script>
+./tableItems ../../composables/tableItems
```
VividLemon, line 551: What is this. How did this not error immediately? lol

packages/bootstrap-vue-next/src/components/BTable/BTable.vue
```diff
@@ -685,9 +537,7 @@ watch(
 )
 
 onMounted(() => {
```
VividLemon, line 539: Technically you can make this now
`onMounted(callItemsProvider)`

VividLemon (changes requested):

apps/docs/src/components/TableOfContentsNav.vue
devhus: I know the icon doesn't need to be a function, but it prevents others from having issues when converting the variable to a reactive variable if they ever do.
So it will stay as a function because there is no harm from doing so

packages/bootstrap-vue-next/src/composables/useTableItems.ts
devhus: That's what a call a "God blessing that avoids the bomb"

packages/bootstrap-vue-next/src/components/BTable/BTable.vue
devhus: again, a God blessing that avoids the bomb

devhus:
@VividLemon that should do it, I don't think any more updates will be added to this PR (except fixes). Feel free to review and publish when you can so i can start writing the Docs.

VividLemon:
@devhus an issue that I noticed recently on the table component was the lack of customization for the selectable feature in the btable component.

The issue is that youre not able to modify the selectable list from outside the component, thus you're also unable to set default selectables afaik. This is drawback that bootstrap-vue suffers from. My idea: replace the isSelected list with a passive vmodel. This way we can modify the selected list externally, and also set defaults.

I've made this issue https://github.com/bootstrap-vue-next/bootstrap-vue-next/issues/1310

PR #1306, opened 2023-07-19
fix(BCarousel): carousel-item class disappears after transition

### LINKED ISSUE #1303: BCarousel carousel-item class disappears after transition
BCarouselSlide components are created with the class `carousel-item`. But after transition from one slide to another, both components (previous and current slide) lose this class.

I assume the problem is related to the presence of `carousel-item` in the computed properties `enterClasses` and `leaveClasses` that configure the TransitionGroup and the comment about the hack to make the transition work.

bielfrontera:
# Describe the PR
fixes #1303.

It adds `carousel-item` class to the slide component using the `afterEnter` event. Therefore, the component recovers the class removed by the transition.

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

### DIFF
--- a/packages/bootstrap-vue-next/src/components/BCarousel/BCarousel.vue
+++ b/packages/bootstrap-vue-next/src/components/BCarousel/BCarousel.vue
@@ -32,6 +32,7 @@
         :leave-to-class="leaveClasses"
         @before-leave="onBeforeLeave"
         @after-leave="onAfterLeave"
+        @after-enter="onAfterEnter"
       >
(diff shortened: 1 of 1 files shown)


VividLemon:
I'm going to let this sit for a bit and revisit it later. I want to understand why this class is even declared. I remember having an issue with it simply being static, but I'm not sure why it behaves so strangely. Perhaps you, the author, could do some research and possibly remove the hack altogether?

Issue #957, opened 2023-03-02
BTable fields don't work with nested objects for keys


NNeto:
**Describe the bug.**
In BootstrapVue we can do something like this:
```
fields: [
  { key: "user.name", label: "User", sortable: true },
  { key: "user.role", label: "Role", sortable: true },
],
```
with items like this:
```
items: [
  { id: 0, title: "AAAA", qty: "0", user: { name: "John", role: "admin" } },
]
```
But this doesn't work in BootstrapVueNext. The column becomes empty.

**To Reproduce.**

1. Put an attribute of a nested object of items as key in fields
2. The column in the table will be empty

Here is a codesandbox with the BootstrapVue and working example:
https://codesandbox.io/s/c5858k

Here is another codesandbox, with the same example but failing field in BootstrapVueNext:
https://codesandbox.io/s/5650c9

**Expected behavior.**
The column should be filled in with the value of the path specified at the key in the fields.

**Environment:**

- OS: Fedora Linux 35
- Package Version: 0.7.3
- Npm/Yarn/Pnpm Version: pnpm@7.28.0

xvaara:
For now you can flatten the object. For your example this would be fitting: https://www.30secondsofcode.org/js/s/flatten-object

makroxyz:
This is an important feature, hope it will be implemented soon

VividLemon:
Seems to be working now

PR #1308, opened 2023-07-19
BTable items nested object

### LINKED ISSUE #1307: BTable busy slot with provider prop
version 0.9.26

When BTable has provider property the ```table-busy``` slot is always shown as last line of the table.

in BTable.vue row 383
```vue
const classesArray = [{'b-table-static-busy': props.items.length === 0}]
```

b-table-static-busy class is always true since props.items.length is empty when ```provider``` property is set 

this is a screenshot of what I say:
Screenshot_20230719_142454

should be: 
Screenshot_20230719_142657

makroxyz:
# This PR solves #957 

Show nested objects table rows also using a dot key. Look at issue #957 

Fixed #1307

Added ```busyLoadingText``` prop

### DIFF
--- a/packages/bootstrap-vue-next/src/components/BTable/BTableLite.vue
+++ b/packages/bootstrap-vue-next/src/components/BTable/BTableLite.vue
@@ -70,7 +70,7 @@
             <slot
               v-if="$slots['cell(' + field.key + ')'] || $slots['cell()']"
               :name="$slots['cell(' + field.key + ')'] ? 'cell(' + field.key + ')' : 'cell()'"
-              :value="item[field.key]"
+              :value="get(item, field.key)"
               :index="itemIndex"
               :item="item"
               :field="field"
@@ -152,6 +152,7 @@ import type {
 import BTableSimple from './BTableSimple.vue'
 import {filterEvent} from './helpers/filter-event'
 import type {TableFieldObjectFormatter} from '../../types/TableFieldObject'
+import {get} from '../../utils/object'
 
 const props = withDefaults(
   defineProps<{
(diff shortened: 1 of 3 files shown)


VividLemon:
I can't accept an addition to the dependencies. I apologize

makroxyz:
Ok, tomorrow I will try w/o adding any dependency

VividLemon:
> Ok, tomorrow I will try w/o adding any dependency

Its an implicit rule. I debated for awhile if adding vueuse was worth it. However, in that case, the decision was great as we use vueuse extensively. You can either copy the code from their library, reinvent the wheel, or pick one up on SO. There doesn't seem to be some great out of the box way in JS to achieve that feature.

> Anyway can I add a busyText or loadingText prop to BTable in order to translate the "loading" message in table-busy slot without override the entire slot?

Certainly. I'm not sure the best name for the prop, perhaps "busyLoadingText"? It could be beneficial to add the prop and also a slot to make it more vercetile. Any hard coded text should be in a prop or slot for obvious i18n reasons.

dnewkerk:
@makroxyz thanks for working on this... need this feature from the old BootstrapVue as well.

@VividLemon instead of a dependency, perhaps we could just add a `get` method to `utils/object.ts`?:
https://github.com/bootstrap-vue-next/bootstrap-vue-next/blob/main/packages/bootstrap-vue-next/src/utils/object.ts

This is the implementation in Radash:
https://github.com/rayepps/radash/blob/master/src/object.ts#L214
https://radash-docs.vercel.app/docs/object/get

makroxyz:
Have a look if now it's OK

packages/bootstrap-vue-next/src/components/BTable/BTableLite.vue
```diff
@@ -152,6 +152,7 @@ import type {
 import BTableSimple from './BTableSimple.vue'
 import {filterEvent} from './helpers/filter-event'
 import type {TableFieldObjectFormatter} from '../../types/TableFieldObject'
+import {get} from '../../utils/object'
```
VividLemon, line 155: Namespace this from utils/ not utils/object

VividLemon (changes requested):

makroxyz:
done! 
I also removed row 358  from BTableLite.vue

PR #1355, opened 2023-08-15
fix (BPopover): cancel show if hide called #1300

### LINKED ISSUE #1300: BPopover does not maintain showDelay
"bootstrap-vue-next": "^0.9.26"
Issue can be seen in sandbox:- https://codesandbox.io/s/relaxed-antonelli-7zhdwy

Create a number of popovers in a list. Initially moving from the top item to the next, in each case waiting for the popover the delay, it behaves as expected.

Now moving the mouse quickly over each item, the popovers are shown immediately with no delay.

Issue can be seen in sandbox:- https://codesandbox.io/s/relaxed-antonelli-7zhdwy

As mentioned move slowly down the list hovering over each link, the delay works fine. Move slowly up the list, again the delay is fine. Now move the mouse quickly over the links and now no delays the popovers are shown right away.

xvaara:
close #1300 

we cancel show (clearTimeout) if hide is called.

### DIFF
--- a/packages/bootstrap-vue-next/src/components/BPopover.vue
+++ b/packages/bootstrap-vue-next/src/components/BPopover.vue
@@ -378,6 +378,7 @@ const buildTriggerableEvent = (
     componentId: computedId.value,
   })
 
+let showTimeout: ReturnType<typeof setTimeout> | undefined
 const show = () => {
   const event = buildTriggerableEvent('show', {cancelable: true})
   emit('show', event)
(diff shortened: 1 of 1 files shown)


xvaara:
> It may be preferred to use the https://vueuse.org/shared/useTimeoutFn/#usetimeoutfn utilities to get rid of the showTimeout/clearTimeout stuffs

I feel that just hides the clear timeout logic behind the curtains of vueuse function. we don't save on code lines or complexity, just change the functions to non native.

PR #1364, opened 2023-08-17
Remove unused blur BOverlay (BModal)

adiantek:
Blur with 0px on latest version of Chrome 115.0.5790.171 with **CPU only** causes an FPS drop

# Describe the PR

Blur with 0 px on the latest Chrome (production version) causes an FPS drop. In performance, it's observable as "Paint". Blur with 0 px doesn't change visually much (some pixels has a slightly different color, but really - it's not visible), so IMHO it's simply unused. 
image

Maybe if you want to create in the future a property to set the blur, it should check if size > 0 (it looks cool except for performance if you use the software rendering).

## Small replication

Disable GPU rendering:
image

Create any modal and test it:
image

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

### DIFF
--- a/packages/bootstrap-vue-next/src/components/BModal.vue
+++ b/packages/bootstrap-vue-next/src/components/BModal.vue
@@ -90,7 +90,6 @@
             no-spinner
             fixed
             no-wrap
-            blur="0px"
             @click="hide('backdrop')"
           />
         </slot>


VividLemon:
Removing 0px on BOverlay will cause the default value to be set - 2px blur. Which is what we are avoiding. 

Using BOverlay as the backdrop for BModal is to utilize it as a backdrop, not the blur. So the stuff that comes with it, the variant, events, etc. 

If you want to add this feature, what you will need to do is

Fix BOverlay to 
```ts
blur?: string | null
```

Then 
```html
:blur="null"
```
Or 
```html
blur=""
```

The top will need to be implemented regardless, the bottom can be fixed in either/or, with my preferred solution being null as it's more explicit.

VividLemon (changes requested):
Removing 0px on BOverlay will cause the default value to be set - 2px blur. Which is what we are avoiding. 

Using BOverlay as the backdrop for BModal is to utilize it as a backdrop, not the blur. So the stuff that comes with it, the variant, events, etc. 

If you want to add this feature, what you will need to do is

Fix BOverlay to 
```ts
blur?: string | null
```

Then 
```html
:blur="null"
```
Or 
```html
blur=""
```

The top will need to be implemented regardless on BOverlay, it's type is missing, the bottom can be fixed in either/or, with my preferred solution being null as it's more explicit.

VividLemon:
Also 
> Pull request title and all commits follow the [Conventional Commits](https://www.conventionalcommits.org/) convention or has an [override](https://github.com/googleapis/release-please#how-can-i-fix-release-notes) in this pull request body This is very important, as the CHANGELOG is generated from these messages, and determines the next version type. Pull requests that do not follow conventional commits or do not have an override will be denied

VividLemon:
Building off of the work described here, I merged in https://github.com/bootstrap-vue-next/bootstrap-vue-next/pull/1373. Closing this

Issue #1276, opened 2023-07-11
top-row and bottom-row slot missing from b-table


stevemandl:
**Is this feature a Bootstrap-vue feature?** Yes. Documentation here: https://bootstrap-vue.org/docs/components/table#table-helper-components

When a `<template #top-row=...` or `<template #top-row=...` is included inside the b-table, the corresponding table row should be added.

VividLemon:
Seems to be working now

PR #1365, opened 2023-08-17
Top and bottom row slot support in BTable

therecluse26:
feat: Added missing support for top-row and bottom-row slots in BTableLite (and consequently BTable)

Enhancement for #1276


VividLemon:
feat: added top-row and bottom-row slots to b-table and b-table-lite

Issue #1398, opened 2023-08-26
Regression introduced by version 0.10.8 when using component BFormInput 


tbl0605:
Hi,
In my case, the patch [0b3fcc5](https://github.com/bootstrap-vue-next/bootstrap-vue-next/commit/0b3fcc58f84f056ab25c1fb772145da8989563af) (see issue #1089) from version [0.10.8](https://github.com/bootstrap-vue-next/bootstrap-vue-next/compare/bootstrap-vue-next-v0.10.7...bootstrap-vue-next-v0.10.8) introduced a regression when using an BFormInput component. In some cases its content is not updated correctly, probably because I manually update its v-model, the underlying input content and use the `focus()` method (all at same time). 

I tried to provide a short demo project, by extracting relevant code from my project, but it's still a bit complicated code, sorry for that.

Also note that equivalent code for Vuejs 2.7 and bootstrap-vue v2.23.0 works as expected (since a long time now).

**To Reproduce.** Steps to reproduce the behavior:

1. get the [demo](https://github.com/tbl0605/boostrap-vue-next-demo2) project: `git clone https://github.com/tbl0605/boostrap-vue-next-demo2`
2. npm install
3. npx vite --port=4000
4. write "PAR" (without the quotes) in the BFormInput field
5. click on the "75000 PARIS" line **using the mouse** (this point is important)
6. the field content should be updated to "PARIS" but it doesn't work anymore

![chrome-capture-2023-7-26](https://github.com/bootstrap-vue-next/bootstrap-vue-next/assets/3315731/7dec7a5a-cfa3-4248-b868-c4ebec7c3703)

**Expected behavior.**
When you write "PAR" (without the quotes) in the BFormInput field and click on the "75000 PARIS" line using the mouse, the field content should be updated to "PARIS".

**Workaround.**
Using version <= 0.10.7 of bootstrap-vue-next will make everything work again.

**Environment:**

- OS: Windows 10
- Browsers: latest Chrome and Firefox
- Package Version: >= 0.10.8

xmatthias:
i think an even simpler reproduction can be with a input-field and a "reset" button - where the UI will simply ignore the reset-button (but the model below is properly reset).

I'm not sure about the >=10.8 version - i've tried this now with 10.7, and experience the same - all external updates to the v-model are ignored by the component - also on 10.7

xvaara:
[...]

PR #1399, opened 2023-08-27
Fix for regression introduced by version 0.10.8 when using component BFormInput #1398

tbl0605:
# Describe the PR
See bug report #1398
This a minimal & quick fix for my issue, hope it doesn't introduce other regressions...


VividLemon:
Doing this overwrites why I removed the code in the first place. There is likely a better solution, one that isn't known to me at this time.

tbl0605:
> Doing this overwrites why I removed the code in the first place. There is likely a better solution, one that isn't known to me at this time.

Ok, I thought the initial problem you tried to solve was that no "change" event was ever sent ^^

Replacing my patch with:
```
    const nextModel = _getModelValue(formattedValue)
    if (modelValue.value !== nextModel) {
      inputValue = value
      updateModelValue(formattedValue, true)
    }
```
would also solve the problem for me (I tested it). I think it's a valid alternative, because I don't see why the code should "force update" the model value (using debounce or not, lazy or not) when the value hasn't changed meanwhile. The same reasoning could probably be applied to the `onBlur` method. Of course, I could be wrong, I didn't dive into your code, but for sure the current code breaks things that worked well before ^^

PR #1417, opened 2023-09-05
Navbar fixes

### LINKED ISSUE #1283: [BNavItem] Allow disabling navbar auto-collapse
When you click on a `b-nav-item` in a `b-collapse` inside a `b-navbar`, the `b-collapse` is auto collapsed. This behavior has [recently](https://github.com/bootstrap-vue-next/bootstrap-vue-next/pull/1090) been altered; now it auto-collapses `b-collapse` solely in `b-navbar`.

I am using a `b-navbar` in my sidebar, styled it to be vertical, and added some `b-nav-item` and `b-collapse` components, a few levels deep. If I click on an item on e.g. the third level,  the parent `b-collapse` auto-collapses. I would like to be able to opt-out of this behavior.

I tried adding the `is-nav` prop to the `b-collapse` component, but (correct me if I'm wrong) that only "force"-shows the content, and disables the ability to `show` or `hide` the `b-collapse` via the `v-b-toggle` directive.

### LINKED ISSUE #1393: BNavItemDropdown is missing show & hide events
**Is this feature a Bootstrap-vue feature?** 
see: https://bootstrap-vue.org/docs/components/nav#comp-ref-b-nav-item-dropdown-events
@hide is not handled/firing:
`
<b-nav-item-dropdown left no-caret @hide="showStatus = false">
   <b-dropdown-item v-for="status in statuses" :key="status.id" @click="selectedStatus = status">{{ status.name }}</b-dropdown-item>
</b-nav-item-dropdown>
`

**Additional context.** 
show en hide are emitted from BDropdown child component, but they are not emitted from BNavItemDropdown

xvaara:
fix(BNavItemDropdown): all props and emits from dropdown
closes #1393 
feat(BNavbar): add auto-close to disable collapse closing
closes #1283

### DIFF
--- a/packages/bootstrap-vue-next/src/components/BDropdown/BDropdown.vue
+++ b/packages/bootstrap-vue-next/src/components/BDropdown/BDropdown.vue
@@ -74,7 +74,7 @@ import {onClickOutside, onKeyStroke, useToNumber, useVModel} from '@vueuse/core'
 import {computed, provide, readonly, ref, toRef, watch} from 'vue'
 import {useBooleanish, useId} from '../../composables'
 import type {Booleanish, ButtonType, ButtonVariant, ClassValue, Size} from '../../types'
-import {BvEvent, dropdownInjectionKey, resolveFloatingPlacement} from '../../utils'
+import {BvTriggerableEvent, dropdownInjectionKey, resolveFloatingPlacement} from '../../utils'
 import BButton from '../BButton/BButton.vue'
 import type {RouteLocationRaw} from 'vue-router'
 
@@ -160,9 +160,9 @@ const props = withDefaults(
 )
 
 const emit = defineEmits<{
-  'show': [value: BvEvent]
+  'show': [value: BvTriggerableEvent]
   'shown': []
-  'hide': [value: BvEvent]
+  'hide': [value: BvTriggerableEvent]
   'hidden': []
   'hide-prevented': []
   'show-prevented': []
@@ -290,16 +290,7 @@ const buttonClasses = computed(() => [
 const dropdownMenuClasses = computed(() => props.menuClass)
 
 const onButtonClick = () => {
-  emit('toggle')
-  const currentModelValue = modelValueBoolean.value
-  const e = new BvEvent(currentModelValue ? 'hide' : 'show')
-  currentModelValue ? emit('hide', e) : emit('show', e)
-  if (e.defaultPrevented) {
-    currentModelValue ? emit('hide-prevented') : emit('show-prevented')
-    return
-  }
-  modelValue.value = !currentModelValue
-  currentModelValue ? emit('hidden') : emit('shown')
+  toggle()
 }
 
 const onSplitClick = (event: MouseEvent) => {
@@ -322,13 +313,22 @@ const onClickInside = () => {
 }
 
 const close = () => {
-  modelValue.value = false
+  modelValue.value && toggle()
 }
 const open = () => {
-  modelValue.value = true
+  modelValue.value || toggle()
 }
 const toggle = () => {
-  modelValue.value = !modelValueBoolean.value
+  emit('toggle')
+  const currentModelValue = modelValueBoolean.value
+  const e = new BvTriggerableEvent(currentModelValue ? 'hide' : 'show')
+  currentModelValue ? emit('hide', e) : emit('show', e)
+  if (e.defaultPrevented) {
+    currentModelValue ? emit('hide-prevented') : emit('show-prevented')
+    return
+  }
+  modelValue.value = !currentModelValue
+  currentModelValue ? emit('hidden') : emit('shown')
 }
 
 watch(modelValueBoolean, update)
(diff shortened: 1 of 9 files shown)


packages/bootstrap-vue-next/src/components/BLink/BLink.vue
```diff
@@ -164,7 +159,10 @@ const clicked = (e: MouseEvent): void => {
     return
   }
 
-  if (collapseData?.isNav?.value === true) {
+  if (
+    (collapseData?.isNav?.value === true && navbarData === null) ||
+    (navbarData !== null && navbarData?.autoClose?.value === true)
```
VividLemon, line 164:
```suggestion
    (navbarData !== null && navbarData.autoClose?.value === true)
```

packages/bootstrap-vue-next/src/components/BNav/BNavItemDropdown.vue
```diff
@@ -15,65 +28,101 @@
...
-import type {Booleanish, ButtonVariant, ClassValue, Size} from '../../types'
-import {omit} from '../../utils'
+import type {Boundary, Middleware, RootBoundary, Strategy} from '@floating-ui/vue'
+import {ref} from 'vue'
+import type {Booleanish, ButtonType, ButtonVariant, ClassValue, Size} from '../../types'
+import {BvTriggerableEvent} from '../../utils'
 import BDropdown from '../BDropdown/BDropdown.vue'
+import type {RouteLocationRaw} from 'vue-router'
 
 // TODO share props with BDropdown
```
VividLemon, line 38: You may be better off implementing this TODO note

VividLemon (changes requested):

packages/bootstrap-vue-next/src/components/BNav/BNavItemDropdown.vue
xvaara: It seems like vue doesn't yet implement complex props with typescript. They have one difference (isNav) so the need to have their own. Or am I mistaken here? I did think of that and tried to look it up, and came to that conclusion.

Other component that could use this is the tooltip.

xvaara: Well it works with the same props now.

packages/bootstrap-vue-next/src/components/BDropdown/BDropdown.vue
```diff
@@ -74,95 +74,91 @@ import {onClickOutside, onKeyStroke, useToNumber, useVModel} from '@vueuse/core'
...
-    noCaret: false,
-    toggleText: 'Toggle dropdown',
-    variant: 'secondary',
-    modelValue: false,
-    strategy: 'absolute',
-    splitTo: undefined,
-    boundary: 'clippingAncestors',
-  }
-)
+export interface Props {
```
VividLemon, line 84: Move this to /types/ComponentProps

xvaara:
I'll do a real props merge in a different pull request and take a look at the popover/tooltip props at the sametime.

packages/bootstrap-vue-next/src/components/BNav/BNavItemDropdown.vue
VividLemon: > It seems like vue doesn't yet implement complex props with typescript. They have one difference (isNav) so the need to have their own. Or am I mistaken here? I did think of that and tried to look it up, and came to that conclusion.

I don't understand what you mean here. You're capable of doing these things in complex ways. BTable is a good example of this with extending prop definitions around and omitting certain values -- or blink, which often omits 'event' (and one other prop I forgot the name of). 

Technically you could do `defineProps<Omit<Props, 'isNav'>>` but in this case, the way I see it, it's on them to make such a mistake... Or we could just prevent them from making silly mistakes idrc. Maybe there's a reason to turn off isNav, but probably not.

> It seems like vue doesn't yet implement complex props with typescript 

You'd have to define what this is and what you're trying to do. Because it's not quite clear.

VividLemon (changes requested):

packages/bootstrap-vue-next/src/components/BNav/BNavItemDropdown.vue
VividLemon: Keep in mind, that part of the issue might exist with actually doing

```ts
export interface Props {
```

in the actual vue file itself. You _can_ do as I described with omit and whatnot, but you can't do it if it's in another Vue file.

packages/bootstrap-vue-next/src/components/BDropdown/BDropdown.vue
xvaara: done

packages/bootstrap-vue-next/src/components/BDropdown/BDropdown.vue
```diff
@@ -67,97 +67,53 @@ import {
...
   type RootBoundary,
   shift,
-  type Strategy,
   useFloating,
 } from '@floating-ui/vue'
 import {onClickOutside, onKeyStroke, useToNumber, useVModel} from '@vueuse/core'
 import {computed, provide, readonly, ref, toRef, watch} from 'vue'
 import {useBooleanish, useId} from '../../composables'
-import type {Booleanish, ButtonType, ButtonVariant, ClassValue, Size} from '../../types'
+import type {BDropdownProps} from 'src/types'
```
VividLemon, line 75: SRC/types is invalid

PR #1430, opened 2023-09-11
feat: Add useModal() and useModalController()

asika32764:
# Describe the PR

This PR introduce a `useModal()` and `useModalController()` hooks to help us control modal programmatically and globally.

## `useModal()`

```ts
const { hide, modal } = useModal(); // Get closest modal in child components

if (...) {
  hide()
}
```

```vue
<template>
<BModal id="hello"></BModal>
</template>

<script>
const { show, hide, modal } = useModal('hello'); // Get modal by id, you can use this anywhere

// Must check modal is exists
if (modal.value) {
  show()
}
</script>
```

## `useModalController()`

modalController can use in anywhere, currently we only implement `hid()` and `hideAll()`

the `hide()` will hide last modal in the stack.

```ts
const modalController = useModalController()

modalController.hide()
modalController.hideAll()
```

## Test

Use this code to test

`App.vue`

```vue
<template>
  <BContainer>
    <button type="button" @click="showById">Show By ID</button>

    <BModal id="hello">
      <ModalTest />
    </BModal>
  </BContainer>
</template>

<script setup lang="ts">
import {BModal, useModal} from './BootstrapVue'
import ModalTest from './ModalTest.vue'

const modal = useModal('hello')

function showById() {
  modal?.show()
}
</script>
```

`ModalTest.vue`

```vue
<script setup lang='ts'>
import {ref} from 'vue'
import {type BModalController, useModal, useModalController} from './BootstrapVue'

const { hide } = useModal()
const modalController: BModalController = useModalController()
const secondModal = ref(false);

function close() {
  modalController.hide()
}
</script>

<template>
<div>
  Hello
  <button type='button' @click='hide'>
    Hide Closest Modal
  </button>
  <button type='button' @click='modalController.hide'>
    Hide By Controller (Sequential)
  </button>
  <button type='button' @click='secondModal = true'>
    Show Second Modal
  </button>

  <BModal v-model='secondModal'>
    <button type='button' @click='modalController.hide'>
      Hide By Controller (Sequential)
    </button>
    <button type='button' @click='modalController.hideAll'>
      Hide All
    </button>
  </BModal>
</div>
</template>

<style scoped lang='scss'>

</style>

```

and run `pnpm dev`

## PR checklist

- [x] Feature - `feat(...)`
- [ ] Other (please describe)


IGx89:
This looks really useful, thank you!

asika32764:
the controller show() must direct to a particular modal, that is the purpose of `useModal()` and Vue ElementRef.

modalController is to simply close modals in the stack and no need to known which is active.

See also Ionic, their `modalController` also has no `show()` method, only `dismiss()` method. https://ionicframework.com/docs/api/modal#methods

packages/bootstrap-vue-next/src/composables/useModalManager.ts
```diff
@@ -1,16 +1,21 @@
...
+
+export const useSharedModalStack = createSharedComposable(() => {
+  const registry = ref<ModalInstance[]>([])
+  const stack = ref<ModalInstance[]>([])
+  const count = computed(() => stack.value.length)
+  const last = computed(() => stack.value[stack.value.length - 1])
+  const push = (modal: ModalInstance) => stack.value.push(modal)
+  const pop = () => stack.value.pop()
+  const remove = (modal: ModalInstance) =>
+    (stack.value = stack.value.filter((item: ModalInstance) => item.uid !== modal.uid))
```
VividLemon, line 16:
```suggestion
  const remove = (modal: ModalInstance) => {
    stack.value = stack.value.filter((item: ModalInstance) => item.uid !== modal.uid)
  }
```
The issue with this, and the ones above is that doing parenthesis, or just leaving it without, is that you can (sometimes) get incorrect return types. Take for example 

https://www.typescriptlang.org/play?#code/DYUwLgBAxhC8EEYBQSoHsB2BnSBDOEAFAJRwB8EA3khLdATANSJIC+K62kARgSQFwQAbmgCWAE3IRcJFAHoAVKkw4IveCSnU602ewVyOKyJPjdZy7GlAA6YGgDmhccUtZrIO48JRiQA

https://www.typescriptlang.org/play?#code/DYUwLgBAxhC8EEYBQSoHsB2BnSBDOEAFAJRwB8EA3khLRAE7gCu9G0BMA1IkgL4rpskAEYESALggA3NAEsAJuQi4SKAPQAqVJhwRR8Ekup1lq-hrUCdkRfGGrt2NKAB0wNAHNC84o6zOQN09CKGIgA

Let's say for example you wanted b to return `void`. Now you have some refactoring to do. Let's say at some point a switched from void => number. Subsequently, b would switch unexpectedly. There is two ways to be sure in this context, either declare the return type with `void`, or to simply wrap in `{}`. I prefer the latter (as you would need to anyways, usually)

This is not _usually_ an issue with built-in methods such as Array.filter, Array,pop, as shown above, (as they will never change! Imagine the chaos!) But it's best to be safe

packages/bootstrap-vue-next/src/composables/useModalManager.ts
```diff
@@ -1,16 +1,21 @@
...
+  const find = (id: string): ModalInstance | null =>
```
VividLemon, line 17: It could be useful that instead of returning null, you return undefined. This would bring it more in line with Array.find.

packages/bootstrap-vue-next/src/composables/useModalController.ts
```diff
@@ -0,0 +1,29 @@
+import {useSharedModalStack} from './useModalManager'
+
+export interface BModalController {
```
VividLemon, line 3: I'd prefer this type to be in the types folder. I'm a bit half and half about this, as it makes sense for the type to be declared here. But I'm preferring standardization rather than logic

packages/bootstrap-vue-next/src/composables/exports/index.ts
```diff
@@ -1 +1 @@
-export {useBreadcrumb, useColorMode} from '..'
+export {useBreadcrumb, useColorMode, useModal, useModalController, type BModalController} from '..'
```
VividLemon, line 1: Subsequently the moving of the type, you wouldn't export the type from here.

VividLemon (changes requested):

packages/bootstrap-vue-next/src/composables/useModalManager.ts
asika32764: It seems the code fixer in git hooks makes this parenthesis.

packages/bootstrap-vue-next/src/composables/useModalManager.ts
asika32764: Mostly I will use undefined, but in this case, I try to keep a same behavior of Vue `getCurrentInstance()`

```
// Vue getCurrentInstance declaration
export declare const getCurrentInstance: () => ComponentInternalInstance | null;
```

See https://github.com/bootstrap-vue-next/bootstrap-vue-next/pull/1430/files/672a7f782fcf1f04bd0d027a7ee2af65ffa8de37#diff-5d21bc2f3956bd83b4d165bb5d182a28f24483d905793b29221ca708243ade52R10

the `find()` works here that can keep the same return types as instance. And I think `find()` is a internal fuction so this will not be a big issue, unless we want to expose it.

packages/bootstrap-vue-next/src/composables/useModalController.ts
asika32764: I'll modify this when I reach my working device.

asika32764:
There's some more changes:
- Correct all type hinting
- To avoid using internal api, we must expose BModal id value.

packages/bootstrap-vue-next/src/composables/useModalManager.ts
```diff
@@ -1,26 +1,28 @@
...
   const last = computed(() => stack.value[stack.value.length - 1])
-  const push = (modal: ModalInstance) => {
-    stack.value.push(modal)
-  }
-  const pop = () => {
-    stack.value.pop()
-  }
-  const remove = (modal: ModalInstance) => {
-    stack.value = stack.value.filter((item: ModalInstance) => item.uid !== modal.uid)
+  const push = (modal: ComponentInternalInstance) => stack.value.push(modal)
```
asika32764, line 20: It's seems git hooks still changes some code style

packages/bootstrap-vue-next/src/composables/useModalManager.ts
```diff
@@ -1,16 +1,28 @@
...
 
 const MODAL_OPEN_CLASS_NAME = 'modal-open'
 
-const useSharedModalCounter = createSharedComposable(() => {
-  const {count, inc, dec} = useCounter()
+export const useSharedModalStack = createSharedComposable(() => {
+  // Currently vue-tsc will complain a type error when using `ref<ComponentInternalInstance[]>`
+  // We use `as Ref<ComponentInternalInstance[]>` instead.
+  const registry = ref([]) as Ref<ComponentInternalInstance[]>
+  const stack = ref([]) as Ref<ComponentInternalInstance[]>
```
VividLemon, line 17:
```suggestion
  const registry: Ref<ComponentInternalInstance[]> = ref([])
  const stack: Ref<ComponentInternalInstance[]> = ref([]) 
```

packages/bootstrap-vue-next/src/composables/useModalManager.ts
```diff
@@ -1,16 +1,28 @@
...
+  watch,
+} from 'vue'
 
 const MODAL_OPEN_CLASS_NAME = 'modal-open'
 
-const useSharedModalCounter = createSharedComposable(() => {
-  const {count, inc, dec} = useCounter()
+export const useSharedModalStack = createSharedComposable(() => {
+  // Currently vue-tsc will complain a type error when using `ref<ComponentInternalInstance[]>`
+  // We use `as Ref<ComponentInternalInstance[]>` instead.
```
VividLemon, line 15:
```suggestion
```

VividLemon (changes requested):

PR #1461, opened 2023-09-16
package.json export src

xvaara:
I think we should export the whole src directory.


VividLemon:
Yeah this likely would have caused an error warning talking about "this import is possible, but will not respect the libraries exports" if you were to import anything else... I think importing from /src/types would suffer from this.

PR #1487, opened 2023-09-25
BTableLite foot-clone with foot() slots

makroxyz:
# Describe the PR

When ```foot-clone``` is enabled header will be cloned to footer.
PR also allows footer slots ```foot()``` / ```foot({key})```

### DIFF
--- a/packages/bootstrap-vue-next/src/components/BTable/BTableLite.vue
+++ b/packages/bootstrap-vue-next/src/components/BTable/BTableLite.vue
@@ -124,6 +124,7 @@
     </tbody>
     <tfoot v-if="footCloneBoolean">
       <tr>
+        <slot name="tfoot-tr-prefix" />
         <th
           v-for="field in computedFields"
           :key="field.key"
--- a/packages/bootstrap-vue-next/src/components/BTable/BTableLite.vue
+++ b/packages/bootstrap-vue-next/src/components/BTable/BTableLite.vue
@@ -127,15 +127,28 @@
         <th
           v-for="field in computedFields"
           :key="field.key"
-          v-bind="field.thAttr"
           scope="col"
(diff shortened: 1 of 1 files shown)


packages/bootstrap-vue-next/src/components/BTable/BTableLite.vue
```diff
@@ -127,15 +127,28 @@
...
           :title="field.headerTitle"
           :abbr="field.headerAbbr"
           :style="field.thStyle"
-          @click="headerClicked(field, $event, true)"
+          v-bind="field.thAttr"
+          @click="headerClicked(field, $event)"
         >
-          {{ field.label }}
+          <div class="d-inline-flex flex-nowrap align-items-center gap-1">
+            <slot name="field-prefix" :field="field" />
```
VividLemon, line 139: `name="field-prefix"` this seems strange

makroxyz: it is a copy of lines  from 20 to 48... to be honest a <slot name="tfoot-tr-prefix" /> is missing before <th> in line 127

packages/bootstrap-vue-next/src/components/BTable/BTableLite.vue
```diff
@@ -124,18 +124,32 @@
     </tbody>
     <tfoot v-if="footCloneBoolean">
       <tr>
+        <slot name="tfoot-tr-prefix" />
```
VividLemon, line 127: What is this

makroxyz: The header block has ’’’thead-tr-prefix’’’ I don't know if it is used or not. Probably the ’’’tfoot-tr-prefix’’’ is not so useful

PR #1496, opened 2023-09-29
fix(BModal): Allow prevent hide event.

bradlis7:
# Describe the PR

If I do `@ok.prevent`, I don't want it to hide the Modal.

## Small replication

N/A

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

### DIFF
--- a/packages/bootstrap-vue-next/src/components/BModal.vue
+++ b/packages/bootstrap-vue-next/src/components/BModal.vue
@@ -456,7 +456,9 @@ const hide = (trigger = '') => {
   if (trigger === 'close') {
     emit(trigger, event)
   }
-  emit('hide', event)
+  if (!event.defaultPrevented) {
+    emit('hide', event)
+  }
 
   if (
     event.defaultPrevented ||


VividLemon:
This already works 
image

bradlis7:
Hm, yes in this case, it does. But I want to have a hide callback that doesn't get called if prevented.

```
<BModal
  :model-value="componentState.show"
  @ok.prevent="onOkClick()"
  @hide="onHide"
>
```

I'm doing this for now.

```
  @ok.prevent="onOkClick"
  @close="onHide"
  @cancel="onHide"
```

I could check for $event.defaultPrevented in my own code, but I don't think it should trigger `@hide` at all if it's prevented earlier.

VividLemon:
> Hm, yes in this case, it does. But I want to have a hide callback that doesn't get called if prevented.
> 
> ```
> <BModal
>   :model-value="componentState.show"
>   @ok.prevent="onOkClick()"
>   @hide="onHide"
> >
> ```
> 
> I'm doing this for now.
> 
> ```
>   @ok.prevent="onOkClick"
>   @hide.prevent
>   @close="onHide"
>   @cancel="onHide"
> ```
> 
> I could check for $event.defaultPrevented in my own code, but I don't think it should trigger `@hide` at all if it's prevented earlier.

This is Bootstrap-vue compatibility

image

It's just how it works in their system. 

You have two options. In our system, you have the `hide-prevented` emit. Which you can use to modify behavior. Alternatively, which is likely the better option, the event emitted also has access to `event.defaultPrevented`. You can check this on your side of the app and modify from there.

bradlis7:
👍 I'll work around it.

PR #1508, opened 2023-10-04
fix(BModal): Fix recent regression breaking body scroll locking

IGx89:
# Describe the PR

Fix recent regression breaking body scroll locking when a modal is open and bodyScrolling isn't set to true.

Also removed unused 'show' prop, added some docs for scrollable and centered, and added some more unit tests. Docs and tests were mostly copied from bootstrap-vue.

## Small replication

Go to https://bootstrap-vue-next.github.io/bootstrap-vue-next/docs/components/modal.html, open a modal, and check the style attribute on document.body. Notice it doesn't exist, when it should be set to `overflow: hidden`

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

### DIFF
--- a/packages/bootstrap-vue-next/src/composables/useSafeScrollLock.ts
+++ b/packages/bootstrap-vue-next/src/composables/useSafeScrollLock.ts
@@ -2,8 +2,8 @@ import {type MaybeRefOrGetter, onMounted, readonly, toRef, watch} from 'vue'
 import {useScrollLock} from '@vueuse/core'
 
 export default (isOpen: MaybeRefOrGetter<boolean>, bodyScroll: MaybeRefOrGetter<boolean>) => {
-  const resolvedIsOpen = readonly(toRef(() => isOpen))
-  const resolvedBodyScrolling = readonly(toRef(() => bodyScroll))
+  const resolvedIsOpen = readonly(toRef(isOpen))
+  const resolvedBodyScrolling = readonly(toRef(bodyScroll))
 
   /**
    * We use the inverse because bodyScrolling === true means we allow scrolling, while bodyScrolling === false means we disallow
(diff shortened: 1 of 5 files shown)


VividLemon:
image
This is the solution

IGx89:
> image This is the solution

That's an alternate solution for sure. There just seemed to be a lot of unnecessary code in there so I simplified it too. I can change it to what you suggest above if you'd prefer however?

VividLemon:
> That's an alternate solution for sure. There just seemed to be a lot of unnecessary code in there so I simplified it too. I can change it to what you suggest above if you'd prefer however?

Yes

IGx89:
@VividLemon done, take a look now

packages/bootstrap-vue-next/src/composables/useSafeScrollLock.ts
```diff
@@ -1,9 +1,9 @@
-import {type MaybeRefOrGetter, onMounted, toRef, watch} from 'vue'
+import {type MaybeRefOrGetter, onMounted, readonly, toRef, watch} from 'vue'
 import {useScrollLock} from '@vueuse/core'
 
 export default (isOpen: MaybeRefOrGetter<boolean>, bodyScroll: MaybeRefOrGetter<boolean>) => {
-  const resolvedIsOpen = toRef(() => isOpen)
-  const resolvedBodyScrolling = toRef(() => bodyScroll)
+  const resolvedIsOpen = readonly(toRef(() => isOpen))
+  const resolvedBodyScrolling = readonly(toRef(() => bodyScroll))
```
VividLemon, line 6:
```suggestion
  const resolvedIsOpen = readonly(toRef(isOpen))
  const resolvedBodyScrolling = readonly(toRef(bodyScroll))
```

Issue #1510, opened 2023-10-05
NavBar Right Align doesn't work


dwgray:
### Describe the bug

I am using the first navbar sample from the [documentation](https://bootstrap-vue-next.github.io/bootstrap-vue-next/docs/components/navbar.html) and the second `BNavbarNav`, which is supposed to be right aligned according to the comments, just butts up against the first `BNavbarNav`.

![image](https://github.com/bootstrap-vue-next/bootstrap-vue-next/assets/5083020/88c8295c-05dd-4352-b509-1b3d41884ecc)

I would expect Lang and User to righted align as the comments states rather than left aligned with link and disabled.

Is there a different way to do this?  If the sample in the documentation is out of date, I'd be happy to update it.

### Reproduction

https://stackblitz.com/edit/bootstra-vue-next-template-nyzjqn?file=src%2Fcomponents%2FComp.vue

### System Info

```Shell
I don't believe this is system/browser specific - it repros with the above stackblitz on everything I have access to.
```

### Used Package Manager

n/a

### Validations

- [X] Have tested with the latest version. This is still alpha version and sometime things change rapidly.
- [X] Follow our [Code of Conduct](https://github.com/bootstrap-vue-next/bootstrap-vue-next/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guide](https://github.com/bootstrap-vue-next/bootstrap-vue-next/blob/main/CONTRIBUTING.md).
- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A, please open a GitHub Discussion instead.
- [X] The provided reproduction is a [minimal reproducible](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

VividLemon:
That's all just classes. If you want to align your items and stuff, you're going to need to use flexbox and whatnot.

dwgray:
Thanks for the quick response @VividLemon.  So this looks like it's a doc bug - I have a [stackblitz](https://stackblitz.com/edit/bootstra-vue-next-template-jvghf8?file=src%2Fcomponents%2FComp.vue) with the fix to the sample.  I just changed me-auto to ms-auto as suggested by [this SO](https://stackoverflow.com/a/65365121/2197085) answer.  If that makes sense to you, I'll take a look at fixing the sample in the docs.

PR #1511, opened 2023-10-05
docs: Navbar doc fixes 

### LINKED ISSUE #1507: `type` of `Navbar` does not control color as it is supposed to
According to [docs](https://github.com/bootstrap-vue-next/bootstrap-vue-next/blob/ad587341290cdb730d69bfda44253e5f6523738b/apps/docs/src/docs/components/navbar.md?plain=1#L92C1-L92C1), the `type` attribute of `Navbar` should control whether the text is dark or light. It doesn't work though, and it's necessary to instead set `data-bs-theme`. We could fix this either by updating the docs, or changing the `Navbar` so that it internally sets `data-bs-theme` based on `type`.

dwgray:
# Fixes two issues with the NavBar docs

1. The use of type="light|dark" has been replaced with data-bs-theme="light|dark" in order to modify the text color of the NavBar.  I've replaced all cases in the samples and updated the description.  This addresses #1507 and makes the samples readable again.
2. Changed the classes in the initial sample to make the comment stating that a portion of the navbar would be right aligned accurate.  This was a matter of changing me-auto to ms-auto to conform with bootstrap 5's utility classes.  This addresses #1510 

## Small replication

Fixes #1507 and #1510

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

### DIFF
--- a/apps/docs/src/docs/components/navbar.md
+++ b/apps/docs/src/docs/components/navbar.md
@@ -15,7 +15,7 @@ The component `BNavbar` is a wrapper that positions branding, navigation, and ot
 </div>
 
 <HighlightCard>
-  <BNavbar toggleable="lg" type="dark" variant="dark">
+  <BNavbar toggleable="lg" data-bs-theme="dark" variant="dark">
     <BNavbarBrand href="#">NavBar</BNavbarBrand>
     <BNavbarToggle target="nav-collapse" />
     <BCollapse id="nav-collapse" is-nav>
@@ -49,7 +49,7 @@ The component `BNavbar` is a wrapper that positions branding, navigation, and ot
   <template #html>
 
 ```vue-html
-<BNavbar toggleable="lg" type="dark" variant="dark">
+<BNavbar toggleable="lg" data-bs-theme="dark" variant="dark">
   <BNavbarBrand href="#">NavBar</BNavbarBrand>
   <BNavbarToggle target="nav-collapse" />
   <BCollapse id="nav-collapse" is-nav>
@@ -87,9 +87,9 @@ The component `BNavbar` is a wrapper that positions branding, navigation, and ot
 
 ## Color schemes
 
-`BNavbar` supports the standard Bootstrap v4 available background color variants. Set the variant prop to one of the following values to change the background color: primary, success, info, warning, danger, dark, or light.
+`BNavbar` supports the standard Bootstrap v5 available background color variants. Set the variant prop to one of the following values to change the background color: primary, secondary, success, info, warning, danger, dark, light, or any of the \*-subtle variants.
 
-Control the text color by setting `type` prop to `light` for use with light background color variants, or dark for dark background color variants.
+Control the text color by setting `data-bs-theme` prop to `dark` for use with dark background color variants, or `light` for light background color variants.
 
 ## Placement
 
@@ -118,14 +118,14 @@ Navbars come with built-in support for a handful of sub-components. Choose from 
 The `BNavbarBrand` generates a link if href is provided, or a `RouterLink` if to is provided. If neither is given it renders as a `<div>` tag. You can override the tag type by setting the tag prop to the element you would like rendered:
 
 <HighlightCard>
-  <BNavbar variant="faded" type="light">
+  <BNavbar variant="faded" data-bs-theme="light">
     <BNavbarBrand href="#">BootstrapVueNext</BNavbarBrand>
   </BNavbar>
   <template #html>
 
 ```vue-html
 <!-- As a link -->
-<BNavbar variant="faded" type="light">
+<BNavbar variant="faded" data-bs-theme="light">
   <BNavbarBrand href="#">BootstrapVueNext</BNavbarBrand>
 </BNavbar>
 ```
@@ -134,14 +134,14 @@ The `BNavbarBrand` generates a link if href is provided, or a `RouterLink` if to
 </HighlightCard>
 
 <HighlightCard>
-  <BNavbar variant="faded" type="light">
+  <BNavbar variant="faded" data-bs-theme="light">
     <BNavbarBrand tag="h1" class="mb-0">BootstrapVue</BNavbarBrand>
   </BNavbar>
   <template #html>
 
 ```vue-html
 <!-- As a heading -->
-<BNavbar variant="faded" type="light">
+<BNavbar variant="faded" data-bs-theme="light">
   <BNavbarBrand tag="h1" class="mb-0">BootstrapVue</BNavbarBrand>
 </BNavbar>
 ```
@@ -152,7 +152,7 @@ The `BNavbarBrand` generates a link if href is provided, or a `RouterLink` if to
 Adding images to the `BNavbarBrand` will likely always require custom styles or utilities to properly size. Here are some examples to demonstrate:
 
 <HighlightCard>
-  <BNavbar variant="faded" type="light">
+  <BNavbar variant="faded" data-bs-theme="light">
     <BNavbarBrand href="#">
       <img src="https://placekitten.com/g/30/30" alt="Kitten">
     </BNavbarBrand>
@@ -161,7 +161,7 @@ Adding images to the `BNavbarBrand` will likely always require custom styles or 
 
 ```vue-html
 <!-- Just an image -->
-<BNavbar variant="faded" type="light">
+<BNavbar variant="faded" data-bs-theme="light">
   <BNavbarBrand href="#">
     <img src="https://placekitten.com/g/30/30" alt="Kitten">
   </BNavbarBrand>
@@ -172,7 +172,7 @@ Adding images to the `BNavbarBrand` will likely always require custom styles or 
 </HighlightCard>
 
 <HighlightCard>
-  <BNavbar variant="faded" type="light">
+  <BNavbar variant="faded" data-bs-theme="light">
     <BNavbarBrand href="#">
       <img src="https://placekitten.com/g/30/30" class="d-inline-block align-top" alt="Kitten">
       BootstrapVue
@@ -182,7 +182,7 @@ Adding images to the `BNavbarBrand` will likely always require custom styles or 
 
 ```vue-html
 <!-- Image and text -->
-<BNavbar variant="faded" type="light">
+<BNavbar variant="faded" data-bs-theme="light">
   <BNavbarBrand href="#">
     <img src="https://placekitten.com/g/30/30" class="d-inline-block align-top" alt="Kitten">
     BootstrapVue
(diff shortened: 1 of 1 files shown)


VividLemon:
Use the color mode options rather than the raw data-bs-theme attribute. Then make a section about it, if there isn't one already, discussing all three options -- directive, conposable, and raw attribute

dwgray:
Okay, I looked into the options you talked about in a little more detail and I still think the primary use of overriding the text color is to use the raw attribute to override the underlying theme text color when the variant is set so that changing the light/dark theme doesn't make the text in the navbar unreadable.  This exactly what the docs are doing with the navbar - the custom primary color is dark enough that you need to set the text color to the dark theme color to make it readable.  So I changed the first sample to match that use case.

I added a line to note that the navbar conforms to theme colors if the variant property isn't set.  This seems to be the place where useColorMode is relevant.

The v-b-color-mode directive seems to only take a variable, not an constant.  So it doesn't seem to be a solution to the problem that is being addressed here.  I guess there is a case where one would programmatically change both the variant (not using themes), but that seems like a very corner case.  Let me know if I'm missing something here.

VividLemon:
/

VividLemon:
Screenshot_20231005-171008

dwgray:
Okay, done.  Not sure how I misread the code in the Layout.vue file, but I obviously did.

apps/docs/src/docs/components/navbar.md
```diff
@@ -87,9 +87,25 @@ The component `BNavbar` is a wrapper that positions branding, navigation, and ot
...
+`BNavbar` supports the standard Bootstrap v5 available background color variants. Set the `variant` prop to one of the
+following values to change the background color: primary, secondary, success, info, warning, danger, dark, light, or any of the \*-subtle variants.
 
-Control the text color by setting `type` prop to `light` for use with light background color variants, or dark for dark background color variants.
+If the `variant` property is set, it may be necessary to control the text color so that it contrasts with the variant
+irrespective of the theme. This can be done in one of two ways.
+
+Use the `v-b-color-mode` directive:
+
+```vue-html
```
VividLemon, line 98: Code blocks need to be in highlight cards

apps/docs/src/docs/components/navbar.md
```diff
@@ -91,20 +91,40 @@ The component `BNavbar` is a wrapper that positions branding, navigation, and ot
 following values to change the background color: primary, secondary, success, info, warning, danger, dark, light, or any of the \*-subtle variants.
 
 If the `variant` property is set, it may be necessary to control the text color so that it contrasts with the variant
-irrespective of the theme. This can be done in one of two ways.
+irrespective of the theme. This can be done by use the `v-b-color-mode` directive or the `data-bs-theme` attribute.
```
VividLemon, line 94: This doesn't reference `useColorMode`

apps/docs/src/docs/components/navbar.md
```diff
@@ -91,20 +91,40 @@ The component `BNavbar` is a wrapper that positions branding, navigation, and ot
...
 ```vue-html
-<BNavbar toggleable="lg" variant="primary" v-b-color-mode="'dark'">
+<!-- Using the v-b-color-mode directive -->
+<BNavbar variant="faded">
+  <BNavbarBrand tag="h1" class="mb-0">BootstrapVue</BNavbarBrand>
+</BNavbar>
 ```
 
-Or use the raw `data-bs-theme` attribute:
+</template>
```
VividLemon, line 109: Indent

apps/docs/src/docs/components/navbar.md
```diff
@@ -91,20 +91,40 @@ The component `BNavbar` is a wrapper that positions branding, navigation, and ot
...
+<HighlightCard>
+  <BNavbar variant="primary" v-b-color-mode="'dark'">
+    <BNavbarBrand tag="h1" class="mb-0">BootstrapVue</BNavbarBrand>
+  </BNavbar>
+  <template #html>
 
 ```vue-html
-<BNavbar toggleable="lg" variant="primary" v-b-color-mode="'dark'">
+<!-- Using the v-b-color-mode directive -->
+<BNavbar variant="faded">
```
VividLemon, line 104: The example for the template needs to match the codeblock from above

apps/docs/src/docs/components/navbar.md
```diff
@@ -91,20 +91,40 @@ The component `BNavbar` is a wrapper that positions branding, navigation, and ot
...
+<HighlightCard>
+  <BNavbar variant="primary" data-bs-theme="dark">
+    <BNavbarBrand tag="h1" class="mb-0">BootstrapVue</BNavbarBrand>
+  </BNavbar>
+  <template #html>
 
 ```vue-html
-<BNavbar toggleable="lg" variant="primary" data-bs-theme="dark">
+<!-- Using the data-bs-theme attribute -->
+<BNavbar data-bs-theme="dark">
```
VividLemon, line 120: The example for the template needs to match the codeblock from above

VividLemon (changes requested):

apps/docs/src/docs/components/navbar.md
VividLemon: The comment is unnecessary. But perhaps you can label it actually above each `<HighlightCard>` above

```html
With `v-b-colorMode` more info [here](#pathToVBColorMode)

<HighlightCard>
...
```

etc etc for each example

dwgray:
Sorry about that last one, just plain sloppy on my part.  

On the consistency issue - do you have any thoughts on tools to assist or enforce that consistency between the codeblock and the template?  I can think of a number of different things that _might_ improve this, but don't want to jump into anything if you've already got thoughts...

VividLemon:
> if you've already got thoughts...

I have no suggestions

dwgray:
On the template issue, I imagine that it's not that high a priority because once you get into the workflow of editing the docs it's probably easy enough to get used to doing a copy-paste of one to the other rather than editing them individually (which is what I'll do going forward). I'll think about it a bit more and if I think it's worth pursuing I'll post.  Would that be most appropriate on the discussion section or as an issue?

apps/docs/src/docs/components/navbar.md
dwgray: I believe this is addressed by `BNavbar` will conform to the current color theme if the `variant` prop is not set. If you'd like me to explicitly include how you set the current theme and put a link to useColorMode I will.

apps/docs/src/docs/components/navbar.md
dwgray: I believe this is now addressed.

apps/docs/src/docs/components/navbar.md
dwgray: I've moved these to text above the cards with links.

apps/docs/src/docs/components/navbar.md
dwgray: This is now fixed

apps/docs/src/docs/components/navbar.md
dwgray: This should now be fixed.

dwgray:
Okay, I believe I've addressed all of the comments - and hopefully the last commit with this indenting covered what you were asking about and didn't cause new problems :-)  Sorry for the amount of back and forth on this - new process for me and I use Azure Devops rather than Git in my day job, so that caused a few hiccups.  Thanks for your patience.  Let me know if you need anything else from me to get this in.

apps/docs/src/docs/components/navbar.md
```diff
@@ -87,9 +87,49 @@ The component `BNavbar` is a wrapper that positions branding, navigation, and ot
 
 ## Color schemes
 
-`BNavbar` supports the standard Bootstrap v4 available background color variants. Set the variant prop to one of the following values to change the background color: primary, success, info, warning, danger, dark, or light.
+`BNavbar` supports the standard Bootstrap v5 available background color variants. Set the `variant` prop to one of the
+following values to change the background color: primary, secondary, success, info, warning, danger, dark, light, or any of the \*-subtle variants.
 
-Control the text color by setting `type` prop to `light` for use with light background color variants, or dark for dark background color variants.
+If the `variant` property is set, it may be necessary to control the text color so that it contrasts with the variant
```
VividLemon, line 93:
```suggestion
If the `variant` prop is set, it may be necessary to control the text color so that it contrasts with the variant
```

apps/docs/src/docs/components/navbar.md
```diff
@@ -87,9 +87,49 @@ The component `BNavbar` is a wrapper that positions branding, navigation, and ot
...
+irrespective of the theme. This can be done by use the `v-b-color-mode` directive or the `data-bs-theme` attribute.
```
VividLemon, line 94:
```suggestion
irrespective of the theme. This can be done by use the `v-b-color-mode` directive or `useColorMode` composable
```

VividLemon (changes requested):

apps/docs/src/docs/components/navbar.md
```diff
@@ -87,9 +87,49 @@ The component `BNavbar` is a wrapper that positions branding, navigation, and ot
...
+
+```vue-html
+<BNavbar variant="primary" data-bs-theme="dark">
+  <BNavbarBrand tag="h1" class="mb-0">BootstrapVue</BNavbarBrand>
+</BNavbar>
+```
+
+  </template>
+</HighlightCard>
+
```
VividLemon, line 131:
```suggestion
```

apps/docs/src/docs/components/navbar.md
```diff
@@ -87,9 +87,30 @@ The component `BNavbar` is a wrapper that positions branding, navigation, and ot
...
 
-`BNavbar` supports the standard Bootstrap v4 available background color variants. Set the variant prop to one of the following values to change the background color: primary, success, info, warning, danger, dark, or light.
+`BNavbar` supports the standard Bootstrap v5 available background color variants. Set the `variant` prop to one of the
+following values to change the background color: primary, secondary, success, info, warning, danger, dark, light, or any of the \*-subtle variants.
 
-Control the text color by setting `type` prop to `light` for use with light background color variants, or dark for dark background color variants.
+If the `variant` prop is set, it may be necessary to control the text color so that it contrasts with the variant
+irrespective of the theme. This can be done by use the `v-b-color-mode` directive or `useColorMode` composable
+
+An example using `v-b-colorMode`. Form more information see the [BColorMode directive](/docs/directives/BColorMode).
```
VividLemon, line 96:
```suggestion
An example using `v-b-colorMode`. For more information see the [BColorMode directive](/docs/directives/BColorMode).
```

apps/docs/src/docs/components/navbar.md
```diff
@@ -87,9 +87,30 @@ The component `BNavbar` is a wrapper that positions branding, navigation, and ot
...
+<template #html>
+
+```vue-html
+<BNavbar variant="primary" v-b-color-mode="'dark'">
+  <BNavbarBrand tag="h1" class="mb-0">BootstrapVue</BNavbarBrand>
+</BNavbar>
+```
+
+  </template>
+</HighlightCard>
```
VividLemon, line 112:
```suggestion
</HighlightCard>

```

Issue #1516, opened 2023-10-11
Modal opens behind another modal if element is defined first


mclmax:
### Describe the bug

If I defined two modals, Modal1 and Modal2, where a button in one modal opens the other and the second modal to open is defined before the first, it opens behind the first modal.

Modal1 opens first, button in Modal1 opens Modal2, which opens underneath Modal1.
```
<Modal2 />
<Modal1 />
```

In this case it works as you'd expect. Modal1 opens, button in Modal1 opens Modal2 on top of Modal1.
```
<Modal1 />
<Modal2 />
```

In my mind the stacking should work independent of element order.

I'm super keen to get a PR for this as well if this is unintended behaviour. Was hoping to get your feedback before diving into that. If you point me in the right direction more than happy to fix.

Thanks

### Reproduction

https://stackblitz.com/edit/bootstra-vue-next-template-69h4lk?file=src%2FApp.vue

### System Info

```Shell
Not sure how to run this on stackblitz
```

### Used Package Manager

pnpm

### Validations

- [X] Have tested with the latest version. This is still alpha version and sometime things change rapidly.
- [X] Follow our [Code of Conduct](https://github.com/bootstrap-vue-next/bootstrap-vue-next/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guide](https://github.com/bootstrap-vue-next/bootstrap-vue-next/blob/main/CONTRIBUTING.md).
- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A, please open a GitHub Discussion instead.
- [X] The provided reproduction is a [minimal reproducible](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

VividLemon:
Didn't even mention that no-stacking isn't working

susanne99:
z-index is always the same. hardcoded in BModal.vue
no calculation of it !

```
.modal-dialog {
  z-index: 1051;
}
```

In my case I have an EditDialog open, and after 10 minutes of inactivity a modal dialog appears with the message that you have been logged out. Unfortunately you can't see it because it's hidden behind the edit dialog due to the same z-index.

I know that with jquery or my own typescript function I can close all open modals. A z-index calculation wouldn't be bad

mclmax:
[...]

PR #1521, opened 2023-10-16
fix(BModal): Fix new modals stacking behind active modals

mclmax:
# Describe the PR

This PR introduces a small change to how the z-indices of active modals are calculated.

Currently there are cases ([issue 1516](https://github.com/bootstrap-vue-next/bootstrap-vue-next/issues/1516)) where a newly opened modal will open behind an active modal.

The change to behaviour here is to calculate the z-index of a modal based on it's position in the stack of active modals. All modals have a z-index lower than or equal to 1055. The newer (high up on the stack) the modal is the close to 1055. The latest having a z-index of 1055. Inactive modals have a z-index of 1055 by default, the highest value. Meaning when they open they are on top by default until the z-index is adjusted.

There is a slightly different approach I tried which was just to calculate the z-index as `1055 + indexInStack` however, this meant that for a small time window the new modal would be behind any active modal until the new z-index was calculated. This meant that the modal wouldn't fade in smoothly and instead would pop in quite abruptly. 

## Small replication

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

### DIFF
--- a/packages/bootstrap-vue-next/src/components/BModal.vue
+++ b/packages/bootstrap-vue-next/src/components/BModal.vue
@@ -19,6 +19,7 @@
         :aria-describedby="`${computedId}-body`"
         tabindex="-1"
         v-bind="$attrs"
+        :style="computedZIndex"
       >
         <div class="modal-dialog" :class="modalDialogClasses">
           <div v-if="lazyShowing" class="modal-content" :class="contentClass">
@@ -100,7 +101,15 @@
 </template>
 
 <script setup lang="ts">
-import {computed, reactive, ref, type RendererElement, toRef, useSlots} from 'vue'
+import {
+  computed,
+  type CSSProperties,
+  reactive,
+  ref,
+  type RendererElement,
+  toRef,
+  useSlots,
+} from 'vue'
 import {
   useBackgroundVariant,
   useBooleanish,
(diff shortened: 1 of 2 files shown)


packages/bootstrap-vue-next/src/composables/useModalManager.ts
```diff
@@ -62,4 +72,9 @@ export default (modalOpen: Ref<boolean>): void => {
     },
     {immediate: true}
   )
+
+  return {
+    activePosition: computed(() => stack.value.findIndex((el) => el.exposed?.id === toValue(id))),
+    activeModalCount: computed(() => stack.value.length),
```
VividLemon, line 78:
```suggestion
    activeModalCount: toRef(() => stack.value.length),
```

packages/bootstrap-vue-next/src/components/BModal.vue
```diff
@@ -506,7 +515,17 @@ const onAfterLeave = () => {
...
-useModalManager(isActive)
+const {activePosition, activeModalCount} = useModalManager(isActive, computedId)
+const defaultModalDialogZIndex = 1056
+const computedZIndex = computed<CSSProperties>(() => ({
+  // Make sure that newly opened modals have a higher z-index than currently active ones.
+  // All active modals have a z-index of ('defaultZIndex' - 'stackSize' - 'positionInStack').
+  //
+  // This means inactive modals will already be higher than active ones when opened.
+  'z-index': isActive.value
+    ? defaultModalDialogZIndex - (activeModalCount.value - activePosition.value)
```
VividLemon, line 526: What affect does having `(activeModalCount.value` in here have

VividLemon (changes requested):

packages/bootstrap-vue-next/src/components/BModal.vue
mclmax: it's used to flip the "priority" of the modals on the stack. so if we have [Modal1, Modal2, Modal3] (0, 1, 2), we want Modal3 to have the highest z-index, Modal2 second highest etc. without the count we would have [1056, 1055, 1054] which isn't what we want. subtracting stack size first gives: [1053, 1054, 1055].

mclmax: there is the risk that things could get slightly weird if folks are opening hundreds of modals i guess

packages/bootstrap-vue-next/src/composables/useModalManager.ts
mclmax: made the change

PR #1525, opened 2023-10-18
fix(BTabs): tabs using provide/inject 

xvaara:
# Describe the PR
fix(BTabs): tabs using provide/inject 
fix(BTabs): missing aria prop
feat(BTabs): keyboard navigation
feat(BTabs): v-model:active-id to monitor/control the tab.

## Small replication

## PR checklist

- [x] Feature - `feat(...)`
- [ ] Other (please describe)


packages/bootstrap-vue-next/src/types/ComponentProps.ts
```diff
@@ -228,3 +229,42 @@ export interface BPopoverProps {
   html?: Booleanish
   noninteractive?: Booleanish
 }
+
+export interface BTabsProps {
```
VividLemon, line 233: Export of component props is reserved for components that share props with another. Like how multiple components define and use BLinkProps. It's not simply for moving code from one place to another

packages/bootstrap-vue-next/src/components/BTabs/BTab.vue
```diff
@@ -1,60 +1,70 @@
...
-defineSlots<{
+const slots = defineSlots<{
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   default?: (props: Record<string, never>) => any
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  title?: (props: Record<string, never>) => any
+}>()
+
+const emit = defineEmits<{
+  'update:active': [value: boolean]
```
VividLemon, line 58: I think the addition of this prop & emit isn't the best solution to the problem that it's trying to fix. It doesn't quite feel right that the child has some responsibility over updating its own active state.

packages/bootstrap-vue-next/src/components/BTabs/BTab.vue
```diff
@@ -63,27 +73,85 @@ const lazyBoolean = useBooleanish(
 )
 
 const lazyRenderCompleted = ref(false)
+const el = ref<HTMLElement | null>(null)
+
+const {onClick, ...attrs} = useAttrs()
+
+const tab = toRef(
```
VividLemon, line 80: toRef doesn't really work for non-primitive values. This should be a computed.

Essentially, if you're returning a non-primitive, doing array logic (filter, map, etc) (or similarly object manipulation), or the logic is fairly complex (use judgement), then you will want to use computed.

packages/bootstrap-vue-next/src/components/BTabs/BTab.vue
```diff
@@ -63,27 +73,85 @@ const lazyBoolean = useBooleanish(
...
+      titleLinkClass: props.titleLinkClass,
+      onClick,
+      el: el.value,
+    }) as TabType
+)
+
+// parentData?.registerTab(tab)
+onMounted(() => {
+  if (!parentData) return
+  parentData?.registerTab(tab)
```
VividLemon, line 99: Right above this line you do !parentData check, but still optional chain. `onUnmounted` also contains this

packages/bootstrap-vue-next/src/components/BTabs/BTab.vue
```diff
@@ -63,27 +73,85 @@ const lazyBoolean = useBooleanish(
...
-const computedActive = toRef(() => activeBoolean.value && !disabledBoolean.value)
+const computedActive = toRef(() => isActive.value && !disabledBoolean.value)
 const showSlot = computed(() => {
   const hasLazyRenderedOnce =
     computedLazy.value && computedLazyOnce.value && lazyRenderCompleted.value
   return computedActive.value || !computedLazy.value || hasLazyRenderedOnce
 })
-const show = ref(activeBoolean.value)
+
+// const show = ref(activeBoolean.value)
```
VividLemon, line 123:
```suggestion
```

packages/bootstrap-vue-next/src/components/BTabs/BTab.vue
```diff
@@ -63,27 +73,85 @@ const lazyBoolean = useBooleanish(
...
+      titleComponent: slots.title,
+      titleItemClass: props.titleItemClass,
+      titleLinkAttributes: props.titleLinkAttributes,
+      titleLinkClass: props.titleLinkClass,
+      onClick,
+      el: el.value,
+    }) as TabType
+)
+
+// parentData?.registerTab(tab)
```
VividLemon, line 96:
```suggestion
```

packages/bootstrap-vue-next/src/components/BTabs/BTab.vue
```diff
@@ -63,27 +73,85 @@ const lazyBoolean = useBooleanish(
...
+const computedActive = toRef(() => isActive.value && !disabledBoolean.value)
 const showSlot = computed(() => {
   const hasLazyRenderedOnce =
     computedLazy.value && computedLazyOnce.value && lazyRenderCompleted.value
   return computedActive.value || !computedLazy.value || hasLazyRenderedOnce
 })
-const show = ref(activeBoolean.value)
+
+// const show = ref(activeBoolean.value)
+watch(isActive, (active) => {
```
VividLemon, line 124: Like stated before, not a huge fan of all of this. Seems like there's some alternative here.

packages/bootstrap-vue-next/src/components/BTabs/BTab.vue
```diff
@@ -63,27 +73,85 @@ const lazyBoolean = useBooleanish(
...
+  if (active) {
+    emit('update:active', true)
+    setTimeout(() => {
+      show.value = true
+    }, 0)
+  } else {
+    show.value = false
+    emit('update:active', false)
+  }
```
VividLemon, line 133:
```suggestion
    }, 0)
    return
  }
  show.value = false
  emit('update:active', false)
```

packages/bootstrap-vue-next/src/components/BTabs/BTabs.vue
```diff
@@ -80,61 +67,39 @@
...
+  small: false,
+  tag: 'div',
+  vertical: false,
+  modelValue: -1,
+  activeId: undefined,
+})
 
 const emit = defineEmits<{
   'update:modelValue': [value: number]
+  'update:activeId': [value: string]
```
VividLemon, line 102: What exactly does this prop/emit seek to achieve?

packages/bootstrap-vue-next/src/utils/keys.ts
```diff
@@ -12,6 +19,12 @@ export const carouselInjectionKey: InjectionKey<{
 export const tabsInjectionKey: InjectionKey<{
   lazy: Readonly<Ref<boolean>>
   card: Readonly<Ref<boolean>>
+  noFade: Readonly<Ref<boolean>>
+  registerTab: (tab: Readonly<Ref<TabType>>) => void
+  unregisterTab: (id: string) => void
+  activateTab: (id: string | undefined) => void
+  activeTabClass: ClassValue
```
VividLemon, line 26:
```suggestion
  activeTabClass: Readonly<Ref<ClassValue>>
```

packages/bootstrap-vue-next/src/components/BTabs/BTabs.vue
```diff
@@ -302,75 +198,108 @@ const handleClick = (event: MouseEvent, index: number) => {
...
+    return
+  }
+  activateTab(tabs.value.map((tab) => !tab.disabled).indexOf(true))
+}
 
 provide(tabsInjectionKey, {
   lazy: lazyBoolean,
   card: cardBoolean,
+  noFade: noFadeBoolean,
+  activeTabClass: props.activeTabClass,
```
VividLemon, line 295:
```suggestion
  activeTabClass: toRef(() => props.activeTabClass),
```

packages/bootstrap-vue-next/src/components/BTabs/tabs.spec.ts
```diff
@@ -499,4 +499,73 @@ describe('tabs', () => {
 
     expect(wrapper.findComponent({name: 'b-tab'}).find('p').text()).toBe('bar')
   })
+  it('renders in complex structure and is reactive ', async () => {
```
VividLemon, line 502: This single test seems like it can be broken up into multiple independent tests.

packages/bootstrap-vue-next/src/components/BTabs/BTabs.vue
```diff
@@ -302,75 +198,108 @@ const handleClick = (event: MouseEvent, index: number) => {
...
+  registerTab,
+  unregisterTab,
+  activeId,
+  activateTab: (id) => {
+    const idx = tabs.value.findIndex((t) => t.id === id)
+    if (id === undefined || idx < 0) activateTab(nextIndex(0, 1))
+    else activateTab(idx)
```
VividLemon, line 302:
```suggestion
    if (id === undefined || idx < 0) {
      activateTab(nextIndex(0, 1))
      return
    }
    activateTab(idx)
```

packages/bootstrap-vue-next/src/components/BTabs/BTabs.vue
```diff
@@ -302,75 +198,108 @@ const handleClick = (event: MouseEvent, index: number) => {
...
-    activateTab(0)
+watch(activeId, (newValue, oldValue) => {
+  const index = tabs.value.findIndex((t) => t.id === newValue)
+  if (newValue === oldValue) return
+  if (tabs.value.length <= 0) {
     return
   }
-  if (index >= tabs.value.length) {
-    activateTab(tabs.value.length - 1)
+  if (index < 0) {
```
VividLemon, line 242:
```suggestion
  if (index === -1) {
```

packages/bootstrap-vue-next/src/components/BTabs/BTabs.vue
```diff
@@ -302,75 +198,108 @@ const handleClick = (event: MouseEvent, index: number) => {
...
   lazy: lazyBoolean,
   card: cardBoolean,
+  noFade: noFadeBoolean,
+  activeTabClass: props.activeTabClass,
+  registerTab,
+  unregisterTab,
+  activeId,
+  activateTab: (id) => {
+    const idx = tabs.value.findIndex((t) => t.id === id)
+    if (id === undefined || idx < 0) activateTab(nextIndex(0, 1))
```
VividLemon, line 301:
```suggestion
    if (id === undefined || idx === -1) activateTab(nextIndex(0, 1))
```

VividLemon (changes requested):
I'd prefer if this wasn't merged until I finish my laundry list of changes. I have like a 75 file pr in progress and would like to avoid the conflict.

packages/bootstrap-vue-next/src/types/ComponentProps.ts
xvaara: Ok, I'll move them back.

packages/bootstrap-vue-next/src/components/BTabs/BTab.vue
xvaara: The prop was there already, and the emit is there just to update the prop. They aren't the main logic to switch the tab. It's just another way to control active tab. It's used in  `<b-tab active> ` or `<b-tab v-model:active="foo">` if you want to control/check the activity state of a singular tab. They always go to the parent to actually do the work.

packages/bootstrap-vue-next/src/components/BTabs/BTab.vue
xvaara: Ok, I'm not so clear on when to use toRef and when toComputed, the difference isn't that clear in the docs for me. Mane meed to read it again.

packages/bootstrap-vue-next/src/components/BTabs/BTab.vue
xvaara: I blame eslint for that :)

packages/bootstrap-vue-next/src/components/BTabs/BTab.vue
xvaara: I'm not really sure what you mean here. the tab has to have a way of getting the active state from parent. This is the whole point of this refactor. I'm at a little loss here.

packages/bootstrap-vue-next/src/components/BTabs/BTabs.vue
xvaara: check TTabs.vue. It's another way to control which tab is open, using the id of the tab, instead of index.

packages/bootstrap-vue-next/src/components/BTabs/BTabs.vue
xvaara: Why do you dislike if/else? To me it's easier to read.

packages/bootstrap-vue-next/src/components/BTabs/BTab.vue
VividLemon: It's likely not something you'd find in the docs.

toRef has many uses. The main point of its changes in Vue 3.3 was to normalize getter functions, allowing composables to accept `MaybeRefOrGetter`. If you remember, the only way to pass in reactive variables to a composable was doing something like `useBooleanish(toRef(props, 'someProp'))`. This was pretty inefficient as it creates an unnecessary intermediate ref. With the changes, you can do `MaybeRefOrGetter` to condense the code to `useBooleanish(() => props.someProp)`. 

Then on the composable side you have two options of normalizing these values, toRef, or toValue. toValue is generally used more in cases where you have computed values, like how the actual `useBooleanish` function implements it. There are other cases where toRef is more appropriate. Like in `useCountdown`

This is all a bit basic stuff, it's all history at this point on why toRef was expanded. Subsequently the same code that allows a composable to accept a getter function, means that toRef can accept a getter function argument. This has major implications as it means that a toRef can essentially behave very similar to how computed works.

`computed(() => someValue ?? someOtherValue)` can also be represented as `toRef(() => someValue ?? someOtherValue)`, since toRef is now able to accept getter functions, you can do this.

So it begs the question on, _why_? computed functions carry overhead in their _caching system_. When the caching system was developed, the Vue core team likely thought that "The performance gain from caching outweighs all disadvantages". However, in practice this wasn't always true. That being said, the major difference between the two functions is that `computed` caches its value and `toRef` does not.

But as it turned out, "always caching" didn't always yield the best performance, and instead always retriggering the getter function can sometimes be more performant as it doesn't have the same overhead. So, simple quick logic can be somewhat faster to just redo each time than setting/getting the cached value.

So this is the major difference. But, there are some caveats as I had mentioned before. toRef doesn't work well when you return a non-primitive (the most common being arrays and objects). Why exactly this is, idk. So, if you ever need to return an object or array off of the getter function, toRef simply wont work. `toRef(() => [])`, `toRef(() => ({}))`. Not property simple property access, those two are different.

In addition to this, there are situations where caching is a major benefit. Say if you were doing Array.filter, every subsequent rerender would cause the filter of the array, a much more expansive computation comparatively speaking. So, if you are doing things like Array.filter, map, etc, these will benefit from computed caching.

Then, in the same vein, expansive computations of any kind are going to benefit from caching. I can't quite come up with an example, but just things that take time. So, these things should use computed.

Another example of when to use computed is when you have long functions. Generally, I won't ever do something like
```ts
toRef(() => {
// blah blah blah
})
```
If I can't represent a toRef statement in one line, then it's not worth. `toRef(() => )`. Some of these may benefit from caching, but I'm fine with that. Just keep things one line if you're using toRef, but that's just my opinion. 

tldr, toRef allowing getter function access means that the lines between computed and toRef are blurred, where only under certain caveats will you want to choose one over the other... However, with all that being said, the _true affects_ are very, very low. From an application perspective, you are likely not ever going to see much of a performance difference. However, from a _library perspective_, I will take what I can get. In mind mind, a 2% increase in performance for 1k applications, and say if each of those 1k apps gets 10 daily users, then when you put that into perspective, the global shared gain across all of those people is worth it. But again, this isn't really something to put too much time investment over thinking "when to use this over that". Once you know the caveats, it's pretty much gut feeling.

xvaara:
> I'd prefer if this wasn't merged until I finish my laundry list of changes. I have like a 75 file pr in progress and would like to avoid the conflict.

@VividLemon thanks for the review. 
I'll wait and rebase when you're done. Sounds like a huge pr :smile:

packages/bootstrap-vue-next/src/components/BTabs/BTabs.vue
VividLemon: https://www.youtube.com/watch?v=CFRhGnuXG-4&t=270s

VividLemon:
> @VividLemon thanks for the review. I'll wait and rebase when you're done. Sounds like a huge pr 😄

The changes I made are pushed. Though, it is not functional.

VividLemon:
@xvaara Am I correct that this closes the linked 3 issues?

xvaara:
> @xvaara Am I correct that this closes the linked 3 issues?

@VividLemon yes, this closes them also.

VividLemon (approved):

PR #1527, opened 2023-10-19
fixed BOffcanvas and BModal hide event

### LINKED ISSUE #1526: BOffcanvas hide event when no needed
### Describe the bug

Even with ```no-close-on-esc``` or ```no-close-on-backdrop``` attributes BOffcanvas fires ```hide``` event when pressing ESC or clicking on backdrop

### Reproduction

https://stackblitz.com/edit/github-bzyfdw?file=src%2Fmain.js,src%2FApp.vue

### System Info

```Shell
latest bootstrap-vue-next
```

### Used Package Manager

pnpm

### Validations

makroxyz:
# Describe the PR

fixes #1526

### DIFF
--- a/packages/bootstrap-vue-next/src/components/BModal.vue
+++ b/packages/bootstrap-vue-next/src/components/BModal.vue
@@ -471,6 +471,7 @@ const hide = (trigger = '') => {
   if (trigger === 'close') {
     emit(trigger, event)
   }
+    emit('hide', event)
   if (event.defaultPrevented) {
     emit('hide-prevented')
     return
--- a/packages/bootstrap-vue-next/src/components/BModal.vue
+++ b/packages/bootstrap-vue-next/src/components/BModal.vue
@@ -471,12 +471,12 @@ const hide = (trigger = '') => {
   if (trigger === 'close') {
     emit(trigger, event)
   }
-    emit('hide', event)
+  emit('hide', event)
+
   if (event.defaultPrevented) {
     emit('hide-prevented')
     return
   }
-
 
   modelValue.value = false
 }
--- a/packages/bootstrap-vue-next/src/components/BModal.vue
+++ b/packages/bootstrap-vue-next/src/components/BModal.vue
@@ -454,6 +454,14 @@ const buildTriggerableEvent = (
 const hide = (trigger = '') => {
   const event = buildTriggerableEvent('hide', {cancelable: trigger !== '', trigger})
 
+  if (
+    (trigger === 'backdrop' && noCloseOnBackdropBoolean.value) ||
+    (trigger === 'esc' && noCloseOnEscBoolean.value)
+  ) {
+    emit('hide-prevented')
+    return
+  }
+
   if (trigger === 'ok') {
     emit(trigger, event)
   }
@@ -463,16 +471,13 @@ const hide = (trigger = '') => {
   if (trigger === 'close') {
     emit(trigger, event)
   }
-  emit('hide', event)
-
-  if (
-    event.defaultPrevented ||
-    (trigger === 'backdrop' && noCloseOnBackdropBoolean.value) ||
-    (trigger === 'esc' && noCloseOnEscBoolean.value)
-  ) {
+  if (event.defaultPrevented) {
     emit('hide-prevented')
     return
   }
+
+  emit('hide', event)
+
   modelValue.value = false
 }
 
--- a/packages/bootstrap-vue-next/src/components/BModal.vue
+++ b/packages/bootstrap-vue-next/src/components/BModal.vue
@@ -452,8 +452,6 @@ const buildTriggerableEvent = (
   })
 
 const hide = (trigger = '') => {
-  const event = buildTriggerableEvent('hide', {cancelable: trigger !== '', trigger})
-
   if (
     (trigger === 'backdrop' && noCloseOnBackdropBoolean.value) ||
     (trigger === 'esc' && noCloseOnEscBoolean.value)
@@ -461,6 +459,8 @@ const hide = (trigger = '') => {
     emit('hide-prevented')
     return
   }
+
+  const event = buildTriggerableEvent('hide', {cancelable: trigger !== '', trigger})
 
   if (trigger === 'ok') {
     emit(trigger, event)
--- a/packages/bootstrap-vue-next/src/components/BModal.vue
+++ b/packages/bootstrap-vue-next/src/components/BModal.vue
@@ -477,7 +477,6 @@ const hide = (trigger = '') => {
     return
   }
 
-  emit('hide', event)
 
   modelValue.value = false
 }
--- a/packages/bootstrap-vue-next/src/components/BOffcanvas/BOffcanvas.vue
+++ b/packages/bootstrap-vue-next/src/components/BOffcanvas/BOffcanvas.vue
@@ -269,7 +269,6 @@ const hide = (trigger = '') => {
   const event = buildTriggerableEvent('hide', {cancelable: trigger !== '', trigger})
 
   if (
-    event.defaultPrevented ||
     (trigger === 'backdrop' && noCloseOnBackdropBoolean.value) ||
     (trigger === 'esc' && noCloseOnEscBoolean.value)
   ) {
@@ -285,10 +284,11 @@ const hide = (trigger = '') => {
   }
   emit('hide', event)
 
-if(event.defaultPrevented) {
+  if (event.defaultPrevented) {
     emit('hide-prevented')
     return
-}
+  }
+
   modelValue.value = false
 }
 
--- a/packages/bootstrap-vue-next/src/components/BOffcanvas/BOffcanvas.vue
+++ b/packages/bootstrap-vue-next/src/components/BOffcanvas/BOffcanvas.vue
@@ -266,8 +266,6 @@ const buildTriggerableEvent = (
   })
 
 const hide = (trigger = '') => {
-  const event = buildTriggerableEvent('hide', {cancelable: trigger !== '', trigger})
-
   if (
     (trigger === 'backdrop' && noCloseOnBackdropBoolean.value) ||
     (trigger === 'esc' && noCloseOnEscBoolean.value)
@@ -275,6 +273,8 @@ const hide = (trigger = '') => {
     emit('hide-prevented')
     return
   }
+
+  const event = buildTriggerableEvent('hide', {cancelable: trigger !== '', trigger})
 
   if (trigger === 'close') {
     emit(trigger, event)
--- a/packages/bootstrap-vue-next/src/components/BOffcanvas/BOffcanvas.vue
+++ b/packages/bootstrap-vue-next/src/components/BOffcanvas/BOffcanvas.vue
@@ -285,6 +285,10 @@ const hide = (trigger = '') => {
   }
   emit('hide', event)
 
+if(event.defaultPrevented) {
+    emit('hide-prevented')
+    return
+}
   modelValue.value = false
 }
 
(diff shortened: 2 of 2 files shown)


VividLemon:
```ts
const hide = (trigger = '') => {
  const event = buildTriggerableEvent('hide', {cancelable: trigger !== '', trigger})

  if (
    event.defaultPrevented ||
    (trigger === 'backdrop' && noCloseOnBackdropBoolean.value) ||
    (trigger === 'esc' && noCloseOnEscBoolean.value)
  ) {
    emit('hide-prevented', event)
    return
  }

  if (trigger === 'close') {
    emit(trigger, event)
  }
  if (trigger === 'esc') {
    emit(trigger, event)
  }
  emit('hide', event)

  if (event.defaultPrevented) {
    emit('hide-prevented', event)
    return
  }

  modelValue.value = false
}

const show = () => {
  const event = buildTriggerableEvent('show', {cancelable: true})
  emit('show', event)
  if (event.defaultPrevented) {
    modelValue.value = false
    emit('show-prevented', event)
    return
  }
  modelValue.value = true
}
```

Other files suffer from a similar issue.

makroxyz:
Sorry I don't understand if this is expected or we need to fix other files too

VividLemon:
> Sorry I don't understand if this is expected or we need to fix other files too

The codeblock I gave is likely the solution you should update to. In addition, multiple other files should be fixed with that fix as well.

makroxyz:
Ok, I'm not on PC now so I thought was code from my PR I also "switched"
the if sequence.
Tomorrow I will check for other components

Il gio 19 ott 2023, 16:48 Issayah ***@***.***> ha scritto:

makroxyz:
Hi I found also BModal  and fixed it.

makroxyz:
@VividLemon 

I think this is better and modal.spec.ts doesn't fail
```js
const hide = (trigger = '') => {
  const event = buildTriggerableEvent('hide', {cancelable: trigger !== '', trigger})

  if (trigger === 'ok') {
    emit(trigger, event)
  }
  if (trigger === 'cancel') {
    emit(trigger, event)
  }
  if (trigger === 'close') {
    emit(trigger, event)
  }

  if (
    event.defaultPrevented ||
    (trigger === 'backdrop' && noCloseOnBackdropBoolean.value) ||
    (trigger === 'esc' && noCloseOnEscBoolean.value)
  ) {
    emit('hide-prevented')
    return
  }

  emit('hide', event)

  modelValue.value = false
}
```

if I press close button  and I want to prevent modal close (like in test), I emit ```close``` event then close event doesn't  fire ```hide``` if prevented or ```no-close-on-backdrop```/```no-close-on-esc``` are true

the same for BOffcanvas
```js
const buildTriggerableEvent = (
  type: string,
  opts: Partial<BvTriggerableEvent> = {}
): BvTriggerableEvent =>
  new BvTriggerableEvent(type, {
    cancelable: false,
    target: element.value || null,
    relatedTarget: null,
    trigger: null,
    ...opts,
    componentId: computedId.value,
  })

const hide = (trigger = '') => {
  const event = buildTriggerableEvent('hide', {cancelable: trigger !== '', trigger})

  if (trigger === 'close') {
    emit(trigger, event)
  }
  if (trigger === 'esc') {
    emit(trigger, event)
  }

  if (
    event.defaultPrevented ||
    (trigger === 'backdrop' && noCloseOnBackdropBoolean.value) ||
    (trigger === 'esc' && noCloseOnEscBoolean.value)
  ) {
    emit('hide-prevented')
    return
  }

  emit('hide', event)

  modelValue.value = false
}
``` 
I press ESC event ```esc``` should be fired but it doesn't fire ```hide``` if prevented  or ```no-close-on-backdrop```/```no-close-on-esc``` are true

VividLemon:
> I think this is better and modal.spec.ts doesn't fail

Then fix the tests. A failing test is only an issue if the expected behavior is not correct. (Too many double negatives? If something was not correct when the test was written, then it's just a bad test!)

> if I press close button and I want to prevent modal close (like in test), I emit close event then close event doesn't fire hide if prevented or no-close-on-backdrop/no-close-on-esc are true

Let's get two things straightened out here. First off, in my code example you're able to prevent modal closing in all cases. Second off, you're misunderstanding the purpose of the hide event. It is not to indicate that the modal has been hidden, that's what the hidden event is for. Functionally it is in line with https://getbootstrap.com/docs/5.3/components/modal/#events , though our system utilizes a more extra events for specific targets. 

To add to this, the hide event also emits the Event, which allows it to be preventable. Those events are grouped up because they all signify the same point in time for the event... Assuming that makes sense. 

So, my code is in-line with what the spec is for Bootstrap v5 and how I would prefer it to be. Unless you can tell me if and how it is functionally not in compliance with the Bootstrap v5, then that is how I would prefer it to be fixed. 

Though hypothetically the 
```ts
if (
    event.defaultPrevented ||
    (trigger === 'backdrop' && noCloseOnBackdropBoolean.value) ||
    (trigger === 'esc' && noCloseOnEscBoolean.value)
  ) {
    emit('hide-prevented', event)
    return
  }
```
`event.defaultPrevented` for the first check is redundant, as you couldn't have preventedDefault yet. 

Functionally the movement of event emitting like `close` `ok`, etc is minor, but when you put it into the context of it being the same as the hide event, then I think it makes more sense. But you need to know what the purpose of the hide event is for above all else.

packages/bootstrap-vue-next/src/components/BOffcanvas/BOffcanvas.vue
```diff
@@ -284,6 +276,15 @@ const hide = (trigger = '') => {
...
   }
+
+  if (trigger === 'close') {
+    emit(trigger, event)
+  }
+  if (trigger === 'esc') {
+    emit(trigger, event)
+  }
+  emit('hide', event)
+
```
VividLemon, line 287:
```suggestion

if(event.defaultPrevented) {
    emit('hide-prevented')
    return
}
```

We need to ensure that you can prevent the closing when this event is triggered

packages/bootstrap-vue-next/src/components/BOffcanvas/BOffcanvas.vue
```diff
@@ -268,14 +268,6 @@ const buildTriggerableEvent = (
...
-  if (trigger === 'close') {
-    emit(trigger, event)
-  }
-  if (trigger === 'esc') {
-    emit(trigger, event)
-  }
-  emit('hide', event)
-
   if (
     event.defaultPrevented ||
```
VividLemon, line 272:
```suggestion
```
You can't prevent default at this point, so this check is unnecessary

VividLemon: There might be some linting stuff to look at with this change

packages/bootstrap-vue-next/src/components/BModal.vue
```diff
@@ -463,16 +471,13 @@ const hide = (trigger = '') => {
   if (trigger === 'close') {
     emit(trigger, event)
   }
```
VividLemon, line 473:
```suggestion
  }
    emit('hide', event)
```

VividLemon: I think the suggestion contains a linting tab issue

packages/bootstrap-vue-next/src/components/BModal.vue
```diff
@@ -463,16 +471,13 @@ const hide = (trigger = '') => {
...
-    event.defaultPrevented ||
-    (trigger === 'backdrop' && noCloseOnBackdropBoolean.value) ||
-    (trigger === 'esc' && noCloseOnEscBoolean.value)
-  ) {
+  if (event.defaultPrevented) {
     emit('hide-prevented')
     return
   }
+
+  emit('hide', event)
```
VividLemon, line 479:
```suggestion
```
This is moved up with the others, (close, ok, etc)

VividLemon (changes requested):

makroxyz:
Hi I commited changes but I'm not sure this is right  in both files
```js
 emit('hide', event)

  if (event.defaultPrevented) {
    emit('hide-prevented')
    return
  }
```
I emit hide before checking if it's prevented...

VividLemon:
> Hi I commited changes but I'm not sure this is right in both files
> 
> ```js
>  emit('hide', event)
> 
>   if (event.defaultPrevented) {
>     emit('hide-prevented')
>     return
>   }
> ```
> 
> I emit hide before checking if it's prevented...

Yes, that's the point. it's so you can do things like 
```html
<BModal @hide.prevent @hide-prevented="consoleLog('foo')" />
```

makroxyz:
@VividLemon 
I'm facing this issues (is it really an issue?)

When a BModal is conditioned by v-if, ```show/shown``` events are not fired. Old Bootstrap-vue does it.
I think because event are fired on BTransition tag.

```js
<template>
  <BContainer>
    <BRow>
      <BCol> 
        <BModal v-if="show" v-model="show" @show="() => console.log('show')">
          test
        </BModal>
        <BButton @click="show = true">Test</BButton>
      </BCol>
    </BRow>
  </BContainer>
</template>

<script setup lang="ts">
import { ref } from 'vue';
const show = ref(false)
</script>
````

VividLemon:
> @VividLemon I'm facing this issue
> 
> When a BModal is conditioned by v-if, `show/shown` events are not fired. Old Bootstrap-vue does it. I think because event are fired on BTransition tag.
> 
> ```js
> <template>
>   <BContainer>
>     <BRow>
>       <BCol> 
>         <BModal v-if="show" v-model="show" @show="() => console.log('show')">
>           test
>         </BModal>
>         <BButton @click="show = true">Test</BButton>
>       </BCol>
>     </BRow>
>   </BContainer>
> </template>
> 
> <script setup lang="ts">
> import { ref } from 'vue';
> const show = ref(false)
> </script>
> ```
> 
> Update: the same when a modal is used in dynamic component
> 
> ```
> <component :is="modalComponent" v-show="show" />   // fires show* event only on SECOND open since modalComponent can be undefined or null
> ```
> 
> https://vuejs.org/guide/built-ins/transition.html#the-transition-component
> 
> dynamic component or v-if component should be inside transition

This would be weird for a person to do. But regardless, I'm not sure how to fix it. Unlike attributes, directives don't fall through to $attrs. So there is no way for the component to know if a directive is attached to it (that I know of)

VividLemon (changes requested):
This is the only remaining feedback I have.

makroxyz:
> > @VividLemon I'm facing this issue
> > When a BModal is conditioned by v-if, `show/shown` events are not fired. Old Bootstrap-vue does it. I think because event are fired on BTransition tag.
> > ```js
> > <template>
> >   <BContainer>
> >     <BRow>
> >       <BCol> 
> >         <BModal v-if="show" v-model="show" @show="() => console.log('show')">
> >           test
> >         </BModal>
> >         <BButton @click="show = true">Test</BButton>
> >       </BCol>
> >     </BRow>
> >   </BContainer>
> > </template>
> > 
> > <script setup lang="ts">
> > import { ref } from 'vue';
> > const show = ref(false)
> > </script>
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Update: the same when a modal is used in dynamic component
> > ```
> > <component :is="modalComponent" v-show="show" />   // fires show* event only on SECOND open since modalComponent can be undefined or null
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > https://vuejs.org/guide/built-ins/transition.html#the-transition-component
> > dynamic component or v-if component should be inside transition
> 
> This would be weird for a person to do. But regardless, I'm not sure how to fix it. Unlike attributes, directives don't fall through to $attrs. So there is no way for the component to know if a directive is attached to it (that I know of)

That's right. But this could be a problem... I don't have any idea on how to solve it. While a v-if on a BModal should be "managed" by ```lazy``` attribute, using dynamic component isn't a bad practice and can be an issue

VividLemon:
> That's right. But this could be a problem... I don't have any idea on how to solve it. While a v-if on a BModal should be "managed" by `lazy` attribute, using dynamic component isn't a bad practice and can be an issue

You should make an issue ticket for this

PR #1537, opened 2023-10-24
feat(BAccordion): add class props for elements

### LINKED ISSUE #1535: Adding classes on BAccordion

Hi there, 

First of all, thanks a lot for maintaining this repository, it definitely makes it much more pleasant to use on vue/nuxt instead of the standard bootstrap. 👍 

My new feature proposal is related to the accordion component:

The BAccordion component doesn't allow adding custom classes to the generated buttons that toggle the collapse nor the parent element or the "content" part of the accordion.

It appears as bcollapse has a hardcoded "accordion-collapse" and each collapse button has "accordion-button" and "collapsed" classes only.

It would be useful to have some props so we could add some classes to either of them.

### Suggested solution

I would suggest some props that allow classes to be added to the parent accordion as well as each collapse button and collapse content. Something simillar to how the BTab(s) component has their "contentClass", "navClass", etc props.

### Alternative

_No response_

### Additional context

_No response_

### Validations

- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

xvaara:
# Describe the PR
feat(BAccordion): add class props for elements
closes #1535 

## Small replication

```html
<BAccordion class="p-1 border border-primary m-1">
  <BAccordionItem
    title="Accordion #1"
    class="p-1 border border-danger m-1"
    :header-attrs="{class: 'p-1 border border-success m-1', onClick: log}"
    :button-attrs="{class: 'bg-success', onClick: log, style: 'cursor: move;'}"
    :collapse-attrs="{class: 'p-1 border border-primary m-1', onClick: log}"
    :body-attrs="{class: 'bg-primary', onClick: log}"
    toggle
  >
    1
  </BAccordionItem>
  <BAccordionItem title="#2"> 2 </BAccordionItem>
  <BAccordionItem title="#3"> 3 </BAccordionItem>
</BAccordion>
```

## PR checklist

- [x] Feature - `feat(...)`
- [ ] Other (please describe)


packages/bootstrap-vue-next/src/components/BAccordion/BAccordionItem.vue
```diff
@@ -32,24 +38,25 @@
...
 import BCollapse from '../BCollapse.vue'
 import {accordionInjectionKey, BvTriggerableEvent} from '../../utils'
 import {useId} from '../../composables'
-import type {Booleanish} from '../../types'
+import type {Booleanish, ClassValue} from '../../types'
 
 defineOptions({
   inheritAttrs: false,
 })
+const {class: wrapperClass, ...attrs} = useAttrs()
```
VividLemon, line 59: wrapperClass could just be a prop and used in the same way. There's no reason to treat it as an attr or do anything special with useAttrs

packages/bootstrap-vue-next/src/components/BAccordion/BAccordionItem.vue
```diff
@@ -62,6 +69,10 @@ const props = withDefaults(
     horizontal?: Booleanish
     visible?: Booleanish
     isNav?: Booleanish
+    headerClass?: ClassValue
+    buttonClass?: ClassValue
+    collapseClass?: ClassValue
+    bodyClass?: ClassValue
```
VividLemon, line 75: You should also do bodyAttrs, collapseAttrs, etc to satisfy https://github.com/orgs/bootstrap-vue-next/projects/1?pane=issue&itemId=31679414

packages/bootstrap-vue-next/src/components/BAccordion/BAccordionItem.vue
```diff
@@ -19,10 +20,15 @@
...
+        <component
+          :is="headerTag"
+          :id="`${computedId}-heading`"
+          class="accordion-header"
+          :class="headerClass"
+        >
           <button
             class="accordion-button"
-            :class="{collapsed: !toggleVisible}"
+            :class="{collapsed: !toggleVisible, [buttonClass]: !!buttonClass}"
```
VividLemon, line 31:
```suggestion
            :class="[buttonClass, {collapsed: !toggleVisible}]"
```

VividLemon: Pretty sure that works the same. But smaller

packages/bootstrap-vue-next/src/components/BAccordion/BAccordionItem.vue
```diff
@@ -1,10 +1,11 @@
 <template>
-  <div class="accordion-item">
+  <div class="accordion-item" :class="wrapperClass">
     <BCollapse
       :id="computedId"
       v-model="modelValue"
       class="accordion-collapse"
-      v-bind="$attrs"
+      :class="collapseClass"
+      v-bind="attrs"
```
VividLemon, line 8: this v-bind should probably be near the buttom. If someone were to do pass through attrs.aria-labelledby, it would always be overwritten.

VividLemon (changes requested):

packages/bootstrap-vue-next/src/components/BAccordion/BAccordionItem.vue
VividLemon: In this case, it would make more sense for the attrs to all go on one element. Rather than randomly attaching to different elements.

packages/bootstrap-vue-next/src/components/BAccordion/BAccordionItem.vue
xvaara: if we have bodyAttrs is bodyClass even needed? bodyAttrs="{class: 'bg-danger'}" would just do the "trick".

packages/bootstrap-vue-next/src/components/BAccordion/BAccordionItem.vue
xvaara: For me it would make sense for the class="foo" to go to the wrapping element (as it will go by default). That seems "correct" for me.

xvaara: maybe remove inheritAttrs:false and  have the root to inherit all the attrs and have attrs-props for all the other places needed?

VividLemon: > For me it would make sense for the class="foo" to go to the wrapping element (as it will go by default). That seems "correct" for me.

The issue that I have is that even though it does make sense, I think the attrs inheritance should fall onto the most common component. 

Like if you were to think about from a users perspective about 
```html
<div class="form-check">
  <input class="form-check-input" type="checkbox" value="" id="flexCheckDefault">
  <label class="form-check-label" for="flexCheckDefault">
    Default checkbox
  </label>
</div>
```
The most commonly modified component would be the input. How many people actually want to affect the wrapper div? Perhaps <5%?

xvaara: Sure there is a point in that for inputs. For example I might want to add margin to the wrapping with class="my-3" and it would be weird if that went to the input element.

packages/bootstrap-vue-next/src/components/BAccordion/BAccordionItem.vue
xvaara: I changed all them to attrs and it "feels" fine. See example in the beginning.

packages/bootstrap-vue-next/src/components/BAccordion/BAccordionItem.vue
xvaara: you meen we want to allow user overwrite the aria-labelledby or we want to protect the user from overwriting it?

packages/bootstrap-vue-next/src/components/BAccordion/BAccordionItem.vue
```diff
@@ -62,6 +69,10 @@ const props = withDefaults(
     horizontal?: Booleanish
     visible?: Booleanish
     isNav?: Booleanish
+    headerAttrs?: AttrsValue
+    buttonAttrs?: AttrsValue
+    collapseAttrs?: AttrsValue
+    bodyAttrs?: AttrsValue
```
VividLemon, line 75: I meant doing 
```suggestion
    headerAttrs?: AttrsValue
    headerClass?: ClassValue
    buttonAttrs?: AttrsValue
    buttonClass?: ClassValue
    collapseAttrs?: AttrsValue
    collapseClass?: ClassValue
    bodyAttrs?: AttrsValue
    bodyClass?: ClassValue
```

VividLemon (changes requested):

packages/bootstrap-vue-next/src/components/BAccordion/BAccordionItem.vue
VividLemon: I don't really care if someone overwrites the values. That's their problem. But also their benefit as they can make it what they want

packages/bootstrap-vue-next/src/components/BAccordion/BAccordionItem.vue
VividLemon: > if we have bodyAttrs is bodyClass even needed? bodyAttrs="{class: 'bg-danger'}" would just do the "trick".

It would work just fine, but having both is simply how I'd prefer it. imo it makes it easier.

packages/bootstrap-vue-next/src/components/BAccordion/BAccordionItem.vue
VividLemon: > Sure there is a point in that for inputs. For example I might want to add margin to the wrapping with class="my-3" and it would be weird if that went to the input element.

The tie breaker would be whatever element attrs falls onto in Bootstrap-vue. But that does seem somewhat strange as they inherit attrs onto the input, but the class falls into the wrapper element :thinking: Hmm. Not a great example for me. They do this quite a bit with their input elements. 

So, going off of this, lets simply say that attribute inheritance should match the behavior of Bootstrap-vue. Then we add new props to fill in the gaps. Sound good?

VividLemon: Keep in mind, if you were to expand this PR to fully finish https://github.com/orgs/bootstrap-vue-next/projects/1?pane=issue&itemId=31679414 , it might be a good time now since the next version to go out will be 0.15. Which still isn't going to be released for awhile, so no rush, assuming you want to do it.

xvaara: So now I kinda reverted the change I made to my previous. So now the ...attrs go to BCollapse and class goes to wrapper. Should this be the "way" forward for the other components?

packages/bootstrap-vue-next/src/components/BAccordion/BAccordionItem.vue
```diff
@@ -1,11 +1,12 @@
...
-  <div class="accordion-item">
+  <div class="accordion-item" v-bind="wrapperAttrs" :class="[wrapperClass, directWrapperClass]">
     <BCollapse
       :id="computedId"
       v-model="modelValue"
       class="accordion-collapse"
-      v-bind="$attrs"
+      :class="collapseClass"
       :aria-labelledby="`${computedId}-heading`"
+      v-bind="{...attrs, ...collapseAttrs}"
```
VividLemon, line 9: collapseAttrs can be removed, as you should just use basic attrs

packages/bootstrap-vue-next/src/components/BAccordion/BAccordionItem.vue
```diff
@@ -1,11 +1,12 @@
 <template>
-  <div class="accordion-item">
+  <div class="accordion-item" v-bind="wrapperAttrs" :class="[wrapperClass, directWrapperClass]">
```
VividLemon, line 2: wrapperClass can also be removed as you should just use the `directWrapperClass`

packages/bootstrap-vue-next/src/components/BAccordion/BAccordionItem.vue
xvaara: I was thinking: It's not needed, but for the sake of symmetry, all the places have xxxClass and xxxAttrs, just to make sure that if some finds it confusing where the class and other attrs go. and the class and other attrs go where they go in bv2

VividLemon: I think it should be fine. Like I said, I'd like to add this stuff to the documentation, but I don't have a plan yet on how to relay this information.

xvaara: is it's fine to leave or it's fine to stay? Just want to be clear on that :smile:

VividLemon: Good to remove.

VividLemon:
Moving forward this will likely be the same workflow as other components for https://github.com/orgs/bootstrap-vue-next/projects/1/views/1?pane=issue&itemId=31679414 . Though, be sure that the attrs/class inheritance follows BV

PR #1551, opened 2023-11-03
docs: update shows component reference only when data is present

dploc96:
# Describe the PR

Only show props, event or slot when data is available

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)


apps/docs/src/components/ComponentReference.vue
```diff
@@ -119,6 +119,9 @@ const sections = ['Properties', 'Events', 'Slots'] as const
 
 const buildCompReferenceLink = (str: string): string => `#comp-reference-${str}`.toLowerCase()
 
+const sectionAvailable = (component: ComponentReference) =>
```
VividLemon, line 122: I subjectively feel like this filter could go in the above sortData computed somehow. Without playing around with it, I'm not sure how to do it, but it could be more performant since its in the computed.

dploc96: Thank you @VividLemon for clarifying the issue for me. But something is preventing me from passing the tests.

VividLemon:
Just need to fix the merge conflict then I can merge

PR #1553, opened 2023-11-03
docs(BModal): Update props and events

dwgray:
# Describe the PR

Updates all of the props and events for the BModal component.

Where possible, I copied the documentation from bootstrap-vue. I tested each property and event to ensure the documentation is consistent with the current behavior of bootstrap-vue-next.

This led to one issue - it looks like [BS5 Close Button](https://getbootstrap.com/docs/5.3/components/close-button/) does not support variants, and classes aren't passed through.  I don't see an easy way to make the code support headerCloseVariant or headerCloseClass, so I marked them as deprecated.  I would be happy to pull them out of the code, but would prefer to do that as a separate PR.

As I got into the events, I discovered that the documentation framework doesn't implement Events and Slots correctly, so I didn't continue on to documenting props.  I would be happy to create an issue to implement Events and Slots in the documentation framework if that would be helpful.  I may also have time next week to tackle that issue if no-one else if looking at it.

Finally, it looks like most of the comments around line 130 seem out of date.  Most of these seem to be implemented. The exception is `close on escape when autofocus`, which if rewritten as `close on escape when autofocus===true` still seems to be true. If someone can validate my finding there, I'd be happy to remove the comments and add an issue for the one remaining bug.

## Small replication

A small replication or video walkthrough can help demonstrate the changes made. This is optional but can help observe the intended changes. A mentioned issue that contains a replication also works.

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

- [X ] Pull request title and all commits follow the [**Conventional Commits**](https://www.conventionalcommits.org/) convention or has an [**override**](https://github.com/googleapis/release-please#how-can-i-fix-release-notes) in this pull request body **This is very important, as the `CHANGELOG` is generated from these messages, and determines the next version type. Pull requests that do not follow conventional commits or do not have an override will be denied**


dwgray:
I'm new to github processes, but from what I can tell, it looks like all PRs are currently failing the ci /test-lint and ci /test-unit checks.  And I tried creating a trivial PR from my currently fully sync'd fork, and I'm getting the same failures.  If I'm missing something and I'm causing these failures, I'd appreciate a nudge in the right direction so that I can resolve the failures.

apps/docs/src/data/components/modal.data.ts
```diff
@@ -5,355 +5,435 @@ export default {
...
           prop: 'headerClass',
           type: 'ClassValue',
           default: undefined,
+          description: "CSS class (or classes) to apply to the '.modal-header' wrapper element",
         },
         {
           prop: 'headerCloseClass',
           type: 'ClassValue',
           default: undefined,
+          description: 'Deprecated by BS V5',
```
VividLemon, line 148: You've declared this, and some other items "deprecated". But they're perfectly valid classes. Curious where you got this information

apps/docs/src/data/components/modal.data.ts
```diff
@@ -5,355 +5,435 @@ export default {
...
         },
         {
           prop: 'footerTextVariant',
           type: 'ColorVariant | null',
           default: null,
+          description: 'Applies one of the Bootstrap theme color variants to the footer text',
         },
         {
           prop: 'fullscreen',
           type: 'boolean | string',
```
VividLemon, line 121: After I had saw this line, I had gone on a spree to fix up any values that were 'boolean | ${something else}'. I made a bunch of changes in https://github.com/bootstrap-vue-next/bootstrap-vue-next/pull/1555 , and this is one of them. So you may want to double check the changes you made to see if they line up with the new current values

packages/bootstrap-vue-next/src/components/BModal.vue
```diff
@@ -201,44 +201,39 @@ const props = withDefaults(
...
-    modalClass: undefined,
-    id: undefined,
     busy: false,
     lazy: false,
     buttonSize: 'md',
     cancelDisabled: false,
     cancelTitle: 'Cancel',
     cancelVariant: 'secondary',
     centered: false,
+    contentClass: undefined,
```
VividLemon, line 215: I don't quite get what's going on here. I think you alphabetized this. But backdropVariant and autoFocusButton are at the end. Which is weird.

VividLemon (changes requested):

VividLemon:
> I'm new to github processes, but from what I can tell, it looks like all PRs are currently failing the ci /test-lint and ci /test-unit checks. And I tried creating a trivial PR from my currently fully sync'd fork, and I'm getting the same failures. If I'm missing something and I'm causing these failures, I'd appreciate a nudge in the right direction so that I can resolve the failures.

They are failing because I have some active WIP stuff regarding major BTable changes.

packages/bootstrap-vue-next/src/components/BModal.vue
dwgray: I noted this in the PR description, but I tend to be wordy, so I should have put it closer to the code.  This set of properties is listed three times:  1. The declaration (in the defineProps statement), 2: the default assignment (here) 3: the documentation.  I don't have any strong preference as to the order, but it was quite confusing when the declaration and the defaults were in a different order (it makes it a lot easier to verify that everything matches).  Since the order in the documentation file and the order in the definition were basically the same, I went with that and just re-arranged the defaults to match that order.  With this long a list, I think there is some argument to just make it alphabetical (that's what the user-visible version in the documentation does).  If you agree, I'm happy to do that.

apps/docs/src/data/components/modal.data.ts
dwgray: Okay - I'll look into that.

packages/bootstrap-vue-next/src/components/BModal.vue
VividLemon: Alphabetize is likely the best. Though naturally when adding new properties its easy to forget.

apps/docs/src/data/components/modal.data.ts
dwgray: Just two places.  You're right on this one. I re-tested it, and it works.  I'll push a correction for that in my next round.  I'll add a comment in the other place.

apps/docs/src/data/components/modal.data.ts
```diff
@@ -5,355 +5,435 @@ export default {
...
           prop: 'headerCloseLabel',
           type: 'string',
           default: 'Close',
+          description: 'Accessibility label for the header close button',
         },
         {
           prop: 'headerCloseVariant',
           type: 'ButtonVariant | null',
           default: 'secondary',
+          description: 'Deprecated by BS V5',
```
dwgray, line 160: It looks like [BS5 Close Button](https://getbootstrap.com/docs/5.3/components/close-button/) does not support variants.  Setting this property does not do anything, as far as I can tell.  If I'm missing something, please let me know.  Otherwise, I'm happy to pull the code around this and remove the property from the documentation.

[StackBlitz repo](https://stackblitz.com/edit/bootstra-vue-next-template-vk45ap?file=src%2Fcomponents%2FComp.vue) or just past this into the test app.vue in the dev environment.
```
<template>
  <BButton @click="modal = !modal"> Toggle modal </BButton>
  <BModal
    v-model="modal"
    title="Testing"
    header-text-variant="danger"
    header-close-variant="danger"
    header-close-label="Custom Close"
  >
    foobar
  </BModal>
</template>

<script setup lang="ts">
import {ref} from 'vue'

const modal = ref(false)
</script>
```

VividLemon: Its a good thing you can use a slot which converts it to a regular button

packages/bootstrap-vue-next/src/components/BModal.vue
VividLemon: You can revert your changes to the .vue file, I had alphabetized the code already. So everything should be in order there

dwgray: Okay, reverted + removed the comment in the PR.

apps/docs/src/data/components/modal.data.ts
dwgray: This instance is now documented correctly.

apps/docs/src/data/components/modal.data.ts
dwgray: I'm happy to look into this, but I don't think I should change anything until your changes for the alphabetizing of the properties go through to avoid conflicts.  Can we let this PR go through as is and I'll look into this as a followup?

apps/docs/src/data/components/modal.data.ts
dwgray: Let me know if you would like me to pull the code that backs this and then either remove the documentation for the property or add a more detailed description of why it's deprecated (I would prefer to do the former - just clean it up) - probably more appropriate to put the deprecation in a migration document?  In any case, I don't want to touch the code until your alphabetize PR goes through, so would prefer to do this as a follow up.

apps/docs/src/data/components/modal.data.ts
dwgray: Sorry, I misread your comments.  I didn't realize that you touched this on Friday so I thought you wanted me to look into it, not that you had already fixed it in the code and just needed the documentation to stay in sync.  I've updated this in the documentation.

apps/docs/src/data/components/modal.data.ts
dwgray: @VividLemon Let me know how you'd like me to handle this.  I think it's the only outstanding issue on this PR.

apps/docs/src/data/components/modal.data.ts
```diff
@@ -5,355 +5,435 @@ export default {
...
+          description:
+            'Emitted when the modal tried to close, but was prevented from closing.  This occurs when preventDefault() is called on the event, the user clicks escape and no-close-onbackdrop is set to true, or the user clicks on the backdrop and no-close-onbackdrop is set to true.',
         },
         {
-          args: [],
-          description: '',
           event: 'show-prevented',
+          args: [],
+          description:
+            'Emitted when the modal tried to open, but was prevented from opening. THis occurs when preventDrault() is called on the event',
```
VividLemon, line 403:
```suggestion
            'Emitted when the modal tried to open, but was prevented from opening. This occurs when preventDefault() is called on the event',
```

apps/docs/src/data/components/modal.data.ts
```diff
@@ -5,355 +5,435 @@ export default {
...
           prop: 'headerCloseLabel',
           type: 'string',
           default: 'Close',
+          description: 'Accessibility label for the header close button',
         },
         {
           prop: 'headerCloseVariant',
           type: 'ButtonVariant | null',
           default: 'secondary',
+          description: 'Deprecated by BS V5',
```
VividLemon, line 173:
```suggestion
          description: 'Applies a variant to the header close button when the header close button uses the header-close slot',
```

VividLemon (changes requested):

Issue #1573, opened 2023-11-08
Add docs search feature


dploc96:
### Clear and concise description of the problem

The documents are increasing more and more. Adding a search feature is really necessary.

### Suggested solution

n

### Validations

- [X] Follow our [Code of Conduct](https://github.com/bootstrap-vue-next/bootstrap-vue-next/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guide](https://github.com/bootstrap-vue-next/bootstrap-vue-next/blob/main/CONTRIBUTING.md).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

PR #1577, opened 2023-11-09
docs: add search feature

dploc96:
# Describe the PR

Resolve https://github.com/bootstrap-vue-next/bootstrap-vue-next/issues/1573

## Small replication

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)


apps/docs/.vitepress/theme/Layout.vue
```diff
@@ -261,18 +279,29 @@ watch(
 </script>
 
 <style lang="scss">
+@import 'vitepress/dist/client/theme-default/styles/vars.css';
```
VividLemon, line 282: I wonder if its possible to get `VPNavBarSearch` functionality, without using their component. The goal would be to keep it internal and use bootstrap components. However, if this is too much work, I have no issue with using their component.

VividLemon (changes requested):

apps/docs/.vitepress/theme/Layout.vue
VividLemon: Some stuff is here 

https://github.com/vuejs/theme/blob/main/src/vitepress/components/VPAlgoliaSearchBox.vue

https://github.com/vuejs/theme/blob/main/src/vitepress/composables/config.ts

Not sure how much this helps though

Issue #1578, opened 2023-11-09
BModal's overlay is a circle


dwgray:
### Describe the bug

This appears to be a regression from #1555

useRadiusElementClasses.ts is returning a value of `${strValue}rounded-circle` when false is passed in.  I believe it should return '' to match the [documentation](https://bootstrap-vue-next.github.io/bootstrap-vue-next/docs/components/overlay.html)

This reproduces with any modal that doesn't override backdrop properties (explicitly the rounded property).

![image](https://github.com/bootstrap-vue-next/bootstrap-vue-next/assets/5083020/d59a06fd-60c8-4f9c-a79f-d7e1e106c283)

```
<template>
  <BButton @click="modal = !modal"> Toggle modal </BButton>
  <BModal v-model="modal" title="Test Modal"> foobar </BModal>
  Some background content
</template>

<script setup lang="ts">
import {ref} from 'vue'

const modal = ref(false)
</script>
```

### Reproduction

See code/screenshot above

### System Info

```Shell
This is not system-dependent
```

### Used Package Manager

pnpm

### Validations

- [X] Have tested with the latest version. This is still alpha version and sometime things change rapidly.
- [X] Follow our [Code of Conduct](https://github.com/bootstrap-vue-next/bootstrap-vue-next/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guide](https://github.com/bootstrap-vue-next/bootstrap-vue-next/blob/main/CONTRIBUTING.md).
- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A, please open a GitHub Discussion instead.
- [X] The provided reproduction is a [minimal reproducible](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

dwgray:
I have a fix for this, so if someone how has access could assign the issue to me, I'd appreciate it.

PR #1579, opened 2023-11-09
fix(BOverlay): fixes rounded prop #1578

dwgray:
# Describe the PR

- Fix the use of the rounded property in BOverlay and BImg.
- Small cleanup of the useRadiusElementClasses to correctly handle toolean inputs (false returns no rounded classes and true returns a bare {border}rounded* class.
- Added unit tests for useRadiusElementClasses

## Small replication

This started out as a fix for PR#1555, which was breaking the test overlay.spec.ts!child BTransition child div first child div has class rounded when prop rounded true.  This PR fixes that test and also extends the test that checked that the rounded class wasn't added when rounded===false to catch future regressions of this nature.

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)


packages/bootstrap-vue-next/src/components/BOverlay/overlay.spec.ts
```diff
@@ -326,14 +326,18 @@ describe('', () => {
...
 
-  it('child BTransition child div first child div does not have class rounded when prop rounded false', async () => {
+  it('child BTransition child div first child div does not have any rounded* class when prop rounded false', async () => {
     const wrapper = mount(BOverlay, {
       props: {show: true, rounded: false},
     })
     const $transition = wrapper.getComponent(BTransition)
     const $div = $transition.get('div')
     const $second = $div.get('div')
+    // First check the explicit calsses that have caused this to break, which will give a more readable error message
```
dwgray, line 336: I thought this was a reasonable compromise between polluting the test space, getting useful failure messages and still catching all the cases that we'd want covered.  The alternative I considered was to create three tests, one for 'rounded', one for 'rounded-circle' and a third catch-all.  Happy to move to that if you prefer.

VividLemon:
This issue was already fixed.

packages/bootstrap-vue-next/src/components/BOverlay/overlay.spec.ts
VividLemon: Tests are meant to be small and test only one part of the behavior. If an item can have multiple different outputs, a unique test should be made to test each of the different outputs. 

However, in this case most of the heavy lifting is to be made as a useRadiusClasses.spec.ts test. The idea being that it is its own function and should be tested individually, as there are multiple places that use it.

dwgray: Fair enough.  A general expectation of one 'expect' statement per test is a reasonable bar.  I'm not seeing useRadiusClasses.spec.ts test, am I missing something?

VividLemon: > I'm not seeing useRadiusClasses.spec.ts test

It was not made

dwgray:
> This issue was already fixed.

@VividLemon I believe this PR is still worthwhile in that it fixes and extends the test.  It is also compatible with bootstrap-vue.

You've introduced a new undocumented `none` value as the default for the rounded prop.  While I can see an argument for that being cleaner than false, anyone who used false explicitly for this prop will have to change it to none (and the way the code in main reads now, they'll get a circular overlay, which seems completely wrong).

An alternative would be to leave the existing boolean behavior in place and add `none` as an option, which is effectively what I have here.  If that works for you, I'll go ahead and document the 'none' values once this goes through.  Otherwise, I'll just go ahead and close this down, as this isn't blocking me.

packages/bootstrap-vue-next/src/components/BOverlay/overlay.spec.ts
dwgray: Okay - I think I can throw tests for useRadiusClasses together pretty easily if you think that would be useful.  Is it all right if I scaffold it and add a couple of tests in a draft PR to make sure I'm on the right track?

VividLemon:
> > This issue was already fixed.
> 
> @VividLemon I believe this PR is still worthwhile in that it fixes and extends the test. It is also compatible with bootstrap-vue.
> 
> You've introduced a new undocumented `none` value as the default for the rounded prop. While I can see an argument for that being cleaner than false, anyone who used false explicitly for this prop will have to change it to none (and the way the code in main reads now, they'll get a circular overlay, which seems completely wrong).
> 
> An alternative would be to leave the existing boolean behavior in place and add `none` as an option, which is effectively what I have here. If that works for you, I'll go ahead and document the 'none' values once this goes through. Otherwise, I'll just go ahead and close this down, as this isn't blocking me.

The issue is that the behavior modifies that of the overlay component and diverges from BV there. In BV their backdrop is a custom made implementation without the overlay component. What I noticed was that the overlay component could be expanded to the modal, and some other places, which made it possible to do some (albeit not very useful) custom stuff like adding backdrop variants to the modal. 

With that being said, I agree that the behavior is strange to have false set the value to circle. I recall seeing something in BV that did that, but it should be no class. So, as it stands I see nothing wrong with the changes in `useRadiusElementClasses.ts`, however, components that use it should likely change up their defaults. Specifically BModal 'none', should probably be false, I think bimg too.

packages/bootstrap-vue-next/src/components/BOverlay/overlay.spec.ts
VividLemon: Yes

dwgray:
> > > This issue was already fixed.
> > 
> > 
> > @VividLemon I believe this PR is still worthwhile in that it fixes and extends the test. It is also compatible with bootstrap-vue.
> > You've introduced a new undocumented `none` value as the default for the rounded prop. While I can see an argument for that being cleaner than false, anyone who used false explicitly for this prop will have to change it to none (and the way the code in main reads now, they'll get a circular overlay, which seems completely wrong).
> > An alternative would be to leave the existing boolean behavior in place and add `none` as an option, which is effectively what I have here. If that works for you, I'll go ahead and document the 'none' values once this goes through. Otherwise, I'll just go ahead and close this down, as this isn't blocking me.
> 
> The issue is that the behavior modifies that of the overlay component and diverges from BV there. In BV their backdrop is a custom made implementation without the overlay component. What I noticed was that the overlay component could be expanded to the modal, and some other places, which made it possible to do some (albeit not very useful) custom stuff like adding backdrop variants to the modal.
> 
> With that being said, I agree that the behavior is strange to have false set the value to circle. I recall seeing something in BV that did that, but it should be no class. So, as it stands I see nothing wrong with the changes in `useRadiusElementClasses.ts`, however, components that use it should likely change up their defaults. Specifically BModal 'none', should probably be false, I think bimg too.

Okay - thanks for the background on that.  Makes sense to me.

BModal doesn't directly consume `useRadiusElementClasses`.  BImg and BAvatar do.  I changed BImage to `false` and BAvatar was already `false`.

VividLemon:
> > > > This issue was already fixed.
> > > 
> > > 
> > > @VividLemon I believe this PR is still worthwhile in that it fixes and extends the test. It is also compatible with bootstrap-vue.
> > > You've introduced a new undocumented `none` value as the default for the rounded prop. While I can see an argument for that being cleaner than false, anyone who used false explicitly for this prop will have to change it to none (and the way the code in main reads now, they'll get a circular overlay, which seems completely wrong).
> > > An alternative would be to leave the existing boolean behavior in place and add `none` as an option, which is effectively what I have here. If that works for you, I'll go ahead and document the 'none' values once this goes through. Otherwise, I'll just go ahead and close this down, as this isn't blocking me.
> > 
> > 
> > The issue is that the behavior modifies that of the overlay component and diverges from BV there. In BV their backdrop is a custom made implementation without the overlay component. What I noticed was that the overlay component could be expanded to the modal, and some other places, which made it possible to do some (albeit not very useful) custom stuff like adding backdrop variants to the modal.
> > With that being said, I agree that the behavior is strange to have false set the value to circle. I recall seeing something in BV that did that, but it should be no class. So, as it stands I see nothing wrong with the changes in `useRadiusElementClasses.ts`, however, components that use it should likely change up their defaults. Specifically BModal 'none', should probably be false, I think bimg too.
> 
> Okay - thanks for the background on that. Makes sense to me.
> 
> BModal doesn't directly consume `useRadiusElementClasses`. BImg and BAvatar do. I changed BImage to `false` and BAvatar was already `false`.

I was referring to BOverlay, not BModal. 'none' will work, but it should probably remain false

dwgray:
@VividLemon please hold off on looking at my latest set of changes - I misread the bootstrap documentation and was convinced that there wasn't a bare 'rounded' class, so I stripped it out.  But after I pushed these changes and went back to check the documentation to document my changes, I realized my mistake.  I'll have a corrected version up shortly.  Sorry for the confusion.

packages/bootstrap-vue-next/tests/composables/useRadiusElementClasses.spec.ts
```diff
@@ -0,0 +1,210 @@
+import {useRadiusElementClasses} from '../../src/composables'
```
dwgray, line 1: This contains (nearly) exhaustive tests for the useRadiusElementClasses rounded prop and then a single test for the ${border}rounded property to verify that they all pass through.  Since the code is shared between all the properties, this seemed reasonable to me (although I guess that makes it not completely black-box testing). Let me know if you would like additional coverage of the other properties and what kind.

dwgray:
> @VividLemon please hold off on looking at my latest set of changes - I misread the bootstrap documentation and was convinced that there wasn't a bare 'rounded' class, so I stripped it out. But after I pushed these changes and went back to check the documentation to document my changes, I realized my mistake. I'll have a corrected version up shortly. Sorry for the confusion.

Okay - that was easier than I thought.  I believe this PR is ready for review.

packages/bootstrap-vue-next/src/components/BOverlay/overlay.spec.ts
```diff
@@ -336,6 +336,40 @@ describe('', () => {
...
   })
 
+  it('child BTransition child div first child div does not have class rounded-circle when prop rounded false', async () => {
+    const wrapper = mount(BOverlay, {
+      props: {show: true, rounded: false},
+    })
+    const $transition = wrapper.getComponent(BTransition)
+    const $div = $transition.get('div')
+    const $second = $div.get('div')
+    expect($second.classes()).not.toContain('rounded-circle')
```
VividLemon, line 346: This test doesn't assert much. Considering it could be any rounded value. Rounded false could magically make it rounded-3, yet this test wouldn't know

packages/bootstrap-vue-next/src/components/BOverlay/overlay.spec.ts
```diff
@@ -336,6 +336,40 @@ describe('', () => {
...
+  })
+
+  it('child BTransition child div first child div does not have any rounded* class when prop rounded false', async () => {
+    const wrapper = mount(BOverlay, {
+      props: {show: true, rounded: false},
+    })
+    const $transition = wrapper.getComponent(BTransition)
+    const $div = $transition.get('div')
+    const $second = $div.get('div')
+    expect($second.classes().find((str) => str.startsWith('rounded'))).toBeUndefined()
```
VividLemon, line 356: This is more of what the expected test would look like.

packages/bootstrap-vue-next/src/components/BOverlay/overlay.spec.ts
```diff
@@ -336,6 +336,40 @@ describe('', () => {
...
+  it('child BTransition child div first child div does not have any rounded* class when prop rounded false', async () => {
+    const wrapper = mount(BOverlay, {
+      props: {show: true, rounded: false},
+    })
+    const $transition = wrapper.getComponent(BTransition)
+    const $div = $transition.get('div')
+    const $second = $div.get('div')
+    // First check the explicit calsses that have caused this to break, which will give a more readable error message
+    expect($second.classes()).not.toContain('rounded')
+    expect($second.classes()).not.toContain('rounded-circle')
```
VividLemon, line 368: All of these small tests per component are best to be left to the composable test. The only test you'd probably need to do are, testing the default of rounded prop (rounded false, no class) and say rounded true (that it does render "something")

VividLemon (changes requested):

packages/bootstrap-vue-next/src/components/BOverlay/overlay.spec.ts
dwgray: > This test doesn't assert much. Considering it could be any rounded value. Rounded false could magically make it rounded-3, yet this test wouldn't know

Sorry, I'm not sure how this crept in.  I agree that this and the test above `child BTransition child div first child div does not have class rounded when prop rounded false` (which I didn't write) are fairly worthless, which is why I wrote the next test which asserts something stronger.  I'd like to delete this test and the one that assert .not.toContain('rounded').

packages/bootstrap-vue-next/src/components/BOverlay/overlay.spec.ts
dwgray: > All of these small tests per component are best to be left to the composable test. The only test you'd probably need to do are, testing the default of rounded prop (rounded false, no class) and say rounded true (that it does render "something")

Sorry - this and the other test you objected to crept in when I did a cherry-pick from a different branch.  I though I ran through the code after I did that, but didn't do a thorough enough job.  I'll fix this up.

dwgray: Okay - I've trimmed this to two tests at this level: Rounded=true, Rounded=false.  I agree that the composable tests I've added provide sufficient coverage for the rest.

VividLemon:
gw

PR #1580, opened 2023-11-10
docs(BButtonGroup): update component references

dploc96:
# Describe the PR

Change on BButtonGroup & BButtonToolbar

## Small replication

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)


apps/docs/src/data/components/button.data.ts
```diff
@@ -9,64 +9,195 @@ export default {
...
+          description: `<RouterLink> prop: Setting append prop always appends the relative path to the current path`,
+        },
+        {
+          prop: 'disabled',
+          type: 'Booleanish',
+          default: undefined,
+          description: `When set to 'true', disables the component's functionality and places it in a disabled state`,
+        },
+        {
+          prop: 'event',
```
VividLemon, line 99: So, instead of declaring the extended props, say those in BLink, instead what we should do is find a way to _extend_ the props list. 

This would make it so you don't need to change like 10 other files if BLink changes (or another thing that extends another, like BProgress)

What I would like to see is in the rendered table section for the component reference, a new "group" is made, the group has the header of its extended props, then in its section it declares all of its data.

What I would like to see is something like this:
image
Its a bit of a challenge, technically speaking. You'd have to: create a new property on the component reference data, import the sub-data, then copy the omit() function to remove keys like `omit(BButtonProps, ['variant'])`, put the correct prop data on the component reference, then go into the Vue file and have the logic build out. I'm also doubting that BTable is in a position to fulfill this out of the box.

However, when its done, it will be a lot more straightforward than re-declaring each prop

VividLemon (changes requested):

apps/docs/src/data/components/button.data.ts
VividLemon: (https://github.com/bootstrap-vue-next/bootstrap-vue-next/issues/1584)

apps/docs/src/data/components/ComponentReference.ts
```diff
@@ -10,16 +10,19 @@ export interface ComponentReference {
     type: string
     description?: string
     default?: unknown
+    children?: Omit<ComponentReference['props'][number], 'children'>[]
```
VividLemon, line 13: A better way would be to extend the entire props list at once. Rather than declaring a prop a child of something else. Then just using the entire list

Of course, you will need to omit the real values, you can probably reuse the util -- its called `omit`. I think its in object.ts

Something like 
```ts
parentProps: {
  BLink: omit(BLinkProps, ['event', '/* some other omitted props*/]),
  SomeOtherFakeComponent: ... // and so on
}
```
You can then use Object.entries and whatnot

PR #1595, opened 2023-11-16
Fix types: select, checkbox should accept readonly array for options

kalekseev:
# Describe the PR

We have a lot of `const options = [...] as const` in our codebase and tsc complains that b-form-select wants mutable array while we have read only array.
It probably make sense to make all array props readonly throughout all codebase.

## Small replication

```
          <b-form-select
            :options="[{ value: '', text: '' }] as const"
          />
```

Results in `Type 'readonly [{ readonly value: ""; readonly text: ""; }]' is not assignable to type 'unknown[] | Record<string, unknown> | undefined'.`

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

### DIFF
--- a/packages/bootstrap-vue-next/src/components/BFormCheckbox/BFormCheckbox.vue
+++ b/packages/bootstrap-vue-next/src/components/BFormCheckbox/BFormCheckbox.vue
@@ -52,7 +52,7 @@ const props = withDefaults(
     id?: string
     indeterminate?: Booleanish
     inline?: Booleanish
-    modelValue?: CheckboxValue | CheckboxValue[]
+    modelValue?: CheckboxValue | ReadonlyArray<CheckboxValue>
     name?: string
     plain?: Booleanish
     required?: Booleanish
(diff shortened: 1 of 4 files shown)


kalekseev:
> The issue that I have with this PR is that it only targets one place in a disorganized way. 

Fair enough I pushed a new version that makes values and options props readonly for selects and checkboxes. 

> Nor am I able to confirm whether or not downstream users would experience TypeScript breaking down with those not using readonly arrays.

Making props readonly won't breaks typings but in the new version I also changed emits to readonly. Upstream users shouldn't mutate objects that the library emits but that means that TS will complain in cases when user try to assign readonly array to mutable one, in that case user can rebuild array `@update:modelValue = "mutable = [...$event]" or change variable type to readonly.

We can leave emitted values mutable but in my view it better to provide the right intent as TS docs says:

> Much like the readonly modifier for properties, it’s mainly a tool we can use for intent. When we see a function that returns ReadonlyArrays, it tells us we’re not meant to change the contents at all, and when we see a function that consumes ReadonlyArrays, it tells us that we can pass any array into that function without worrying that it will change its contents.

VividLemon (changes requested):
It would be nice if this were extended to objects and happened to all components (I don't know if there are more, I'm just assuming yes)

VividLemon:
Superseded by https://github.com/bootstrap-vue-next/bootstrap-vue-next/pull/1659

VividLemon:
Thank you for the original though

PR #1631, opened 2023-11-25
feat: add Scrollspy composable and directive

xvaara:
# Describe the PR

## Small replication

## PR checklist

- [x] Feature - `feat(...)`
- [ ] Other (please describe)


VividLemon:
The scrollspy in bv was a directive. It could possibly be a composable. But I'm curious on why you made a component

xvaara:
> The scrollspy in bv was a directive. It could possibly be a composable. But I'm curious on why you made a component

Don't like directives. I can't return stuff from directives, so  that was a no-go.
It could be a composable but requires more setup. 

check TScrollspy to se the four different ways you can use this.

VividLemon:
I agree directives aren't the best, but I can't really see a reason why it needs a renderless component variation. I don't really like renderless components either. As renderless components can usually just be composables (and avoid the performance cost of being a component)

Also, the intersection observer code could be replaced with the vueuse utility. Probably wouldn't need to deal with cleanup and whatnot by using it

VividLemon:
If the current directives list could be made composables, I'd find that ideal. But rendering components and shit in directives is a pain, so I'd imagine doing the same in composables would be the same. But if you're adventurous you could try, and we can make a 0.16 lol

xvaara:
Yes, I just needed a scrollspy for a site. Did a quick mockup and then just finished it yesterday. I just know components and grabbed what I know. Well now I know how to make scrollspy, so I'll convert this to composable.

It seems vueuse makes composables in a way that they are easy to use ase directives also.
https://github.com/vueuse/vueuse/blob/main/packages/core/useIntersectionObserver/directive.ts

gonna check that out too.

packages/bootstrap-vue-next/src/composables/useScrollspy.ts
```diff
@@ -0,0 +1,211 @@
...
+  threshold: number | number[]
+  watchChanges: boolean
+}
+
+export default (
+  content: MaybeRefOrGetter<string | ComponentPublicInstance | HTMLElement | null>,
+  target: MaybeRefOrGetter<string | ComponentPublicInstance | HTMLElement | null>,
+  options: Partial<ScrollspyOptions> = {}
+): ScrollspyReturn => {
+  const resolvedContent = toRef(() => getElement(content))
```
VividLemon, line 48: This doesn't quite look right. This will most likely only work if getelement takes in a Ref. And I'd also check to see if reactivity is maintained. Keep in mind that toRef(MaybeRefOrGetter) will get you the value you need. If it doesn't work in the way described, you may want to use toref and a computed together. The concern is that toRef as a getter to unwrap other getters doesn't work. I've been confused by it to

packages/bootstrap-vue-next/src/utils/dom.ts
```diff
@@ -230,3 +230,23 @@ export const getTransitionDelay = (element: HTMLElement) => {
...
   return transitionDelayMs + transitionDurationMs
 }
+
+/**
+ * convert unknown input to HTMLElement
+ *
+ * @param target
+ */
+export const getElement = (
+  target: MaybeRefOrGetter<string | ComponentPublicInstance | HTMLElement | null>
```
VividLemon, line 240: The difference between a util and a composable is _mutability_ Everything is a functional utility until you add in stateful logic (ref, reactive, etc). Once you add one of those items in, it becomes a composable.

I'd probably recommend removing the stateful logic from this.

packages/bootstrap-vue-next/src/utils/dom.ts
```diff
@@ -230,3 +230,23 @@ export const getTransitionDelay = (element: HTMLElement) => {
...
+): HTMLElement | undefined => {
+  const element = unref(toRef(target))
```
VividLemon, line 242: unref(toRef()) is the same as using `toValue`

packages/bootstrap-vue-next/src/composables/useScrollspy.ts
VividLemon: Another way to think about (one of the issues) is that doing 
```ts
const a = ref(1)
const b = computed(() => a)
const c = () => { a.value++ }
```
It just doesn't work because the actual value is never processed. The same issue exists when using getters for toRef. `b` wont change because `a` never "changed".

packages/bootstrap-vue-next/src/composables/useScrollspy.ts
```diff
@@ -0,0 +1,211 @@
...
+  const nodeList = ref<HTMLElement[]>([])
+
+  // are we called in directive?
+  const ctx = getCurrentInstance()
+  if (!ctx) {
+    nextTick(() => {
+      updateList()
+    })
+  } else {
+    onMounted(() => {
```
VividLemon, line 71: The onMounted stuff is good, but I'd also double check to see if it all works in Nuxt. These deep dom manipulations could possibly have some issues if mismanaged in an SSR environment.

VividLemon (changes requested):

packages/bootstrap-vue-next/src/utils/dom.ts
xvaara: Vueuse has unrefElement and I'll refactor to that in another PR. Ignore this for now.

packages/bootstrap-vue-next/src/composables/useScrollspy.ts
xvaara: I'll check the ssr tomorrow

packages/bootstrap-vue-next/src/utils/dom.ts
xvaara: I don't undestand what stateful there is in this? it just unwraps the input and finds the element.

packages/bootstrap-vue-next/src/composables/useScrollspy.ts
xvaara: double checked and it works now. made the check a little better.

VividLemon:
Also I don't think you need to use unrefelement from vueuse. I think you can wrap a mayberef in a ref and use it with .value.classList (not class list, but that's the accessor)

packages/bootstrap-vue-next/src/utils/dom.ts
VividLemon: (Previously introducing a Ref would have been "stateful")

xvaara:
> Also I don't think you need to use unrefelement from vueuse. I think you can wrap a mayberef in a ref and use it with .value.classList (not class list, but that's the accessor)

Yeah, it was only like 2lines that getElement already did. I encountered more challenges with this, but now it's working.

packages/bootstrap-vue-next/src/composables/useScrollspy.ts
```diff
@@ -0,0 +1,226 @@
...
+  const resolvedContent = ref(getElement(content))
+  const resolvedTarget = ref(getElement(target))
+  const updateElements = () => {
+    resolvedContent.value = getElement(content)
+    resolvedTarget.value = getElement(target)
+  }
+  watch([() => toValue(content), () => toValue(target)], () => {
+    updateElements()
+    updateList()
+  })
```
VividLemon, line 60:
```suggestion
  const resolvedContent = readonly(toRef(content))  
  const resolvedTarget = readonly(toRef(target))
  
  const resContent = computed(() => getElement(resolvedContent.value))
  const resTarget = computed(() => getElement(resolvedTarget.value))
  
  watch([resTarget, resContent], () => {
    updateList()
  })
```

Or
```suggestion
  const resolvedContent = computed(() => getElement(toValue(content)))  
  const resolvedTarget = computed(() => getElement(toValue(target)))
  
  watch([resolvedTarget, resolvedContent], () => {
    updateList()
  })
```

Subsequently usage lower can use typical `.value` to get values, such as `if (typeof toValue(content)` => `if(typeof resolvedContent.value...`

packages/bootstrap-vue-next/src/utils/dom.ts
```diff
@@ -230,3 +230,21 @@ export const getTransitionDelay = (element: HTMLElement) => {
...
+
+/**
+ * convert unknown input to HTMLElement
+ *
+ * @param target
+ */
+export const getElement = (
+  target: MaybeRefOrGetter<string | ComponentPublicInstance | HTMLElement | null>
+): HTMLElement | undefined => {
+  const element = toValue(target)
```
VividLemon, line 242:
```suggestion
  target: string | ComponentPublicInstance | HTMLElement | null
): HTMLElement | undefined => {
  const element = target
```
'element' likely doesn't need to exist anymore

The consumption of the util should pass in `.value`, not the Ref

xvaara: so it should be in composables, or the toValue should be on calling end? Which do we want to prefer?

packages/bootstrap-vue-next/src/composables/useScrollspy.ts
xvaara: that doesn't work if the input is string. That is the reason for that cludge. Or do we want to remove the functionality of us doing the querying and only accept input as elements or refs on every component and directive?

VividLemon: There should be no reason why it doesn't work. In both cases the string and HTML element should be gathered

packages/bootstrap-vue-next/src/utils/dom.ts
VividLemon: The value should be given. So, getElement(el.value)

xvaara: so getElement(toValue(el)) since we don't check what user has given us?

packages/bootstrap-vue-next/src/composables/useScrollspy.ts
xvaara: The string doesn't update when the dom is "mounted" and the composable will have no element in the variable. We need to update it in onMounted if it's a string.

packages/bootstrap-vue-next/src/utils/dom.ts
VividLemon: toValue would be if what you have is a MaybeRefOrGetter (could be 1, could be Ref(1) or could be () => 1)

So if you have any of those, and dont know which one it is, then you'd use toValue

If you have a Ref, you'll want to use .value, as its just a ref

So, it depends on what you're working with

packages/bootstrap-vue-next/src/composables/useScrollspy.ts
VividLemon: Depending on what you'd want, you can either exclude the _readonly_ and sync the two refs (I don't think you can do that with a maybe ref or getter, as you can't sync a ref to a function, just doesn't make sense) or you can have it one way sync with a watcher, and you'll need to declare a ref. The second is what you're doing here, but you can write it more succinct by taking some of the recommended changes -- namely adding the toRef

packages/bootstrap-vue-next/src/utils/dom.ts
xvaara: So if I want to keep the toValue there because I don't care what user inputs and I don't want to remember to check  on multiple places on the caller end, it's a composable? For me the better DX is to have the function check it self, so dev doesn't have to care.

VividLemon: The issue to me really comes down to the single responsibility principle in that it should only be getting the value, not converting the parameter into the desired format, then getting the value

packages/bootstrap-vue-next/src/composables/useScrollspy.ts
xvaara: how does toRef help? How do I make it update it's return value?

packages/bootstrap-vue-next/src/utils/dom.ts
xvaara: This has nothing to do with scrollspy so I returned the function to BPopover.vue and copypasted it to useScrollspy.ts

PR #1649, opened 2023-11-29
useScrollspy changes

VividLemon:
@xvaara 

Going back to this code we have two issues.

First off, the MaybeRef/MaybeRefOrGetter being passed through a prop in the BPopover component. This was strange to me as you're not going to actually: A: send a ref, B: send a getter

You're not going to send a getter fn through as if you do 
```html
<BPopover :target="myValue" />

const myValue = ref<string>('')
```
It's already going to be reactive. A getter function would be unnecessary. 

```html 
<button ref="myValue">Click me</button>
<BPopover :target="myValue" />

const myValue = ref<HTMLElement | null>(null)
```
This is going to be the same thing. But "MaybeRef" is unnecessary as -- in the Popover's perspective -- the Ref is unwrapped. It should never actually receive a ref. So, in both of these, I'm confused as to why Refs/Getters are passed, as props are already going to be reactive

The second issue is that of the useScrollspy. By "reactifying" the composable inputs with toRef & by using a computed, you should be able to conver that ref input to a computed "getElement". Reactivity is persevered and you don't actually need to reset the value in onMounted, as long as the correct value is passed, the computed "getElement" should give you the value you're needing.

I haven't tested everything, but both Playground examples before/after appear to be identical. 

Before you had seemed confused as to the behavior of things like toValue, toRef, and possibly the reactivity of composables. This is understandable. Let me know if you have some questions, tbh we could get together at some point and I can explain it to you over discord if you need. But perhaps I'm missing something in these two changes, lmk


xvaara:
> The second issue is that of the useScrollspy. By "reactifying" the composable inputs with toRef & by using a computed, you should be able to conver that ref input to a computed "getElement". Reactivity is persevered and you don't actually need to reset the value in onMounted, as long as the correct value is passed, the computed "getElement" should give you the value you're needing.

As I explained and even left a comment on the code: This doesn't work with a string. I'm open for a different way if you can show me a way to make it work and without vue warnings.

xvaara:
About the bpopover, meybe we don't need to unwrap the getters or refs. Let's pass the responsibility to the user?
I can see someone asking why their array of refs don't work in v-for loop, or their clever getter functionality isn't triggered.

VividLemon:
> As I explained and even left a comment on the code: This doesn't work with a string. I'm open for a different way if you can show me a way to make it work and without vue warnings.

Okay, I understand the issue now. I thought the bulk of your issue just boiled down to managing the refs. However, the issue is actually with getElementById -- in that a string value on mounted can't get set because the dom content hasn't been loaded yet. I've made the change, its a close implementation to your original idea, but it is a bit more concise

> About the bpopover, maybe we don't need to unwrap the getters or refs. Let's pass the responsibility to the user?
I can see someone asking why their array of refs don't work in v-for loop, or their clever getter functionality isn't triggered

Well you have to think about it from the dev perspective. If you do
```html
<BPopover :target="someRef" />
```
someRef is going to be unwrapped automatically because its in the template. The user is probably not going to be able to pass a ref. In regards to the getter function, yes you would be able to pass in the function, but for what practical purpose does this serve? `:target="() => myReactive.foo"` when you could just do `:target="myReactive.foo"`? Its already reactive in the template... MaybeRef & MaybeRefOrGetter are really just for orchestrating reactivity with composables, not really for working with

xvaara:
til. syncRef from vueuse.

> > About the bpopover, maybe we don't need to unwrap the getters or refs. Let's pass the responsibility to the user?
> > I can see someone asking why their array of refs don't work in v-for loop, or their clever getter functionality isn't triggered
> 
> Well you have to think about it from the dev perspective. If you do
> 
> ```
> <BPopover :target="someRef" />
> ```
> 
> someRef is going to be unwrapped automatically because its in the template. The user is probably not going to be able to pass a ref. In regards to the getter function, yes you would be able to pass in the function, but for what practical purpose does this serve? `:target="() => myReactive.foo"` when you could just do `:target="myReactive.foo"`? Its already reactive in the template... MaybeRef & MaybeRefOrGetter are really just for orchestrating reactivity with composables, not really for working with

I've seen it, but I think it's better we keep this simple.

xvaara (approved):

PR #1662, opened 2023-12-05
docs: Create a typescript template for bug reporting

dwgray:
# Describe the PR

I find that since I work in TypeScript, it's an extra step to transform my bug repro to JS.  And, of course, if there is a TypeScript component to the bug, this will be useful.

There isn't a great deal of value-add for maintaining both a ts and js template, so I've just updated the vite template to support TypeScript.

I verified that the template works with stackblitz [here](https://stackblitz.com/~/github.com/dwgray/bootstrap-vue-next-ts-template)

## Small replication

N/A

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)


VividLemon:
I'd be fine with just having one template that is simply TS compatible. No real reason to keep the two separate (Plus the overhead of maintaining two copies, even though it doesn't change much)

dwgray:
> I'd be fine with just having one template that is simply TS compatible. No real reason to keep the two separate (Plus the overhead of maintaining two copies, even though it doesn't change much)

Okay, I just replaced the vite template with the TypeScript version.  I tried to minimize the diff, but there was some fighting between the prettier/lint styles that made that more difficult than I'd like.  Hopefully this is all right.  If I do another pass, I'd probably move to making the styling conform to the rest of the bsv-next project, which would make this diff bigger but probably be easier to maintain moving forward.

PR #1679, opened 2023-12-16
BModal documentation for shown event has incorrect timing

anrolmar:
# Describe the PR

Change the description for the shown event.

## Small replication

https://stackblitz.com/edit/github-vf39ah?file=src%2FApp.vue

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)


VividLemon:
The shown, show, hide and hidden could all be updated.

In addition to that, there are multiple components that define all four of those events. So, you could make a more generic definition, you could reuse the definitions for all components with those events

anrolmar:
> The shown, show, hide and hidden could all be updated.

I'll do that.

> In addition to that, there are multiple components that define all four of those events. So, you could make a more generic definition, you could reuse the definitions for all components with those events

I think this should be another issue

xvaara:
@VividLemon I'm a little confused. Isn't show emited before because it's cancelable? shown is emited after? at least that's how it works on the other components...

VividLemon:
> @VividLemon I'm a little confused. Isn't show emited before because it's cancelable? shown is emited after? at least that's how it works on the other components...

Yes, this PR has changed the wrong description

PR #1693, opened 2023-12-22
fix(BModal): modal emits show and hide more than once

### LINKED ISSUE #1655: BModal emits `show` and `hide` more than once
### Describe the bug

BModal emits `show` multiple times (three for `useModal()` and two for `v-model`) when shown, and emits `hide` twice when dismissed using the backdrop or either of the default buttons. `shown` and `hidden` are only emitted once (as expected).

### Reproduction

https://github.com/corymccarty/bootstrap-vue-next-modal-emits

### Used Package Manager

npm

### Validations

anrolmar:
# Describe the PR

- Emits show event only once when the modal is shown.
- Emits hide event only once when the modal is dismissed

fix #1655

## Small replication

https://github.com/corymccarty/bootstrap-vue-next-modal-emits

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)


VividLemon:
It's probably worthwhile to figure out _whos_ calling the showFn, to see who/why it's being called twice.

anrolmar:
It's calling twice, because showFn is calling in the `onBeforeEnter` and in the `watcher`:

`const onBeforeEnter = () => showFn()`

```
watch(modelValueBoolean, (newValue, oldValue) => {
  if (newValue === oldValue) return
  if (newValue === true) {
    showFn()
  } else {
    hide()
  }
})
```

On the other hand, we have the onAfterEnter event:

```
const onAfterEnter = () => {
  isActive.value = true
  pickFocusItem()
  emit('shown', buildTriggerableEvent('shown'))
  if (lazyBoolean.value === true) lazyLoadCompleted.value = true
}
```

The watcher is executed on the `onAfterEnter` because `modelValueBoolean` is binded with the `isActive` value, and it's set on the onAfterEnter.

That's the reason I added a rule on the showFn method to check if the modal was activated or not:

```
const showFn = () => {
  if (!isActive.value) {
    const event = buildTriggerableEvent('show', {cancelable: true})
    emit('show', event)
    if (event.defaultPrevented) {
      if (modelValue.value) modelValue.value = false
      emit('show-prevented')
      return
    }

    if (!modelValue.value) modelValue.value = true
  }
}
```

VividLemon:
How does one create a PR with 0 file changes?

anrolmar:
My changes were removing in this commit, https://github.com/bootstrap-vue-next/bootstrap-vue-next/pull/1693/commits/7bea51c1fc3c56ef8c0c34498bd97471c98d30c1 when I merged from the main branch.

They're on the current PR, again

packages/bootstrap-vue-next/src/components/BModal/BModal.vue
```diff
@@ -424,14 +424,16 @@ const hide = (trigger = '') => {
 // TODO: If a show is prevented, it will briefly show the animation. This is a bug
 // I'm not sure how to wait for the event to be determined. Before showing
 const showFn = () => {
-  const event = buildTriggerableEvent('show', {cancelable: true})
-  emit('show', event)
-  if (event.defaultPrevented) {
-    if (modelValue.value) modelValue.value = false
-    emit('show-prevented')
-    return
+  if (!isActive.value) {
```
VividLemon, line 427:
```suggestion
  if (isActive.value) return
```

Modify the logic to return early if the condition isn't met. Just makes nested lines cleaner

VividLemon (changes requested):

PR #1718, opened 2024-01-12
fix(docs): build infinite loop

### LINKED ISSUE #1702: Docs build infinite loop
### Describe the bug

There exists some issue with processing the docs. First, during build it executes an infinite loop. The second issue is that when "dev" navigation of pages declares "vitepress data not properly injected in app". I suspect these two things are related. I'm not sure when this started (though its from a series of my changes). Something in https://github.com/bootstrap-vue-next/bootstrap-vue-next/commit/ca16a168687dc0e17d6dc5c6210205c1911dd793

### Reproduction

f

### Used Package Manager

pnpm

### Validations

sidneygijzen:
# Describe the PR

Wrapping the `setInterval` in the SFC setup in a `inBrowser` check solves the infinite loop, see [this comment](https://github.com/vuejs/vitepress/issues/562#issuecomment-1106061640) in the vitepress issue queue. I was able to build the docs successfully locally without an infinite loop. 

Fixes #1702.

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

### DIFF
--- a/apps/docs/src/docs/components/toast.md
+++ b/apps/docs/src/docs/components/toast.md
@@ -242,6 +242,7 @@ import {data} from '../../data/components/toast.data'
 import ComponentReference from '../../components/ComponentReference.vue'
 import {BButtonGroup, BButton, BToast, useToast} from 'bootstrap-vue-next'
 import HighlightCard from '../../components/HighlightCard.vue'
+import {inBrowser} from 'vitepress'
 import {ref, h} from 'vue'
 
 const {show, hide, toasts} = useToast()
@@ -277,9 +278,11 @@ const hideMe = () => {
 
 const toastShowStr = ref('foo')
 
-setInterval(() => {
-  toastShowStr.value = toastShowStr.value === 'foo' ? 'bar' : 'foo'
-}, 1000)
+if (inBrowser) {
+  setInterval(() => {
+    toastShowStr.value = toastShowStr.value === 'foo' ? 'bar' : 'foo'
+  }, 1000)
+}
 
 const showReactive = () => {
   show(toastShowStr, () => ({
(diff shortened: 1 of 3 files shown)


VividLemon (changes requested):
Let's instead put the interval functions into onMounted hooks. Perhaps you can use useIntervalFn from vueuse, I'm not sure if they automatically run onMounted or not... If not, then you can create a composable, something like "useSafeIntervalFn" instead

sidneygijzen:
Thanks for the quick review. I will have a look.

sidneygijzen:
@VividLemon I think you only need one of the two options you described:

instead of
```
if (inBrowser) {
  setInterval(() => {
    firstRef.value.body = `${Math.random()}`
  }, 1000)
}
```

Both the `onMounted` variant
```
onMounted(() => {
  setInterval(() => {
    firstRef.value.body = `${Math.random()}`
  }, 1000)
})
```

or  the vueuse `useIntervalFn` works: 
```
useIntervalFn(() => {
  firstRef.value.body = `${Math.random()}`
}, 1000)
```
`useIntervalFn` [has a built-in check for `isClient`](https://github.com/vueuse/vueuse/blob/main/packages/shared/useIntervalFn/index.ts#L62), so an onMounted hook is not needed.

In both cases the docs build successfully. Which version do you prefer?

VividLemon:
> @VividLemon I think you only need one of the two options you described:
> 
> instead of
> 
> ```
> if (inBrowser) {
>   setInterval(() => {
>     firstRef.value.body = `${Math.random()}`
>   }, 1000)
> }
> ```
> 
> Both the `onMounted` variant
> 
> ```
> onMounted(() => {
>   setInterval(() => {
>     firstRef.value.body = `${Math.random()}`
>   }, 1000)
> })
> ```
> 
> or the vueuse `useIntervalFn` works:
> 
> ```
> useIntervalFn(() => {
>   firstRef.value.body = `${Math.random()}`
> }, 1000)
> ```
> 
> `useIntervalFn` [has a built-in check for `isClient`](https://github.com/vueuse/vueuse/blob/main/packages/shared/useIntervalFn/index.ts#L62), so an onMounted hook is not needed.
> 
> In both cases the docs build successfully. Which version do you prefer?

Just onMounted is fine

sidneygijzen:
Thanks. Changed it to onMounted.

Issue #1611, opened 2023-11-20
checkbox tri state support


VividLemon:
### Missing feature is from

bootstrap-vue

### Docs

http://localhost:8000/bootstrap-vue-next/docs/components/form-checkbox.html#indeterminate-tri-state-support

### Describe the missing feature

Feature is missing https://bootstrap-vue.org/docs/components/form-checkbox#indeterminate-tri-state-support

### Validations

- [X] Have tested with the latest version. This is still alpha version and sometime things change rapidly.
- [X] Follow our [Code of Conduct](https://github.com/bootstrap-vue-next/bootstrap-vue-next/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guide](https://github.com/bootstrap-vue-next/bootstrap-vue-next/blob/main/CONTRIBUTING.md).
- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A, please open a GitHub Discussion instead.
- [X] The provided link is to the documentation of the missing feature.

PR #1723, opened 2024-01-20
BPagination https://github.com/bootstrap-vue-next/bootstrap-vue-next/pull/1686 Rebuild

VividLemon:
# Describe the PR

A clear and concise description of what the pull request does.

## Small replication

A small replication or video walkthrough can help demonstrate the changes made. This is optional, but can help observe the intended changes. A mentioned issue that contains a replication also works.

## PR checklist

**What kind of change does this PR introduce?** (check at least one)

- [ ] Bugfix :bug: - `fix(...)`
- [ ] Feature - `feat(...)`
- [ ] ARIA accessibility - `fix(...)`
- [ ] Documentation update - `docs(...)`
- [ ] Other (please describe)

**The PR fulfills these requirements:**

- [ ] Pull request title and all commits follow the [**Conventional Commits**](https://www.conventionalcommits.org/) convention or has an [**override**](https://github.com/googleapis/release-please#how-can-i-fix-release-notes) in this pull request body **This is very important, as the `CHANGELOG` is generated from these messages, and determines the next version type. Pull requests that do not follow conventional commits or do not have an override will be denied**

PR #1686, opened 2023-12-20
fix(BPagination): re-implemented the missing first-page and last-page…

MewesK:
# Describe the PR

Seems like the first-page/last-page feature of BPagination got lost during the refactoring in commit cb9846230a00e1419dd1bfa7f7b12e4ac5258ba9. I tried to keep the spirit of the refactoring without sacrificing any of the old behaviors. I hope it fits the existing codebase.

## Small replication

You can see the broken behavior in the docs:
https://bootstrap-vue-next.github.io/bootstrap-vue-next/docs/components/pagination.html#goto-first-last-button-type
image

## PR checklist

**What kind of change does this PR introduce?** (check at least one)

- [x] Bugfix :bug: - `fix(...)`
- [ ] Feature - `feat(...)`
- [ ] ARIA accessibility - `fix(...)`
- [ ] Documentation update - `docs(...)`
- [ ] Other (please describe)

**The PR fulfills these requirements:**

- [x] Pull request title and all commits follow the [**Conventional Commits**](https://www.conventionalcommits.org/) convention or has an [**override**](https://github.com/googleapis/release-please#how-can-i-fix-release-notes) in this pull request body **This is very important, as the `CHANGELOG` is generated from these messages, and determines the next version type. Pull requests that do not follow conventional commits or do not have an override will be denied**

PR #1725, opened 2024-01-20
feat(BFormCheckBox): Implements tri-state checkbox

dwgray:
# Describe the PR

This PR implements tri-state checkboxes by creating an indeterminate named model that tracks the indeterminate attribute on the control.  This addresses #1611.

I also implemented unit tests for BFormCheckbox that roughly map to the tests for BFormRadiobutton.  I, of course, added some testing around the indeterminate model, and while I was at it some tests around the primary model.

While writing the tests, I found and fixed a typo in ariaLabelledby (the 'b' in by was upper case).

## Small replication

A small replication or video walkthrough can help demonstrate the changes made. This is optional but can help observe the intended changes. A mentioned issue that contains a replication also works.

## PR checklist

- [ x ] Bugfix :bug: - `fix(...)`
- [ x ] Feature - `feat(...)`
- [ x ] Documentation update - `docs(...)`
- [ ] Other (please describe)

### DIFF
--- a/packages/bootstrap-vue-next/src/components/BFormCheckbox/BFormCheckbox.vue
+++ b/packages/bootstrap-vue-next/src/components/BFormCheckbox/BFormCheckbox.vue
@@ -17,7 +17,7 @@
       :value="value"
       :true-value="value"
       :false-value="uncheckedValue"
-      :indeterminate="indeterminate"
+      :indeterminate="indeterminateBoolean"
       @change="modelValue !== undefined && emit('change', modelValue)"
       @input="modelValue !== undefined && emit('input', modelValue)"
     />
@@ -50,6 +50,7 @@ const props = withDefaults(
     disabled?: Booleanish
     form?: string
     id?: string
+    indeterminate?: Booleanish
     inline?: Booleanish
     modelValue?: CheckboxValue | readonly CheckboxValue[]
     name?: string
@@ -71,6 +72,7 @@ const props = withDefaults(
     disabled: false,
     form: undefined,
     id: undefined,
+    indeterminate: undefined,
     inline: false,
     modelValue: undefined,
     name: undefined,
@@ -88,6 +90,7 @@ const emit = defineEmits<{
   'change': [value: CheckboxValue | CheckboxValue[]]
   'input': [value: CheckboxValue | CheckboxValue[]]
   'update:modelValue': [value: CheckboxValue | CheckboxValue[]]
+  'update:indeterminate': [value: boolean]
 }>()
 
 const slots = defineSlots<{
@@ -96,10 +99,11 @@ const slots = defineSlots<{
 }>()
 
 const modelValue = useVModel(props, 'modelValue', emit, {passive: true})
-const indeterminate = defineModel('indeterminate', {default: false})
+const indeterminate = useVModel(props, 'indeterminate', emit, {passive: false})
 
 const computedId = useId(() => props.id, 'form-check')
 
+const indeterminateBoolean = useBooleanish(() => props.indeterminate)
 const autofocusBoolean = useBooleanish(() => props.autofocus)
 const plainBoolean = useBooleanish(() => props.plain)
 const buttonBoolean = useBooleanish(() => props.button)
--- a/packages/bootstrap-vue-next/src/components/BFormCheckbox/BFormCheckbox.vue
+++ b/packages/bootstrap-vue-next/src/components/BFormCheckbox/BFormCheckbox.vue
@@ -72,7 +72,7 @@ const props = withDefaults(
     disabled: false,
     form: undefined,
     id: undefined,
-    indeterminate: undefined,
+    indeterminate: false,
     inline: false,
     modelValue: undefined,
     name: undefined,
@@ -99,7 +99,7 @@ const slots = defineSlots<{
 }>()
 
 const modelValue = useVModel(props, 'modelValue', emit, {passive: true})
-const indeterminate = useVModel(props, 'indeterminate', emit, {passive: false})
+const indeterminate = useVModel(props, 'indeterminate', emit, {passive: true})
 
 const computedId = useId(() => props.id, 'form-check')
 
--- a/packages/bootstrap-vue-next/src/components/BFormCheckbox/BFormCheckbox.vue
+++ b/packages/bootstrap-vue-next/src/components/BFormCheckbox/BFormCheckbox.vue
@@ -12,7 +12,7 @@
       :name="name || parentData?.name.value"
       :form="form || parentData?.form.value"
       :aria-label="ariaLabel"
-      :aria-labelledby="ariaLabelledBy"
+      :aria-labelledby="ariaLabelledby"
       :aria-required="computedRequired || undefined"
       :value="value"
       :true-value="value"
@@ -42,7 +42,7 @@ defineOptions({
 const props = withDefaults(
   defineProps<{
     ariaLabel?: string
-    ariaLabelledBy?: string
+    ariaLabelledby?: string
     autofocus?: Booleanish
     button?: Booleanish
     buttonGroup?: Booleanish
@@ -64,7 +64,7 @@ const props = withDefaults(
   }>(),
   {
     ariaLabel: undefined,
-    ariaLabelledBy: undefined,
+    ariaLabelledby: undefined,
     autofocus: false,
     button: false,
     buttonGroup: false,
--- a/packages/bootstrap-vue-next/src/components/BFormCheckbox/BFormCheckbox.vue
+++ b/packages/bootstrap-vue-next/src/components/BFormCheckbox/BFormCheckbox.vue
@@ -17,7 +17,7 @@
       :value="value"
       :true-value="value"
       :false-value="uncheckedValue"
-      :indeterminate="indeterminateBoolean"
+      :indeterminate="indeterminate"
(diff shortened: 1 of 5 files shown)


packages/bootstrap-vue-next/src/components/BFormCheckbox/BFormCheckbox.vue
```diff
@@ -98,10 +96,10 @@ const slots = defineSlots<{
 }>()
 
 const modelValue = useVModel(props, 'modelValue', emit, {passive: true})
```
dwgray, line 98: I attempted to change the definition of the main model to use defineModel:
```
const modelValue = defineModel<CheckboxValue | readonly CheckboxValue[]>()
```
And then removed the explicit prop and emit declaration.

It worked, but the typescript definition for the event didn't propagate appropriately - when used like this:
```
@update:model-value="(event) => fired('update:modelValue', event)"
```
I get "Parameter 'event' implicitly has an 'any' type.ts(7006)" for event

Any thoughts?  I didn't see any open issues on vue.js on this.

packages/bootstrap-vue-next/src/components/BFormCheckbox/form-checkbox.spec.ts
```diff
@@ -0,0 +1,768 @@
...
+      const wrapper = mount(BFormCheckbox, {
+        slots: {default: 'foobar'},
+      })
+      const $label = wrapper.get('label')
+      expect($label.text()).toBe('foobar')
+    })
+  })
+
+  describe('model behavior', () => {
+    it('default unchecked checkbox emits input===false event when clicked', async () => {
```
dwgray, line 573: It looks like the input event is fired before the model is changed.  I don't know if this is expected behavior. If not, I'm happy to dig into it.  I don't know how urgent it is, since new code will use the update:modelValue event, which does fire after the model has changed.  I guess it would be somewhat important as a compatibility issue.

packages/bootstrap-vue-next/src/components/BFormCheckbox/BFormCheckbox.vue
VividLemon: The main issue for defineModel is the lack of "passive: true". The purpose of being passive is it uses an internal ref and watcher to update.

This is what functionality you need when you have optional v models. As without an internal system, if the user doesn't bind a value for vmodel, it will not work -- as there is no variable to be updated.

This is why I'm using the vueuse alternative. Even if not using passive, it just make sense to keep things consistent.

Maybe define model has this support, I don't know. I haven't done any research on

packages/bootstrap-vue-next/src/components/BFormCheckbox/form-checkbox.spec.ts
VividLemon: I feel like its not super important as input and changed will be converted to the native variants

packages/bootstrap-vue-next/src/components/BFormCheckbox/BFormCheckbox.vue
VividLemon: However the issue is likely a vue issue. 

It sounds like defineModel has some capability for default values. I'm not sure if it works right though. A good way to test is with the carousel, since its often for that component to not have a value bound.

It also raises some concerns about some other input elements not declaring themselves as passive and whether or not they work without being bounded

dwgray: Okay, thanks for the clear explanation.  I was struggling to understand the intent behind the passive flag in useVModel.  I moved the indeterminate model to use that pattern for consistency.

packages/bootstrap-vue-next/src/components/BFormCheckbox/BFormCheckbox.vue
```diff
@@ -98,6 +99,7 @@ const slots = defineSlots<{
...
+const indeterminate = useVModel(props, 'indeterminate', emit, {passive: false})
```
VividLemon, line 102: Passive false is redundant. It's the default. You can remove the options object.

packages/bootstrap-vue-next/src/components/BFormCheckbox/BFormCheckbox.vue
```diff
@@ -90,6 +90,7 @@ const emit = defineEmits<{
   'change': [value: CheckboxValue | CheckboxValue[]]
   'input': [value: CheckboxValue | CheckboxValue[]]
   'update:modelValue': [value: CheckboxValue | CheckboxValue[]]
+  'update:indeterminate': [value: boolean]
```
VividLemon, line 93: GH is kinda weird in that you can comment on untouched lines...

Anyways, the default value of bootstrap vue is that the indeterminate prop is set to false. That should probably be fixed

VividLemon (changes requested):
Two small change s but other than that lgtm. A rather simple fix 😆 tbh I thought it was going to be more which is why I never really touched it. Ty

packages/bootstrap-vue-next/src/components/BFormCheckbox/BFormCheckbox.vue
dwgray: I think you meant to comment on the default value (line 75) but I'm seeing the comment on the first changed line after (line 93) - so I agree GH is kinda weird.

And good catch.  I was looking right at those docs and missed that the default is false for BSV.  I've changed that in the code and in the docs.

packages/bootstrap-vue-next/src/components/BFormCheckbox/BFormCheckbox.vue
dwgray: I intended passive to be true.  Based on your previous comments, I'd like to have indeterminate behavior work even if I'm not binding to a model value (e.g. if I just want to pass in :indeterminate="true" initially). But I think I may still be confused.  If you think I should remove the passive I'll do that.  The behavior I'm most interested in is encoded in the tests and works with either value.

dwgray:
> Two small change s but other than that lgtm. A rather simple fix 😆 tbh I thought it was going to be more which is why I never really touched it. Ty

Yeah, the actual fix was pretty simple.  But it was good to get this part of the code under test and get my head wrapped around it.

packages/bootstrap-vue-next/src/components/BFormCheckbox/BFormCheckbox.vue
VividLemon: I do think you are misunderstanding the intention. The best example I have is actually the carousel component. Review these following

Notice how this one works without binding anything. This is the power of passive true. It works without `v-model="someRef"
```html
<BCarousel indicators>
  <BCarouselSlide img-src="https://picsum.photos/1024/480/?image=4" />
  <BCarouselSlide img-src="https://picsum.photos/1024/480/?image=5" />
  <BCarouselSlide img-src="https://picsum.photos/1024/480/?image=6" />
</BCarousel>
```

This also works, but what should not work is if you click the indicators (it should always show the #2 slide. It is forever static)
```html
<BCarousel indicators :model-value="2">
  <BCarouselSlide img-src="https://picsum.photos/1024/480/?image=4" />
  <BCarouselSlide img-src="https://picsum.photos/1024/480/?image=5" />
  <BCarouselSlide img-src="https://picsum.photos/1024/480/?image=6" />
</BCarousel>
```

Then finally, a typical vmodel, where the internal system changes, in addition to reacting to external changes
```html
<BCarousel indicators v-model="someRef">
  <BCarouselSlide img-src="https://picsum.photos/1024/480/?image=4" />
  <BCarouselSlide img-src="https://picsum.photos/1024/480/?image=5" />
  <BCarouselSlide img-src="https://picsum.photos/1024/480/?image=6" />
</BCarousel>
```

This is what the passive value does. Its saying that the value will still work, even if it is optional. Whereas in the indeterminate state, this doesn't seem to need to be passive as the checkbox can't mutate its indeterminate state, and doesn't really make sense to want that functionality, but not keep track of the state of it

dwgray: Oh, thank you.  That clicked for me.  I've removed the passive flag from the indeterminate state.

VividLemon (approved):

VividLemon:
If you're up for a slightly more challenging ticket,

https://github.com/bootstrap-vue-next/bootstrap-vue-next/pull/1686

Was implemented originally made, but has conflicts, plus I had some revieiwal issues.

I had created a new PR to fix up some of those issues with some of my other stuff 

https://github.com/bootstrap-vue-next/bootstrap-vue-next/pull/1723

It would be a massive help if that was fixed. As I only have 2 tasks before 0.16 is released, and that is one of them... The logic I had was messed up, the conditions for showing the numbers was wrong. If you don't respond I'll just assume it's a no-go but figured I'd ask 🤣

dwgray:
@VividLemon Yes  - I'd be happy to help.  I've been looking for an opportunity to get my head wrapped around createReusableTemplate.  I should have a big enough block of time to make some progress on this tomorrow.

I don't think I can contribute to your PR directly in Github.  Should I just create a PR with a working version of the BPagination fix and you can pull from that?

VividLemon:
> I don't think I can contribute to your PR directly in Github

There's probably something you could do regarding pushing to my remote... But idk if gh would try to pr it or just reject

Regardless, the easiest way would just be to make a new PR with both of ours as inspiration

PR #1726, opened 2024-01-22
fix(BPagination): re-implemented the missing first-page and last-page

dwgray:
# Describe the PR

Fix several issues with the BPagination
- Re-implement the missing first-page and last-page properties.  This pulled heavily from the work by @MewesK in #1686 and @VividLemon in #1723. 
- Fix the issue where page #2 and page # total-1 would disappear under some circumstances
- Slot scoped properties for the 'textual' (first, prev, next, last) slots are now implemented (these properties were documented in both this project and bsv-legacy)
- Fixed the hideEllipsis property to _always_ hide both ellipsis
- Fixed the aria labels for the prev, next and last buttons (they were all first).
- Changed the limit property from `number` to `Numberish`
- Added tests to validate that the # of page buttons shown are consistent taking into account the ellipsis buttons and firstNumber & lastNumber properties
    - These tests can be tuned to the # of scenarios that they cover.  For check-in tests they're covering a small but useful area, but for developing, the parameters can be expanded for greater coverage

In order to get all of the above working, I ended up rewriting the core logic that lays out the pagination control.  I believe the new system is more straightforward and, therefore, should be easier to maintain in the future.

## Small replication

You can see the broken behavior in the docs:
https://bootstrap-vue-next.github.io/bootstrap-vue-next/docs/components/pagination.html#goto-first-last-button-type

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

### DIFF
--- a/packages/bootstrap-vue-next/src/components/BPagination/BPagination.vue
+++ b/packages/bootstrap-vue-next/src/components/BPagination/BPagination.vue
@@ -15,6 +15,24 @@
         </component>
       </li>
     </ReusableButtuon.define>
+
+    <ReusablePageButton.define v-slot="{page, button, li, clickHandler}">
+      <li v-bind="li">
+        <component v-bind="button" :is="button.is" @click="clickHandler">
+          <slot
+            name="page"
+            :active="isActivePage(page)"
+            :disabled="checkDisabled(page)"
+            :page="page"
+            :index="page - 1"
+            :content="page"
+          >
+            {{ page }}
+          </slot>
+        </component>
+      </li>
+    </ReusablePageButton.define>
+
     <ReusableEllipsis.define>
       <li v-bind="ellipsisProps.li">
         <span v-bind="ellipsisProps.span">
@@ -31,58 +49,34 @@
     />
     <ReusableButtuon.reuse v-bind="prevButtonProps" />
 
+    <ReusablePageButton.reuse
+      v-if="firstNumberBoolean && pages[0] && pages[0].number !== 1"
+      v-bind="firstNumberButtonProps"
+    />
+
     <ReusableEllipsis.reuse v-if="showFirstDots" />
 
-    <li
+    <ReusablePageButton.reuse
       v-for="page in pages"
       :key="`page-${page.number}`"
-      :class="[
-        'page-item',
-        {
-          'disabled': disabledBoolean,
-          'active': isActivePage(page.number),
-          'flex-fill': computedFill,
-          'd-flex': computedFill && !disabledBoolean,
-        },
-        pageClass,
-      ]"
-      role="presentation"
-      :aria-hidden="disabledBoolean || undefined"
-    >
-      <component
-        :is="disabledBoolean ? 'span' : 'button'"
-        :key="`page-${page.number}`"
-        :class="['page-link', {'flex-grow-1': !disabledBoolean && computedFill}]"
-        :aria-controls="ariaControls || undefined"
-        :aria-disabled="disabledBoolean ? true : undefined"
-        :aria-label="labelPage ? `${labelPage} ${page.number}` : undefined"
-        :aria-posinset="page.number"
-        :aria-checked="isActivePage(page.number)"
-        :aria-setsize="numberOfPages"
-        role="menuitemradio"
-        :type="disabledBoolean ? undefined : 'button'"
-        :tabindex="getTabIndex(page.number)"
-        @click="pageClick($event, page.number)"
-      >
-        <slot
-          name="page"
-          :active="isActivePage(page.number)"
-          :disabled="disabledBoolean"
-          :page="page.number"
-          :index="page.number - 1"
-          :content="page.number"
-        >
-          {{ page.number }}
-        </slot>
-      </component>
-    </li>
+      v-bind="getPageButtonProps({page: page.number, dis: disabledBoolean})"
+    />
 
     <ReusableEllipsis.reuse v-if="showLastDots" />
+
+    <ReusablePageButton.reuse
+      v-if="
+        lastNumberBoolean &&
+        pages[pages.length - 1] &&
+        pages[pages.length - 1].number !== numberOfPages
+      "
+      v-bind="lastNumberButtonProps"
+    />
 
     <ReusableButtuon.reuse v-bind="nextButtonProps" />
     <ReusableButtuon.reuse
       v-if="!lastNumberBoolean && !hideGotoEndButtonsBoolean"
-      v-bind="lastEndButtonProps"
+      v-bind="lastButtonProps"
     />
   </ul>
 </template>
@@ -209,7 +203,7 @@ const prevDisabled = computed(() => checkDisabled(modelValueNumber.value - 1))
 const lastDisabled = computed(() => checkDisabled(numberOfPages.value))
 const nextDisabled = computed(() => checkDisabled(modelValueNumber.value + 1))
 
-const getEndButtonProps = ({
+const getButtonProps = ({
   classVal,
   clickHandler,
   dis,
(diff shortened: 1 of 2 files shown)


dwgray:
@VividLemon I ran into a couple of other issues while implementing this, but I now have a working solution.  I'm going to put a couple of notes inline for things that I'm not sure about (I believe they are style rather than functional issues).

packages/bootstrap-vue-next/src/components/BPagination/BPagination.vue
```diff
@@ -208,79 +193,151 @@ const prevDisabled = computed(() => checkDisabled(modelValueNumber.value - 1))
 const lastDisabled = computed(() => checkDisabled(numberOfPages.value))
 const nextDisabled = computed(() => checkDisabled(modelValueNumber.value + 1))
 
-const getEndButtonProps = ({
+const getBaseButtonProps = ({
```
dwgray, line 196: `getButtonProps` and `getPageButtonProps` ended up being similar enough that I merged them.  But that ended up with a `getBaseButtonProps` with an unwieldy number of parameters.  I still think that's better than the duplicate code but I feel like there should be a better way to do this.  Let me know if you have any ideas, or if you'd prefer me to split them back out.

VividLemon (changes requested):
There exists a logic error, this is something I was trying to deal with, and was the reason why I couldn't dedicate extra time to the issue

See this:

```html
<template>
  <h6>Goto first button number</h6>
  <BPagination v-model="ex2CurrentPage" :total-rows="ex2Rows" :per-page="ex2PerPage" first-number />

  <h6>Goto last button number</h6>
  <BPagination v-model="ex2CurrentPage" :total-rows="ex2Rows" :per-page="ex2PerPage" last-number />

  <h6>Goto first and last button number</h6>
  <BPagination
    v-model="ex2CurrentPage"
    :total-rows="ex2Rows"
    :per-page="ex2PerPage"
    first-number
    last-number
  />
</template>

<script setup lang="ts">
const ex2CurrentPage = ref(5)
const ex2PerPage = ref(1)
const ex2Rows = ref(100)
</script>
```

The issue exists here, when the button isn't visible, but there is no automatic number placed for 99

image

The value should look like this

image

A perplexing issue for sure to get it to fit into the current structure, that code is a bit strange looking. Although, a perfect fit for TTD given its simplicity. It manages to get a more confusing when you modify the `limit` prop

dwgray:
> There exists a logic error, this is something I was trying to deal with, and was the reason why I couldn't dedicate extra time to the issue
> 

Got it.  I'll take a look at the logic error and keep the limit property in mind.  I was hoping to get some test coverage around this anyway, so I'll make sure to at least get this case.

dwgray:
> > There exists a logic error, this is something I was trying to deal with, and was the reason why I couldn't dedicate extra time to the issue
> 
> Got it. I'll take a look at the logic error and keep the limit property in mind. I was hoping to get some test coverage around this anyway, so I'll make sure to at least get this case.

Okay, I had some time this week to dive into this and have made some progress.  But I've hit a wall.

I started by writing piecemeal unit tests covering the breaking cases and attempting to tweak the existing code to work.  This turned into a game of whack-a-mole, so I backed out.

What I landed on was writing tests that test that several invariants are true about the rendered pagination control.  
- I check that the number 2 button is there when the first ellipsis is hidden
- And that the max-1 button is there when the second ellipsis is hidden (which is the case you pointed out).  I believe I have those working consistently.  
- The other invariant that I was trying to achieve was that the number of page buttons is consistent - basically that the # of page buttons + the number of ellipses - (firstNumber + lastNumber) == limit.

With those tests in place and validated against my ad-hoc unit tests, I attempted to rework the numberOfLinks algorithm in a way that I could get my head wrapped around.  I thought I had mostly- I had all my tests passing against # of pages == 100 and limits of 3 - 20 (and that is against the exact match). But before calling success, I expanded the tests to try a range of # of pages (size).  There is a flaw in my logic somewhere that I am unable to track down, apparently related to cases where the limit is close to the size.

I don't see any pattern in the madness, and can't see the flaw in my logic, so I'm stuck.  I've attached 
[errors.txt](https://github.com/bootstrap-vue-next/bootstrap-vue-next/files/14060492/errors.txt) a csv with the failures from one of the larger runs - maybe someone else can see a pattern (or find the flaw in my logic)

The good news is - the most egregious issue (the one you originally pointed out) is verifiably fixed, along with several other issues I mention in the description. And the one that I'm stuck on is significantly improved from the state of the checked in code.  I ran my test code against the checked-in code and had a total of 280 failures 
[old-code.txt](https://github.com/bootstrap-vue-next/bootstrap-vue-next/files/14060510/old-code.txt)
while the new code has only 47 
[new-code.txt](https://github.com/bootstrap-vue-next/bootstrap-vue-next/files/14060514/new-code.txt)
.  These runs were checking for off by more than one on runs of about 10,000.

In any case, I'm a bit stuck.  I'll come back to it when I've got some more time if no-one else has a chance to take a look, but I'm hoping that someone else's perspective will show me where I'm missing something.  Also, assuming my analysis is right, this is substantially better than what is already there.

dwgray:
> > > There exists a logic error, this is something I was trying to deal with, and was the reason why I couldn't dedicate extra time to the issue
> > 
> > 
> > Got it. I'll take a look at the logic error and keep the limit property in mind. I was hoping to get some test coverage around this anyway, so I'll make sure to at least get this case.
> 
> Okay, I had some time this week to dive into this and have made some progress. But I've hit a wall.
> 
I let this set for a bit and decided to take a different approach to the basic logic for the component.  It's looking good early on, so if no one else has taken a look at this yet, please hold off for a bit.  I'm hoping to get a more robust solution to this up this weekend.

packages/bootstrap-vue-next/src/components/BPagination/BPagination.vue
```diff
@@ -467,11 +383,111 @@ watch(pagination, (oldValue, newValue) => {
   }
 })
 
-const pages = computed(() =>
-  Array.from({length: numberOfLinks.value}, (_, index) => ({
-    number: startNumber.value + index,
-  }))
-)
+const buttons = computed(() => {
```
dwgray, line 386: I failed to get my head wrapped around the existing algorithm.  I believe the issue was the interactions between the startNumber, numberOfLinks, showStartDots, and showEndDots functions that spread the logic about how to layout the control between them in a hard to grasp way.

Instead, I made a master function that computes the layout of the control.  This is a larger function than I'd like, but it's pretty linear and (at least to me) straightforward.  I've attempted to comment it in a way that will make it reasonable for others to wrap their heads around.  I've also created local variables for a bunch of the props/computed values that are used throughout the function.  That makes it more readable in my opinion, but I'm happy to rework this if there are stylistic objections from the owners.  Alos happy to break this function down if anyone has ideas on a clean way to do that.

dwgray:
@VividLemon I believe I have a robust working implementation of this control.  I had to take a step back and pretty much re-write the core logic to make this work, but once I did that, things fell into place quite easily (I spent considerably less time on the rewrite than I did trying to tweak the original algorithm).  Let me know if you have any issues with it.  I'm still ramping up on this codebase, so I'm happy to take suggestions to improve this code or make it line up more with the rest of the project.

packages/bootstrap-vue-next/src/components/BPagination/BPagination.vue
VividLemon: I would have likely done the same. The original algorithm was confusing

packages/bootstrap-vue-next/src/components/BPagination/BPagination.vue
```diff
@@ -467,11 +383,111 @@ watch(pagination, (oldValue, newValue) => {
...
+      buttons[index + 2 - hideEndButtons] = start + index
+    }
+
+    if (!hideEllipsis) {
+      buttons[2] = ELLIPSIS_BUTTON
+      buttons[buttons.length - 3] = ELLIPSIS_BUTTON
+    }
+  }
+
+  // Enable sanity check for debugging purposes
```
VividLemon, line 480: lol

VividLemon (approved):

Issue #1748, opened 2024-02-05
Inconsistent BAvatar Size Prop


ijay-bm:
### Describe the bug

In the documentation for BAvatar's sizing (https://bootstrap-vue-next.github.io/bootstrap-vue-next/docs/components/avatar.html#sizing), various usages of the `size` prop are shown. However, they are inconsistent when rendered. 

![image](https://github.com/bootstrap-vue-next/bootstrap-vue-next/assets/81517464/c536289c-ef4a-4270-a601-b02819036b74)

### Reproduction

https://stackblitz.com/~/edit/github-wapqys?file=src%2Fcomponents%2FComp.vue%2Csrc%2FApp.vue

### Used Package Manager

pnpm

### Validations

- [X] Have tested with the latest version. This is still alpha version and sometime things change rapidly.
- [X] Follow our [Code of Conduct](https://github.com/bootstrap-vue-next/bootstrap-vue-next/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guide](https://github.com/bootstrap-vue-next/bootstrap-vue-next/blob/main/CONTRIBUTING.md).
- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A, please open a GitHub Discussion instead.
- [X] The provided reproduction is a [minimal reproducible](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

lqbach:
@VividLemon can be closed with #1760

PR #1760, opened 2024-02-11
fix(BAvatar): replace useToNumber in useAvatarSize composable

lqbach:
# Describe the PR
Passing a string value that ends in `em` or`rem` to the `size` prop `BAvatar` converts them to `px` instead of their expected values.

The `BAvatar` component uses the `size` prop to change its size. The component applies styles based on the `size` prop. This prop does not properly parse string values that end in `em` and `rem` because of the `useAvatarSize` composable. This composable uses [VueUse's `useToNumber`](https://vueuse.org/shared/useToNumber/) method to determine if the `size` prop is a numeric value or not. [`useToNumber` source code](https://github.com/vueuse/vueuse/blob/4f84d63730e18d8408a9f05f72809f3de4a9315e/packages/shared/useToNumber/index.ts#L12) uses the [`parseFloat` ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseFloat) method, which according to the docs:

>However, parseFloat() is more lenient than Number() because it ignores trailing invalid characters, which would cause Number() to return NaN

This means trailing characters like `36rem` or `15em` will be parsed as `36` and `15` respectively, resulting in a valid number and having `px` appended to these characters.

This PR will add a new constant to the regex file so that we can a regex test to check values before they are parsed. We will then use the `Number` constructor and convert the value to a number once it passes the test. The regex test will also prevent values that have non-decimal literals with 0x, 0b, or 0o prefixes.

**Related: #1748** 

## Small replication

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

### DIFF
--- a/packages/bootstrap-vue-next/src/components/BAvatar/avatar.spec.ts
+++ b/packages/bootstrap-vue-next/src/components/BAvatar/avatar.spec.ts
@@ -43,5 +43,37 @@ describe('avatar', () => {
     expect(wrapper.element.tagName).toBe('SPAN')
   })
 
+  it('`size` prop should work as expected', () => {
+    const wrapper1 = mount(BAvatar)
+    expect(wrapper1.attributes('style')).toEqual(undefined)
+
+    const wrapper2 = mount(BAvatar, {props: {size: 'sm'}})
+    expect(wrapper2.attributes('style')).toEqual('')
+    expect(wrapper2.classes()).toContain('b-avatar-sm')
+
+    const wrapper3 = mount(BAvatar, {props: {size: 'md'}})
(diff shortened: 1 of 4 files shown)


packages/bootstrap-vue-next/src/components/BAvatar/avatar.spec.ts
```diff
@@ -43,5 +43,37 @@ describe('avatar', () => {
     expect(wrapper.element.tagName).toBe('SPAN')
   })
 
+  it('`size` prop should work as expected', () => {
```
VividLemon, line 46: This display name

VividLemon (approved):

Issue #1758, opened 2024-02-10
[Vue warn]: injection "Symbol(modalManagerPlugin)" not found. 


bcorey85:
### Describe the bug

- BModal components cause a Vue injection warning on fresh install and do not open via `useModal().show`

Steps to Reproduce:
1. npm install
2. npm run dev
3. Observe console warning - [Vue warn]: injection "Symbol(modalManagerPlugin)" not found. 
4. Try opening modal

### Reproduction

https://github.com/bcorey85/bootstrap-vue-next-modal-debug

### Used Package Manager

npm

### Validations

- [X] Have tested with the latest version. This is still alpha version and sometime things change rapidly.
- [X] Follow our [Code of Conduct](https://github.com/bootstrap-vue-next/bootstrap-vue-next/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guide](https://github.com/bootstrap-vue-next/bootstrap-vue-next/blob/main/CONTRIBUTING.md).
- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A, please open a GitHub Discussion instead.
- [X] The provided reproduction is a [minimal reproducible](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

ErikVerheul:
Have the same problem when rendering on the client (not NUXT) for every instance of BModel. Revert to version 0.15.5 which does not have this issue.

bcorey85:
@VividLemon This PR fixed the issue in our core app, but unit tests involving modals are still throwing `injection "Symbol(modalManagerPlugin)" not found` errors.

I updated the original example repo [https://github.com/bcorey85/bootstrap-vue-next-modal-debug](url) with an example: 

1. npm install
2. npm test
3. observe `[Vue warn]: injection "Symbol(modalManagerPlugin)" not found.` thrown from `BModal.test.ts`

alekswebnet:
For me, it has been resolved by installing required plugin:

`import bootstrapVue from 'bootstrap-vue-next'`

`app.use(
    bootstrapVue({
      plugins: {
        modalController: true
      }
    })
  )`

OliveLeb:
Had the warning too and what worked for me was to follow installation instructions from the [docs](https://bootstrap-vue-next.github.io/bootstrap-vue-next/docs.html) 🤡 
```
// main.js/ts
[...]

PR #1761, opened 2024-02-11
fix(nuxt): optimize `bootstrap-vue-next`

danielroe:
# Describe the PR

resolves https://github.com/bootstrap-vue-next/bootstrap-vue-next/issues/1758
see https://github.com/nuxt/nuxt/issues/25735

The issue is not that the plugin is not being called.

CleanShot 2024-02-11 at 20 46 05@2x

The problem is that Vite optimises the dependencies and so there are effectively two versions of bootstrap vue in your project, one registered and one not registered. (This is why it works when you remove the `node_modules` - you can instead just remove `node_modules/.cache` to reliably reproduce.)

(This is why the problem is only on first load.)

You can read more about this process [in the Vite docs](https://vitejs.dev/config/dep-optimization-options.html#optimizedeps-include).

This resolves the issue by ensuring `bootstrap-vue-next` is optimised, though note that this will require users who use pnpm without shamefully-hoisting their dependencies to install `bootstrap-vue-next` as a top-level dependency. (Vite will otherwise complain it can't find it.)

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

### DIFF
--- a/packages/nuxt/src/module.ts
+++ b/packages/nuxt/src/module.ts
@@ -56,6 +56,10 @@ export default defineNuxtModule<ModuleOptions>({
 
     nuxt.options.css.push('bootstrap-vue-next/dist/bootstrap-vue-next.css')
 
+    nuxt.options.vite.optimizeDeps = nuxt.options.vite.optimizeDeps || {}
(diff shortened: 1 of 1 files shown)


VividLemon (approved):
Tyvm. This is very interesting to learn about

Issue #1757, opened 2024-02-09
BTooltip does not stay open when being hovered with the mouse


sceee:
### Describe the bug

When creating a BTooltip with interactive content (e.g. a link), the BTooltip needs to stay open while being hovered by the mouse so that the user can interact with the content (e.g. click the link).

However, at the moment (bootstrap-vue-next 0.16.1) the tooltip hides as soon as you move away from the original hover element making it impossible to interact with the content (e.g. click the link).

**Expected**
* BTooltip stays open while being hovered with the mouse

_Update:_
Actually I just noticed this is a difference between `bootstrap-vue` and `bootstrap` (5.3):
* In `bootstrap-vue` (see https://bootstrap-vue.org/docs/components/tooltip#noninteractive-tooltips ) the tooltips are interactive by default and stay open on hover
* But in bootstrap 5.3 (see https://getbootstrap.com/docs/5.3/components/tooltips/#custom-tooltips ) it is closed when hovering
So it might indeed not be a bug but rather a feature request to be able to keep it open on hover (= have the possibility to keep the BootstrapVue interactive behavior).

_Update 2:_
It can still be made interactive (=stay open), see my [comment below](https://github.com/bootstrap-vue-next/bootstrap-vue-next/issues/1757#issuecomment-1935707454) for an improvement suggestion.

### Reproduction

https://stackblitz.com/edit/github-ydahcf?file=src%2Fcomponents%2FComp.vue

### Used Package Manager

npm

### Validations

- [X] Have tested with the latest version. This is still alpha version and sometime things change rapidly.
- [X] Follow our [Code of Conduct](https://github.com/bootstrap-vue-next/bootstrap-vue-next/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guide](https://github.com/bootstrap-vue-next/bootstrap-vue-next/blob/main/CONTRIBUTING.md).
- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A, please open a GitHub Discussion instead.
- [X] The provided reproduction is a [minimal reproducible](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

sceee:
I now noticed it works if the `BTooltip` gets `:noninteractive="false"` set. Then it stays open on hover.

So this works to keep the BTooltip open on hovering the tooltip:
```
[...]

PR #1768, opened 2024-02-13
feat(BTooltip): add prop interactive

xvaara:
# Describe the PR

in bsv2 the prop for tooltip is `noninteractive` which defaults to false. I think booleans which default to false aren't great for us. In bs5 tooltips aren't interactive at all. So I think this is a middle ground.

## Small replication

## PR checklist

- [x] Feature - `feat(...)`
- [ ] Other (please describe)

### DIFF
--- a/packages/bootstrap-vue-next/src/components/BTooltip.vue
+++ b/packages/bootstrap-vue-next/src/components/BTooltip.vue
@@ -1,5 +1,5 @@
 <template>
-  <BPopover ref="popover" tooltip v-bind="$props">
+  <BPopover ref="popover" tooltip v-bind="computedProps">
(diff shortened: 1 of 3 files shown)


VividLemon (changes requested):
Documentation for these features seems like a necessity. I'm not sure what interactive/noninteractive props are doing here.

xvaara:
There aren't any documentation for these in any case and I don't have the time to write anything from scratch at the moment. you want me to add comment on what they do in the code?

It's a prop that can user interact with the tooltip/popover (ie. is it closed when mouseout of target or do we wait if the mouse goes into the popover element and keep open.)

This is regarding #1757

VividLemon (approved):

PR #1771, opened 2024-02-14
feat(BTable): Enable more sorting options to get closer to BSV parity

dwgray:
# Describe the PR

There were a few issues with table sorting compared to BSV:
- The sortDesc flag wasn't respected
- The sortByFormatted field wasn't used
    - And the sortByFormatted field didn't allow passing in a formatter function

I also added tests to cover these cases and updated the type documentation.

## Small replication

Any sorted table example will replicate these issues.  See the tests in this PR for concrete examples.

## PR checklist

- [x] Feature - `feat(...)`
- [ ] Other (please describe)


packages/bootstrap-vue-next/src/utils/formatItem.ts
```diff
@@ -0,0 +1,11 @@
+import type {TableFieldFormatter, TableItem} from 'src/types'
```
VividLemon, line 1: You won't be able to import from 'src/types'. It needs to be a relative path (types generation for library consumption breaks, even though it appears to work correctly here)

VividLemon (changes requested):

packages/bootstrap-vue-next/src/utils/formatItem.ts
dwgray: Okay, thanks.  I changed this to a relative path.  Is there a way for me to check type generation in the future (other than just knowing to use relative paths)?

VividLemon (dismissed):

packages/bootstrap-vue-next/src/components/BTable/BTable.vue
```diff
@@ -401,20 +402,28 @@ const computedItems = computed<readonly TableItem<T>[]>(() => {
...
+          }
+        }
+        return typeof val === 'object' && val !== null ? JSON.stringify(val) : val?.toString() ?? ''
+      }
+      return (
+        realVal(a).localeCompare(realVal(b), props.sortCompareLocale, props.sortCompareOptions) *
+        (sortDescBoolean.value ? -1 : 1)
       )
     })
-    return sortDescBoolean.value && props.sortCompare === undefined ? sorted.reverse() : sorted
```
VividLemon, line 417: I assume this functionality still is working in `return (
        realVal(a).localeCompare(realVal(b), props.sortCompareLocale, props.sortCompareOptions) *
        (sortDescBoolean.value ? -1 : 1)`

VividLemon (approved):

packages/bootstrap-vue-next/src/components/BTable/BTable.vue
dwgray: Yes - we had a case "conflicting" PRs that solved the same problem in ways that didn't _actually_ cause a conflict in the diffs.  My unit tests caught this. I left the way that I solved it in, as I think it's slightly cleaner (but I'm not a native JS developer, so if you'd prefer doing the explicit reverse at the end rather than sorting in the correct direction in the first place, I'll swap it out.

VividLemon (approved):

PR #1781, opened 2024-02-18
fix(docs): Enable Syntax Highlighting fixes #1738

### LINKED ISSUE #1738: Docs: Syntax highlighting is broken
### Describe the bug

Sometime fairly recently it looks like syntax highlighting was broken in the docs.  This appears to be true for all of the examples, the example below is an example I modified recently and I'm pretty sure it was highlighting correctly when I did that in #1725, but I probably hadn't merged #1723 when I did the testing on that, so that may be a candidate.

image

### Reproduction

See the above image.

### Used Package Manager

pnpm

### Validations

dwgray:
# Describe the PR

Get Syntax Highlighting working again in the documentation.  This fixes #1738 

I'm explicitly pulling in the `vp-code.css` as well as several other files that seem appropriate.  It's unclear to me why https://github.com/vuejs/vitepress/pull/3237 caused this issue, as the file had been created and consumed by the default theme a couple of years ago.  In any case, this fixes the issue.  I haven't nailed down what each of the additional files does and if we're currently consuming them, but they seem like reasonable functionality.  We are using `custom-block.css` in `badge.md` and leaving out that file makes the tip section just disappear from the docs. So while I would normally go for the minimal fix it seems like including all of these files makes sense in this case.

## Small replication

Syntax highlighting is broken in all the documentation.

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

### DIFF
--- a/apps/docs/.vitepress/theme/Layout.vue
+++ b/apps/docs/.vitepress/theme/Layout.vue
@@ -284,6 +284,10 @@ watch(
 @import 'vitepress/dist/client/theme-default/styles/vars.css';
 @import 'vitepress/dist/client/theme-default/styles/base.css';
 @import 'vitepress/dist/client/theme-default/styles/utils.css';
+@import 'vitepress/dist/client/theme-default/styles/components/custom-block.css';
+@import 'vitepress/dist/client/theme-default/styles/components/vp-code.css';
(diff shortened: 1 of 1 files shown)


VividLemon (approved):

VividLemon:
Tyvm

PR #1786, opened 2024-02-21
BTable check selection also using primaryKey attribute

makroxyz:
# Describe the PR

BTable ```selectable``` works fine but...

using BTable ```provider``` attribute with pagination will loose the row selected class when page is changed.

with this PR when the ```primaryKey``` attribute (already present but unused) is set, the selection is checked by comparing the TableItem primaryKey attribute

Old BootstrapVue has primaryKey attribute:
"Name of a table field that contains a guaranteed unique value per row. Needed for tbody transition support, and also speeds up table rendering"

I think it can be used also for other things like selection which is a new feature

### DIFF
--- a/packages/bootstrap-vue-next/src/components/BTable/BTable.vue
+++ b/packages/bootstrap-vue-next/src/components/BTable/BTable.vue
@@ -295,7 +295,20 @@ const selectedItemsSetUtilities = {
   For some reason, the reference of the object gets lost. However, when you use an actual ref(), it works just fine
   Getting the reference properly will fix all outstanding issues
   */
-  has: (item: Readonly<TableItem<T>>) => selectedItemsToSet.value.has(item),
+  has: (item: Readonly<TableItem<T>>) => {
+    if (props.primaryKey) {
+      let isSelected = false
+      const pkey = props.primaryKey as keyof TableItem<T>
+      selectedItemsToSet.value.forEach((selected: TableItem<T>) => {
+        if (selected[pkey] === item[pkey]) {
+          isSelected = true
+        }
+      })
+      return isSelected
+    }
+
+    return selectedItemsToSet.value.has(item)
+  },
 } as const
 
 /**
--- a/packages/bootstrap-vue-next/src/components/BTable/BTable.vue
+++ b/packages/bootstrap-vue-next/src/components/BTable/BTable.vue
@@ -296,19 +296,14 @@ const selectedItemsSetUtilities = {
   Getting the reference properly will fix all outstanding issues
   */
   has: (item: Readonly<TableItem<T>>) => {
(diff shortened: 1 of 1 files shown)


packages/bootstrap-vue-next/src/components/BTable/BTable.vue
```diff
@@ -295,7 +295,20 @@ const selectedItemsSetUtilities = {
...
+  has: (item: Readonly<TableItem<T>>) => {
+    if (props.primaryKey) {
+      let isSelected = false
+      const pkey = props.primaryKey as keyof TableItem<T>
+      selectedItemsToSet.value.forEach((selected: TableItem<T>) => {
+        if (selected[pkey] === item[pkey]) {
+          isSelected = true
+        }
+      })
+      return isSelected
```
VividLemon, line 307: There's probably an alternative way of writing this. Perhaps by using 

```ts
return selectedItemsToSet.value.some((selected) => selected[props.primaryKey] === items[pkey])
```
I think these two sets of code are the same 🤔

VividLemon (changes requested):

packages/bootstrap-vue-next/src/components/BTable/BTable.vue
makroxyz: some isn't available in Set TS object

VividLemon: In that case, this is probably more suitable -- as it terminates the loop early

```ts
for (const selected of selectedItemsToSet.value) {
  if (selected[pkey] === item[pkey]) {
    isSelected = true;
    break;
  }
}
```

makroxyz: Done!

packages/bootstrap-vue-next/src/components/BTable/BTable.vue
```diff
@@ -295,7 +295,21 @@ const selectedItemsSetUtilities = {
...
+        if (selected[pkey] === item[pkey]) {
+          isSelected = true
+          break
+        }
+      }
+      return isSelected
+    }
+
+    return selectedItemsToSet.value.has(item)
+  },
```
VividLemon, line 312:
```suggestion
  has: (item: Readonly<TableItem<T>>) => {
    if (props.primaryKey) {
      const pkey = props.primaryKey as keyof TableItem<T>
      for (const selected of selectedItemsToSet.value) {
        if (selected[pkey] === item[pkey]) {
          return true
        }
      }
      return false
    }

    return selectedItemsToSet.value.has(item)
  },
```

I didn't think about it before, but this could be more concise.

VividLemon (approved):

Issue #1779, opened 2024-02-16
Remove deprecated Bootstrap classes -- or even outdated classes!


VividLemon:
There are instances of using Bootstrap classes that are either deprecated or much worse, removed.

One instance of this is/was btn-block

VividLemon:
These items need to be removed and updated to reflect their bootstrap 5 capabilities.

sceee:
I've worked through https://getbootstrap.com/docs/5.3/migration/#v500 , I found references to the following breaking changes from bootstrap 5 in `bootstrap-vue-next` and think these need to be adressed:

SASS:

- [ ] Previously deprecated `hover-focus-active()` mixin was removed (https://getbootstrap.com/docs/5.3/migration/#sass)

Grid:

- [ ] There is no only `.order-1` to `.order-5` out of the box ( https://getbootstrap.com/docs/5.3/migration/#grid-updates ). `.order-6` is used at least in a `BCol` test

Forms:

- [ ] `.custom-range` (used in the docs) was renamed to `.form-range` (https://getbootstrap.com/docs/5.3/migration/#forms-1)
- [ ] `.input-group-append` and `.input-group-prepend` have been dropped (used in scss) (https://getbootstrap.com/docs/5.3/migration/#forms-1)

Components:

- [ ] `.badge-*` color classes have been dropped (used in the docs) (https://getbootstrap.com/docs/5.3/migration/#badges)
- [ ] `.card-deck` was dropped (used in scss and BCardGroup) (https://getbootstrap.com/docs/5.3/migration/#card)
- [ ] `.card-columns` was dropped (used in scss and BCardGroup) (https://getbootstrap.com/docs/5.3/migration/#card)
- [ ] `flip` option for Dropdown plugin was dropped in favor of native Popper configuration (used in BDropdown) (https://getbootstrap.com/docs/5.3/migration/#dropdowns-1)

Utilities:

- [ ] Renamed `.ml-*` and `.mr-*` to `.ms-*` and `.me-*` (used in the docs and playground) (https://getbootstrap.com/docs/5.3/migration/#utilities-3)
- [ ] Renamed `.sr-only` to `.visually-hidden` (used in docs and playground) (https://getbootstrap.com/docs/5.3/migration/#utilities-3)

I _hope_ I detected most. It's sometimes pretty hard to detect these as you cannot simply search for the deprecated/removed classes as they are sometimes joined together from different strings/props.

[...]

PR #1815, opened 2024-03-15
fix(Bootstrap): remove deprecated classes

anrolmar:
# Describe the PR

This PR remove all the deprecated classes mentioned on https://github.com/bootstrap-vue-next/bootstrap-vue-next/issues/1779#issuecomment-1999436419

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

### DIFF
--- a/apps/docs/src/data/components/modal.data.ts
+++ b/apps/docs/src/data/components/modal.data.ts
@@ -349,10 +349,10 @@ export default {
           description: 'CSS class (or classes) to apply to the title',
         },
         {
-          prop: 'titleSrOnly',
+          prop: 'titleVisibilityHiddenOnly',
           type: 'boolean',
           default: false,
-          description: "Wraps the title in an '.sr-only' wrapper",
+          description: "Wraps the title in an '.visually-hidden' wrapper",
         },
         {
           prop: 'titleTag',
--- a/apps/docs/src/docs/components/input-group.md
+++ b/apps/docs/src/docs/components/input-group.md
@@ -67,7 +67,7 @@ Easily extend form controls by adding text, buttons, or button groups on either 
 
 You can attach addons using either props, named slots and/or sub-components.
 
-**Note** Bootstrap 5 dropped `.input-group-append` and `.input-group-prepend`. You can now just add buttons and `.input-group-text` as direct children of the input groups. The sub-components are added for compatibility reasons.
+**Note** Bootstrap 5 dropped `.input-group-text` and `.input-group-text`. You can now just add buttons and `.input-group-text` as direct children of the input groups. The sub-components are added for compatibility reasons.
 
 ### Using `prepend` and `append` props
 
--- a/apps/docs/src/docs/components/input-group.md
+++ b/apps/docs/src/docs/components/input-group.md
@@ -334,23 +334,42 @@ the control and the hidden label.
 Multiple add-ons are supported and can be mixed with checkbox and radio input versions.
 
 <HighlightCard>
+  <div class="mb-2">
+    <BInputGroup prepend="Item">
+      <BInputGroupPrepend is-text>
+        <input type="checkbox" aria-label="Checkbox for following text input">
+      </BInputGroupPrepend>
+      <BInputGroupPrepend is-text><b>$</b></BInputGroupPrepend>
+      <BFormInput type="number" aria-label="Text input with checkbox" />
+    </BInputGroup>
+  </div>
+  <div>
+    <BInputGroup append="Item">
+      <BFormInput type="number" aria-label="Text input with checkbox" />
+      <BInputGroupAppend is-text>
+        <input type="checkbox" aria-label="Checkbox for following text input">
+      </BInputGroupAppend>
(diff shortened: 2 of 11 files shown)


apps/docs/src/data/components/modal.data.ts
```diff
@@ -349,10 +349,10 @@ export default {
           description: 'CSS class (or classes) to apply to the title',
         },
         {
-          prop: 'titleSrOnly',
+          prop: 'titleVisibilityHiddenOnly',
```
VividLemon, line 352: This is a break change. It would be beneficial if you would reply to this with a collection of all the breaking changes so we can put them out for 0.17

VividLemon (changes requested):

apps/docs/src/data/components/modal.data.ts
anrolmar: I did a search for 'titleSrOnly' and only BModal component uses titleSrOnly at this moment. I did a PR with that change, and I prepared the rest of the components updating ComponentProps, too

apps/docs/src/docs/components/input-group.md
```diff
@@ -67,7 +67,7 @@ Easily extend form controls by adding text, buttons, or button groups on either
 
 You can attach addons using either props, named slots and/or sub-components.
 
-**Note** Bootstrap 5 dropped `.input-group-append` and `.input-group-prepend`. You can now just add buttons and `.input-group-text` as direct children of the input groups. The sub-components are added for compatibility reasons.
```
dwgray, line 70: I think this change is incorrect.  From the [BS5 migration guide](input-group-append): **Breaking** Dropped `.input-group-append` and `.input-group-prepend`. You can now just add buttons and `.input-group-text` as direct children of the input groups and it looks like the changes you made in the code were to change `Dropped .input-group-append` and `.input-group-prepend` to `.input-group-text`.

anrolmar: You're right. It's only remove the classes. It's about to add an `.input-group-text`with buttons.

anrolmar: I don't think so. We're using the right classes as you can see in the `BInputGroup` component:
```
<template>
  <component :is="tag" :id="id" class="input-group" :class="computedClasses" role="group">
    <slot name="prepend">
      <span v-if="hasPrepend" class="input-group-text">
        <!-- eslint-disable-next-line vue/no-v-html -->
        <span v-if="!!prependHtml" v-html="prependHtml" />
        <span v-else>{{ prepend }}</span>
      </span>
    </slot>
    <slot />
    <slot name="append">
      <span v-if="hasAppend" class="input-group-text">
        <!-- eslint-disable-next-line vue/no-v-html -->
        <span v-if="!!appendHtml" v-html="appendHtml" />
        <span v-else>{{ append }}</span>
      </span>
    </slot>
  </component>
</template>
```

The issue here is to fix the styling removing the `.input-group-append`and `.input-group-prepend` classes inside `_input-group.scss` to not miss behavior

VividLemon: @dwgray Your link sends me to the incorrect location. Can you repost it so I can review this and see who's correct, so I can make my decision?

dwgray: Sorry - that was less helpful without the link:  https://getbootstrap.com/docs/5.0/migration/#forms. I agree that @anrolmar is doing the right thing in the code, I just think the comment is wrong.  Note that Github is only highlighting the delete - its' actually the replace right after it that has on the face of it incorrect text. I think there may be an argument to delete the comment altogether (which may be what @anrolmar is saying).

apps/docs/src/docs/components/input-group.md
```diff
@@ -67,7 +67,7 @@ Easily extend form controls by adding text, buttons, or button groups on either
...
+**Note** Bootstrap 5 dropped `.input-group-text` and `.input-group-text`. You can now just add buttons and `.input-group-text` as direct children of the input groups. The sub-components are added for compatibility reasons.
```
dwgray, line 70: This is the text that I believe is wrong on the face based on [this](https://getbootstrap.com/docs/5.0/migration/#forms)

xvaara:
@anrolmar @dwgray @VividLemon  is this pr still in need of changes?

apps/docs/src/docs/components/input-group.md
```diff
@@ -67,8 +67,6 @@ Easily extend form controls by adding text, buttons, or button groups on either
 
 You can attach addons using either props, named slots and/or sub-components.
 
-**Note** Bootstrap 5 dropped `.input-group-append` and `.input-group-prepend`. You can now just add buttons and `.input-group-text` as direct children of the input groups. The sub-components are added for compatibility reasons.
```
VividLemon, line 70: @xvaara It seems like if this has been dropped, that we could remove the sub-components -- I noticed this the other day, that the two components were exactly the same

VividLemon: I removed these two components in my feature branch

xvaara:
I'll create a new pr based on this.

PR #1830, opened 2024-03-27
feat(BTable): add noSortReset option

dnewkerk:
# Describe the PR

Implements the noSortReset option for BTable. When enabled, sorting by a field only toggles between ascending and descending, and the sortBy value is not reset.

## PR checklist

- [x] Feature - `feat(...)`
- [ ] Other (please describe)

### DIFF
--- a/apps/docs/src/data/components/table.data.ts
+++ b/apps/docs/src/data/components/table.data.ts
@@ -42,6 +42,12 @@ export default {
           description: 'Do not select row when clicked',
         },
         {
+          prop: 'noSortReset',
+          type: 'boolean',
+          default: false,
+          description: 'Do not reset sort',
+        },
+        {
           prop: 'sortBy',
           type: 'string',
           default: undefined,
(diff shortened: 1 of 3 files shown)


VividLemon:
This is actually something that I had added in my current branch. Which contains a few other changes to Table. So, I'm going to prioritize that over this. Thanks though

VividLemon:
Actually, I'm going to keep this around. I may have misinterpreted what this prop actually does. When 0.17 is added, I may need you to check this PR again

VividLemon:
Actually, the functionality from bootstrap-vue is close to what I have done already. However, it's not perfect. It's going to be a deprecated prop. 

In bootstrap-vue, the no-sort-reset would only modify the sort state when you click on a column that is not sortable. Meaning that if the entire table is sortable, one could never "unsort" the items.

Instead, the functionality has changed. prop `mustSort` is going to do this, where if it's true, you could never (from the tables pov) unsort a column. If it's false, you can click to 'asc' -> 'desc' -> null (unsorted). In addition, one could make it an array (since it's multisortable) where you could say some fields are mustsort, while others aren't

PR #1831, opened 2024-03-27
fix(BFormCheckbox): Fixes #1822 - Prevent empty string being cast to true

### LINKED ISSUE #1822: Empty string as entry in options array of BFormCheckboxGroup returned as true in v-model
### Describe the bug

Hi, 

I am using the **BFormCheckboxGroup** and provide an `options` array with strings. In my use case one of the options can also be the empty string (`""`), but when the checkbox of the empty string is selected `v-model` returns `[true]` as the selected options instead of `[""]` (see attached stackblitz).

### Reproduction

https://stackblitz.com/edit/github-6phuvv?file=src%2FApp.vue

### Used Package Manager

npm

### Validations

dwgray:
# Describe the PR

Fix #1822 (Empty string as entry in options array of BFormCheckboxGroup returned as true in v-model) by inlining the `string | boolean` part of the `CheckboxValue` type in the `defineProps` of `BFormCheckboxGroup.vue`.

According the the [vue docs](https://github.com/vuejs/core/issues/4480#issuecomment-910753900), the way to disable the boolean casting rule in this case is to make sure `string` comes before `boolean` in the type union.  Changing the type definition in `CheckBoxTypes.ts` wasn't sufficient because the [compiler-sfc doesn't crawl external files](https://github.com/vuejs/core/issues/4480#issuecomment-910753900), so prepended the `string | boolean` part of the union inline in the place that compiler-sfc uses.  This is somewhat ugly, but I believe it is better than the alternatives that I considered:
- Redefine `CheckboxValue` in `BFormCheckboxGroup.vue`
- Move the definitions in `CheckboxTypes.ts` to `BFormCheckboxGroup.vue` and then make `index.ts` in types refer to those definitions

## Small replication

See #1822 

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)


packages/bootstrap-vue-next/src/types/CheckboxTypes.ts
```diff
@@ -1,8 +1,8 @@
 export type CheckboxValue =
   | readonly unknown[]
   | ReadonlySet<unknown>
-  | boolean
```
dwgray, line 4: This change isn't strictly necessary since the compiler-sfc never uses it, and AFAIK typescript itself doesn't care about the order.  But it seems like we should keep things consistent.

packages/bootstrap-vue-next/src/components/BFormCheckbox/BFormCheckbox.vue
```diff
@@ -58,7 +58,11 @@ const props = withDefaults(
     state?: boolean | null
     switch?: boolean
     uncheckedValue?: CheckboxValue
-    value?: CheckboxValue
+    // Since the compiler-sfc doesn't crawl external filed, the redundant string/boolean union is
+    //   necessary to tell it that we don't want it to follow Boolean casting rules
+    //  https://vuejs.org/guide/components/props.html#boolean-casting which would cast the empty
+    //  string to true
+    value?: string | boolean | CheckboxValue
```
VividLemon, line 65: This change shouldn't be necessary.

packages/bootstrap-vue-next/src/types/CheckboxTypes.ts
VividLemon, line 4: But this change should be.

Moving boolean after string is likely to fix the core issue. In relation to https://vuejs.org/guide/components/props.html#boolean-casting , the compiler should be able to crawl this file and apply string casting rules over boolean, with this change.

But I could be wrong, you should test to see if only this change works

VividLemon (changes requested):

packages/bootstrap-vue-next/src/types/CheckboxTypes.ts
dwgray: I tried just changing the type declaration in here first.  That had no effect.  I quoted a reference from Evan You in the description of this PR, but looking at [the vue docs](https://vuejs.org/api/sfc-script-setup) is even clearer:

> In dev mode, the compiler will try to infer corresponding runtime validation from the types. For example here `foo: String` is inferred from the `foo: string` type. If the type is a reference to an imported type, the inferred result will be `foo: null` (equal to `any` type) since the compiler does not have information of external files.

VividLemon:
@dwgray merge conflict detected :robot:

VividLemon (approved):
Get your PR fixed :rofl:

Issue #1730, opened 2024-01-26
Default component ids are not SSR friendly


Al-Rozhkov:
### Describe the bug

If component id is not explicitly specified, it is generated using Math.random(), which leads to hydration errors:

```
Hydration attribute mismatch on <input id=​"__BVID__921175___BV_form-check__" class=​"form-check-input" type=​"checkbox">​ 
  - rendered on server: id="__BVID__921175___BV_form-check__"
  - expected on client: id="__BVID__935395___BV_form-check__"
```

Solution: use the bootstrap-vue strategy. If the component id is not specified, leave it empty during SSR and generate it `onMounted`.

### Reproduction

https://stackblitz.com/edit/nuxt-starter-bavjvy?file=app.vue

### Used Package Manager

pnpm

### Validations

- [X] Have tested with the latest version. This is still alpha version and sometime things change rapidly.
- [X] Follow our [Code of Conduct](https://github.com/bootstrap-vue-next/bootstrap-vue-next/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guide](https://github.com/bootstrap-vue-next/bootstrap-vue-next/blob/main/CONTRIBUTING.md).
- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A, please open a GitHub Discussion instead.
- [X] The provided reproduction is a [minimal reproducible](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

VividLemon:
I'd prefer an alternative solution. Something more like https://github.com/vuetifyjs/vuetify/blob/master/packages/vuetify/src/util/getCurrentInstance.ts#L26

Al-Rozhkov:
It should be considered that `getCurrentInstance` is internal API: https://github.com/vuejs/docs/issues/1422#issuecomment-1032120675.

Which is why it's not in the official documentation.

PR #1836, opened 2024-04-03
feat(utils): inject getId to allow for custom id generation

reubns:
# Describe the PR
> [!WARNING]  
> Once Vue 3.5 is released, this PR will (likely) become obsolete in favour of the official method for generating IDs. If it is acceptable to wait for the 3.5 release, please close this PR.

Closes: #1730

This PR allows users to provide their own implementation for generating IDs. For example in Nuxt >= 3.10, to use SSR-friendly IDs.

```ts
// app.vue
provideGetId(() => useId())
```

> [!NOTE]  
> The Nuxt implementation above doesn't work for components with `inheritAttrs: false` due to the implementation of Nuxt's `useId`. For example, `BFormRadio`.

Credits: https://github.com/tailwindlabs/headlessui/pull/2959 https://github.com/radix-vue/radix-vue/pull/718

## Small replication

Issue is described in #1730

## PR checklist

- [x] Feature - `feat(...)`
- [ ] Other (please describe)


packages/bootstrap-vue-next/src/utils/getId.ts
```diff
@@ -1,2 +1,17 @@
-export default (suffix = ''): string =>
-  `__BVID__${Math.random().toString().slice(2, 8)}___BV_${suffix}__`
+import {inject, type InjectionKey, provide} from 'vue'
```
VividLemon, line 1: This file should now be located in the composables directory, since it's no longer a purely functional util

VividLemon: Likely it's best in /useId

packages/bootstrap-vue-next/src/utils/getId.ts
```diff
@@ -1,2 +1,17 @@
...
+}
+
+/**
+ * This function is allows users to provide a custom id generator
+ * as a workaround for the lack of stable SSR IDs in Vue 3.x
+ *
+ * This lets users of Nuxt swap in the Nuxt `useId` function
+ * which is stable across SSR and client.
+ */
+export const provideGetId = (fn: () => string) => provide(GET_ID, fn)
```
VividLemon, line 17: It may be better to register this as a global plugin. Similar to toastPlugin, etc. 

Then we install it automatically in the createBootstrap function

Then in Nuxt, when we consume createBootstrap(), we pass in the Nuxt useId function, so that way we automatically register this in Vue, and automatically register it in Nuxt for stable ids.

VividLemon: (I'm not sure how provide/inject with a runtime plugin that purpose is to generate stable ids will work. Some spot checking to make sure the ids become stable on the Nuxt playground will be useful)

VividLemon (changes requested):

packages/bootstrap-vue-next/src/utils/getId.ts
reubns: @VividLemon Thank you for the detailed review.

I've created `idPlugin` to allow this to be automatically registered in Vue and Nuxt.

Regarding the playground. Initially, changing to the plugin caused more hydration errors 🥲. Adding `.replace(':', '_'),` to the function provided to Nuxt `createBootstrap` means that the playground raises no errors. Although, this feels like a workaround https://github.com/nuxt/nuxt/pull/26315#issuecomment-2002563619 to accommodate the Nuxt implementation of `useId`.

I thought this a good point to check your thoughts on the current implementation.

xvaara:
Tested with nuxt kitchensink (all the playground components) and most of the hydration mismatches are gone :)

packages/bootstrap-vue-next/src/components/BFormGroup/BFormGroup.vue
```diff
@@ -224,7 +217,7 @@ export default defineComponent({
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     let $label: any = null
     const labelContent = normalizeSlot(SLOT_NAME_LABEL, {}, slots) || props.label
-    const labelId = labelContent ? getId('_BV_label_') : null
+    const labelId = labelContent ? useId(undefined, '_BV_label_').value : null
```
VividLemon, line 220: This stuff should go into the setup area...  I dont know if composables can go into the render func area. I would guess this id would rebuild on every render

packages/nuxt/src/runtime/createBootstrap.ts
```diff
@@ -1,11 +1,16 @@
...
+import {defineNuxtPlugin, useId} from '#imports'
 
 export default defineNuxtPlugin((nuxtApp) => {
   nuxtApp.vueApp.use(
     createBootstrap({
       components: false,
       directives: false,
+      plugins: {
+        id: {
+          getId: () => useId().replace(':', '_'),
```
VividLemon, line 11: Dont really see the point of the replace

VividLemon (changes requested):

packages/bootstrap-vue-next/src/components/BFormGroup/BFormGroup.vue
reubns: @VividLemon Would you prefer a separate PR for this? I can reproduce the id rebuild when rendering but believe the same behaviour existed before this PR.

VividLemon: Yes

VividLemon (approved):

PR #1837, opened 2024-04-03
fix(BPopover and Btooltip): Fixes bootstrap-vue-next#1232 - do not create a new app fro each tooltip or popover

### LINKED ISSUE #1232: One vue App for each tooltip
I just noticed that vue App is created for each tooltip/popover on the page. Is it right?

james-muriithi:
# Describe the PR
The PR fixes #1232 (One App for each tooltip), by eliminating the need to create a new app for each tooltip or popover. I used the `render` function from Vue to achieve this. One issue I experienced with this method was it did not update when props changed. 

To get around this I had to remove the old rendered component, then re-render it on update

## Small replication
check #1232 

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

PS: this is my first open-source contribution, please be easy on me :))

### DIFF
--- a/packages/bootstrap-vue-next/src/directives/BPopover.ts
+++ b/packages/bootstrap-vue-next/src/directives/BPopover.ts
@@ -31,21 +31,12 @@ export default {
 
     if (!text.content && !text.title) return
 
-    if (!el.$__state) {
-      // Same binding as above
-      // This happens when mounting occurs, but binding does not happen ie (if (!text.content && !text.title) return)
-      // So mounting occurs without a title or content set
-      el.$__state = ref({
-        ...resolveDirectiveProps(binding, el),
-        ...text,
-      })
-      bind(el, binding)
-      return
-    }
-    el.$__state.value = {
+    unbind(el)
+    el.$__state = ref({
       ...resolveDirectiveProps(binding, el),
       ...text,
-    }
+    })
+    bind(el, binding)
   },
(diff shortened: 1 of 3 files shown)


VividLemon (commented):
Looking good so far

xvaara:
If I read this correctly we loose reactivity. This is no different from the old way of writing html with div.innerHTML = '...'. All the elements in tooltips are re-rendered on each update. 

If creating an app for tooltip is a problem, user should use components. 

I'm against this change. How we do thing now is the same way https://github.com/Akryum/floating-vue/blob/main/packages/floating-vue/src/directives/v-tooltip.ts does directives that stay reactive.

xvaara:
This breaks so little and the things are weird anyway in directives so it might be ok for me.

I pushed a cleanup of unneeded $__state and $__app and imports. We just pass the props to the bind function.

xvaara (approved):

james-muriithi:
> This breaks so little and the things are weird anyway in directives so it might be ok for me.
> 
> I pushed a cleanup of unneeded $__state and $__app and imports. We just pass the props to the bind function.

Thank you for this

xvaara:
@james-muriithi you probably should be using tooltip as a component.

VividLemon:
Regardless, i agree with the overall change of not polluting the workspace.

PR #1825, opened 2024-03-22
feat(BCheckbox)!: Implement reverse and without label

dwgray:
# Describe the PR

Implement the [reverse](https://getbootstrap.com/docs/5.3/forms/checks-radios/#reverse) and [without-label](https://getbootstrap.com/docs/5.3/forms/checks-radios/#without-labels) features of checkboxes and radio buttons.

Reverse is implemented as a property on BFormCheckbox and BFormCheckboxGroup
Without Labels is implemented by not including the form-checkbox class when there is no default slot

In addition:
- Implemented and tested the same properties in BFormRadio and BFormRadioGroup
- Updated the documentation for BFromCheckbox and BFGormCheckboxGroup
- Made scope option in the ComponentReference data (because it is)
- Removed input and change events from BFormCheckboxGroup - these events were removed from BFromCheckbox in #1696  so it doesn't seem like it makes sense to keep them in the group.

I believe that this PR brings us to parity with BSV legacy and BS5. There is some more clean-up to do with BFormRadio, I didn't intend to touch that at all, but since there was a shared component I did the minimum to get the core feature I was working on to work with Radio controls. 

## Small replication

A small replication or video walkthrough can help demonstrate the changes made. This is optional, but can help observe the intended changes. A mentioned issue that contains a replication also works.

## PR checklist

**What kind of change does this PR introduce?** (check at least one)

- [x] Bugfix :bug: - `fix(...)`
- [x] Feature - `feat(...)`
- [ ] ARIA accessibility - `fix(...)`
- [x] Documentation update - `docs(...)`
- [ ] Other (please describe)

**The PR fulfills these requirements:**

- [x] Pull request title and all commits follow the [**Conventional Commits**](https://www.conventionalcommits.org/) convention or has an [**override**](https://github.com/googleapis/release-please#how-can-i-fix-release-notes) in this pull request body **This is very important, as the `CHANGELOG` is generated from these messages, and determines the next version type. Pull requests that do not follow conventional commits or do not have an override will be denied**

BREAKINGCHANGE: BFormCheckboxGroup and BFormRadioGroup no longer emit `input` or `change` events - please use update:modelValue instead

PR #1840, opened 2024-04-10
feat(BTable)!: add multisort capabilities

dwgray:
# Describe the PR

This is an update to #1835 with some progress.

- I believe I have multisort working and reasonably documented (and added in the 'complete example' that I've been using for testing.
- I've fixed the build errors, although there is an issue with `BLink` that I think might be a more fundamental problem.  But I don't use BLinks in the router scenario, so don't have enough of a sense of what's going on to understand the intent.
- I have _not_ fixed all of the tests.  I believe there is a fundamental issue with how `defineModel` works that is breaking the tests and causing other issues.

----

On the `defineModel` issue, it looks like the model isn't getting defined as expected.  I believe this may be the same issue that I saw in #1822 - According to the [vue docs](https://vuejs.org/api/sfc-script-setup#type-only-props-emit-declarations):

> In dev mode, the compiler will try to infer corresponding runtime validation from the types. For example here foo: String is inferred from the foo: string type. If the type is a reference to an imported type, the inferred result will be foo: null (equal to any type) since the compiler does not have information of external files.

In dev mode I'm seeing errors like this: `[Vue warn]: Invalid prop: type check failed for prop "modelValue". Expected Boolean, got Array` for things like BFormCheckbox
And if I move the type declaration of `BFormCheckboxProps` into `BFormCheckboxVue` I get an error on the when using `defineModel` - `Duplicate key 'modelValue'. May cause name collision in script or template tag.

I've actually tried reverting back to `useVModel ` and the tests still fail.  Also trie reverting the versions of dependencies in `package.json`  I must be misunderstanding something about how the system builds/works because I would expect the combination of those two actions to get back to passing tests.

----

On _multisort_, there were a couple of bugs in the existing implementation, and I'm not 100% certain that I kept the spirit of what @VividLemon intended.  But I documented what I implemented, so hopefully, it's close enough and should be more obvious what the intent is if you want to move in a different direction.

The one thing I didn't do that I think might be worthwhile is to allow `sortBy` to be a single `BTableSortBy` value when `multisort == false `.  I think that would be somewhat more ergonomic for a developer to consume. 

## Small replication

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)


packages/bootstrap-vue-next/src/components/BTable/table-simple.spec.ts
```diff
@@ -46,7 +46,7 @@ describe('BTableSimple', () => {
     })
     expect(wrapper.get('div').attributes('style')).toBe('max-height: 300px;')
     await wrapper.setProps({stickyHeader: false})
-    expect(wrapper.get('div').attributes('style')).toBeUndefined()
```
dwgray: I added logging to prove to myself that `responsiveStyles` returns `undefined` in the `{stickyHeader: false}` case - Vue is adding a style attribute with an empty string as the value, where I would expect it to just not add the style attribute. So I'm not sure what's going on.  The user-visible behavior should be the same, though, so I'm going to move on.

```
const responsiveStyles = computed<StyleValue | undefined>(() => {
  const ret = stickyIsValid.value ? {maxHeight: computedSticky.value} : undefined
  console.log(`responsiveStyles: ${ret}, stickyIsValid: ${stickyIsValid.value}`)
  return ret
})
```

apps/playground/src/components/Comps/TBLink.vue
```diff
@@ -30,3 +20,25 @@
       -->
   </BContainer>
 </template>
+
+<script setup lang="ts">
+import type {RouteLocationRaw} from 'vue-router'
+
```
dwgray, line 26: Note the comment in the code here - I don't understand the intent of BLink usage well enough to unravel this.

apps/playground/src/components/Comps/TPopover.vue
```diff
@@ -150,14 +150,7 @@
     </BRow>
     <BRow class="my-5 position-relative">
       <BCol>
-        <BPopover
-          v-model="value"
-          :placement="popoverPlacemet"
```
dwgray, line 155: popoverPlacement wasn't defined, so I removed it.

VividLemon: Yes, some of the build issues weren't necessarily that there was a breaking change, but rather that the playground had a misconfigured tsconfig and fixing it showed errors that didn't previously throw

packages/bootstrap-vue-next/src/components/BTable/table-simple.spec.ts
VividLemon: For reference, `console.log('responsiveStyles:', ret, 'stickyIsValid:', stickyIsValid.value)` may display better

apps/playground/src/components/Comps/TBLink.vue
VividLemon: I'll need to review the usage of vue-router & blink. I've had similar problems from vue-router when I'm not sure which version to use. The most abstract version in this case is the most appropriate. But it should also just be whatever is acceptable for vue-router's RouterLink 'to' prop -- which is `RouteLocationRaw`

This is the code for vue-router
```ts
/**
 * User-level route location
 */
export declare type RouteLocationRaw = string | RouteLocationPathRaw | RouteLocationNamedRaw;
```
So, it is most likely that you just have the wrong interfaces here. They should likely be `satisfies RouteLocationNamedRaw` & `satisfies RouteLocationPathRaw` respectively

packages/bootstrap-vue-next/src/components/BTable/table-simple.spec.ts
dwgray: Thanks - I always appreciate this sort of tip.

packages/bootstrap-vue-next/src/components/BFormCheckbox/form-checkbox.spec.ts
dwgray, line 1: This is puzzling.  I would have expected the tests that check for `input` and `change` events to start failing with #1825, which is when I removed the events from the code under test.  But instead, all of the tests that are checking for `emit` are failing now.  My first thought was that this was related to the useVmodel/defineModel changes that I noted in the [commit description(https://github.com/bootstrap-vue-next/bootstrap-vue-next/pull/1840#issue-2236507641).  But there is something else going on specifically with the checkbox event tests.

I did _just_ the package updates from this PR in a [separate branch](https://github.com/dwgray/bootstrap-vue-next/tree/checkbox-tests) and the only tests that fail are the checkbox tests (and they fail in the same way).

I checked the changelogs for the three most likely suspects (vue-test-utils, vue, and vitest) and didn't see anything suspicious.

[Releases · vuejs/test-utils (github.com)](https://github.com/vuejs/test-utils/releases)
[core/CHANGELOG.md at main · vuejs/core (github.com)](https://github.com/vuejs/core/blob/main/CHANGELOG.md)
[Release v1.4.0 · vitest-dev/vitest (github.com)](https://github.com/vitest-dev/vitest/releases/tag/v1.4.0)

Then I tried rolling them back explicitly and am still seeing the errors.  I may be misunderstanding how to roll  back packages - what I did was changed the version to the exact version we were previously using (without the '^') did a `pnpm-install` and then checked the `pnpm-lock` file to validate that it was showing the rolled back versions.

In any case, I can fix these specific tests, by testing the model value after the click rather than the emit, but I'm worried that this is reflecting a deeper problem.

VividLemon:
Closing -- commits pulled into https://github.com/bootstrap-vue-next/bootstrap-vue-next/pull/1842

packages/bootstrap-vue-next/src/components/BFormCheckbox/form-checkbox.spec.ts
VividLemon: Some failing tests appear to be the result of https://github.com/search?q=repo%3Abootstrap-vue-next%2Fbootstrap-vue-next%20%3Ato%3D%22container&type=code being changed to whatever they are now.

At the moment, I'm unsure about some of the others. 

Although, it appears that some other issues may be something to do with 
image
Under the circumstance where the wrapper component (rendercomponentorskip) is NOT skipped, it will create a wrapper div. Which means the native input and change events wont bubble up. If it is skipped, they will bubble up. However, we somewhat want these native events to bubble. I assume someone somewhere could use the `Event`. 

Custom components like BFormRadioGroup don't declare these, because there isn't really a purpose.

VividLemon: Related: https://github.com/capricorn86/happy-dom/issues/1409

PR #1847, opened 2024-04-15
fix(BForm*): Revert use of defineModel to prevent model warnings (fixed #1846)

### LINKED ISSUE #1846: Vue core issue tracker `Vue warns of a type check failure`
### Describe the bug

When using BFormRadioGroup, Vue warns of a type check failure for the v-model. It seems to always expect a Boolean. Check the console logs in the StackBlitz reproduction.

### Reproduction

https://stackblitz.com/edit/vue-yqphhw?file=src%2FApp.vue

### Used Package Manager

yarn

### Validations

dwgray:
# Describe the PR

Using defineModel with a type defined outside of the .vue file is causing the sfc to infer the wrong type.  I was unable to get defineModel to work for the BFrom* classes, so this PR is reverting to useVModel.  See my comment [here](https://github.com/bootstrap-vue-next/bootstrap-vue-next/issues/1846#issuecomment-2057882089) for more details.

## Small replication

This repros in the playground examples for BFromRadio* and BFromCheckbox* as well as the smaller repro in #1846.

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

### DIFF
--- a/packages/bootstrap-vue-next/src/components/BFormCheckbox/BFormCheckbox.vue
+++ b/packages/bootstrap-vue-next/src/components/BFormCheckbox/BFormCheckbox.vue
@@ -26,10 +26,10 @@
 </template>
 
 <script setup lang="ts">
-import {useFocus} from '@vueuse/core'
+import {useFocus, useVModel} from '@vueuse/core'
 import {computed, inject, ref, toRef} from 'vue'
 import {getClasses, getInputClasses, getLabelClasses, useId} from '../../composables'
-import type {BFormCheckboxProps, CheckboxValue} from '../../types'
+import type {ButtonVariant, CheckboxValue, Size} from '../../types'
 import {checkboxGroupKey, isEmptySlot} from '../../utils'
 import RenderComponentOrSkip from '../RenderComponentOrSkip.vue'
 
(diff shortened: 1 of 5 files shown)


dwgray:
@VividLemon This reverts the defineModel changes that were causing the sfc to infer the wrong type for the models of these components and throwing runtime warnings.

I'm not sure if there is a functional purpose for declaring the component types in `ComponentProps.ts` so I didn't revert those changes.  But since they are they are not used in the component definitions, I'd lean towards deleting them if they're not used for anything else.

xvaara:
@dwgray This is a bug in vue, it's fixed on the newest version. I'll release a new version tonight, after I get some stuff in.

dwgray:
> @dwgray This is a bug in vue, it's fixed on the newest version. I'll release a new version tonight, after I get some stuff in.

@xvaara Thanks - it looks like they've fixed some bugs in the SFC, including at least one that involves defineModel co-ercing strings to bools (which is what @VividLemon noted).  It still looks like the SFC is _not_ pulling in external type definitions  [vue docs](https://vuejs.org/api/sfc-script-setup#type-only-props-emit-declarations), so I'm not 100% those bug fixes will cover the issue we're seeing with radio button/checkbox models.
> 
> > In dev mode, the compiler will try to infer corresponding runtime validation from the types. For example here foo: String is inferred from the foo: string type. If the type is a reference to an imported type, the inferred result will be foo: null (equal to any type) since the compiler does not have information of external files.
> 

It's easy enough to validate, though - if it's not fixed, you'll see a lot of model warnings in the console window for the radio button and checkbox playground examples.  I'm at OSS this week but I will try to get a few cycles to play with this in the evening.  Hopefully it will just work :-)

PR #1869, opened 2024-04-28
fix(BTable): Pass through original records to slots when primary key is defined

dwgray:
# Describe the PR

For use cases when row items are defined as functional objects, it's convenient for the original object to be passed through to the scoped slots. The object still has to not contain loops and correctly transform through the JSON.parse(JSON.stringify()) loop, but it is returned to the client code intact.

The requirements for this to work are:
1. The table must contain a primary-key property
2. The object keys must not contain "." - I don't believe this is an issue, because the use case for this kind of pass-through is well-behaved typescript objects, which should have property names containing "."

If this is acceptable, I'll get documentation including the limitations (and probably some stuff around it) in as a follow-up PR.

## Small replication

Included in the test cases in the PR.

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

### DIFF
--- a/packages/bootstrap-vue-next/src/components/BTable/BTable.vue
+++ b/packages/bootstrap-vue-next/src/components/BTable/BTable.vue
@@ -434,7 +434,11 @@ const computedItems = computed<T[]>(() => {
 
   let mappedItems = usesProvider.value ? internalItems.value : (props.items as T[])
   mappedItems = mappedItems.map((item) => {
-    if (typeof item === 'object' && item !== null) {
+    if (
+      typeof item === 'object' &&
+      item !== null &&
+      Object.keys(item).some((key) => key.includes('.'))
+    ) {
       // We use any here because the TS doesn't isn't certain that "item" is the same type as our newItem.
       // But we've determined that it's an object, so we can ignore it since they will always be the same "object"
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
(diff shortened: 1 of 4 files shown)


packages/bootstrap-vue-next/src/components/BTable/BTable.vue
```diff
@@ -434,7 +434,11 @@ const computedItems = computed<T[]>(() => {
 
   let mappedItems = usesProvider.value ? internalItems.value : (props.items as T[])
   mappedItems = mappedItems.map((item) => {
-    if (typeof item === 'object' && item !== null) {
+    if (
+      typeof item === 'object' &&
+      item !== null &&
+      Object.keys(item).some((key) => key.includes('.'))
```
dwgray, line 440: This takes care of passing through the original objects to BTableLite except in the case where the a key contains "." - I believe this is an acceptable compromise as the use case for maintaining the object is well-behaved typescript objects, which shouldn't contain "." in property names.  It also has the benefit of not transforming object arbitrarily if they're 
"normal" objects.

packages/bootstrap-vue-next/src/components/BTable/BTableLite.vue
```diff
@@ -273,6 +273,26 @@ watch(
   }
 )
 
+const itemMap = computed(() => {
```
dwgray, line 276: This depends on having a primary key defined.  That's fine for my use case (and I think in general for the kind of case that I think would be important for enabling the pass through) - I don't see a more general way to do this, but if I'm missing something obvious, I'd be happy to take another un at it.  I'll also document this capability and its limitations in a follow-up PR (and probably get some more of the surrounding table docs added).

PR #1876, opened 2024-05-01
docs: Add component reference link in the page table of contents sidebar

dwgray:
# Describe the PR

As I'm spending more time in the component docs, I find it annoying that there isn't a link to the component reference in the page contents sidebar.  This is one solution to that problem (although I'm not sure it's the right one).

## Small replication

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

### DIFF
--- a/apps/docs/src/components/ComponentSidebar.vue
+++ b/apps/docs/src/components/ComponentSidebar.vue
@@ -1,0 +1,20 @@
+<template>
+  <ContentsSidebar>
+    <template #default>
+      <slot />
+    </template>
+    <template #after>
+      <nav class="table-of-contents">
+        <ul>
+          <li>
+            <a href="#component-reference">Component Reference</a>
+          </li>
+        </ul>
+      </nav>
+    </template>
+  </ContentsSidebar>
+</template>
+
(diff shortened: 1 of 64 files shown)


apps/docs/src/components/ComponentSidebar.vue
```diff
@@ -0,0 +1,20 @@
+<template>
+  <ContentsSidebar>
+    <template #default>
+      <slot />
+    </template>
+    <template #after>
```
dwgray, line 6: This works and gives me the right visual indicators, but I'm afraid it might be an accessibility violation having the second nav stacked below the main one.  I'm open to other ways to do this - I don't think we can teleport this into the right place and I didn't see any way to configure the markdown-it nav to help.

The alternatives I considered:
- Just explicitly style a link at the bottom of this section for jumping to the component ref, possibly styled visible differently form the nav above it, which would give us the functionality and possibly be better accessibility-wise
- Back this all out and just manually put a ##Component Reference markdown section just about the component reference and remove the title form the component reference component

Any thoughts?

PR #1882, opened 2024-05-06
fixed BTableLite item _showDetails

### LINKED ISSUE #1881: BTableLite _showDetails never change
### Describe the bug

After #1879  ```_showDetails``` in items never change

### Reproduction

https://stackblitz.com/edit/github-m6wxp7?file=src%2Fcomponents%2FComp.vue,src%2FApp.vue

### Used Package Manager

pnpm

### Validations

makroxyz:
# Describe the PR

fix #1881

### DIFF
--- a/packages/bootstrap-vue-next/src/components/BTable/BTableLite.vue
+++ b/packages/bootstrap-vue-next/src/components/BTable/BTableLite.vue
@@ -361,6 +361,7 @@ const toggleRowDetails = (tr: T) => {
   if (isTableItem(tr)) {
     const prevValue = detailsMap.value.get(tr)
     detailsMap.value.set(tr, !prevValue)
+    tr._showDetails = !prevValue
   }
 }
 


packages/bootstrap-vue-next/src/components/BTable/BTableLite.vue
```diff
@@ -361,6 +361,7 @@ const toggleRowDetails = (tr: T) => {
   if (isTableItem(tr)) {
     const prevValue = detailsMap.value.get(tr)
     detailsMap.value.set(tr, !prevValue)
+    tr._showDetails = !prevValue
```
VividLemon, line 364: This line will subsequently cause this watcher https://github.com/bootstrap-vue-next/bootstrap-vue-next/pull/1882/files#diff-798f265faf0ec8e8d395983b2a3e9eb972f91083f5f91852178ead74b95e4c93R292 to trigger. Its not infinite, but seems like a potential issue. Not sure the best way to fix this as by definition, 'tr' could be readonly (`readonly(ref([]))` or `computed`)

VividLemon: Practically speaking, it probably makes no difference :thinking:

VividLemon (changes requested):

packages/bootstrap-vue-next/src/components/BTable/BTableLite.vue
makroxyz: So, what can we do? 
Without this PR all tables with details I have in project which use _showDetail in items are dead

Issue #1753, opened 2024-02-07
Implement aliasing system


yogeshgalav:
### Clear and concise description of the problem

When migrating from bootstrap-vue  to bootstrap-vue-next, one of the major problem which leads to major code change is
"BInput" component not present, this componen is present but with other name "BFormInput"
Goal: minimum code change while migrating from bootstrap-vue.
Benifit: less boiler plate for component name, less or no code change for migration

### Suggested solution

The Alias for BFormInput should be BInput
just like Bootstrap-vue
This can be achieved with component.d.ts

### Alternative

New component BInput which directly exports BFormInput

### Validations

- [X] Follow our [Code of Conduct](https://github.com/bootstrap-vue-next/bootstrap-vue-next/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guide](https://github.com/bootstrap-vue-next/bootstrap-vue-next/blob/main/CONTRIBUTING.md).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

VividLemon:
Related: 
https://github.com/bootstrap-vue-next/bootstrap-vue-next/issues/1607
https://github.com/bootstrap-vue-next/bootstrap-vue-next/issues/1569

VividLemon:
In my point of view, "internal aliasing" is just bloat. I have mixed feelings about adding official aliased names and instead would prefer a system where users could maintain their own sets of aliases.

xvaara:
IMHO, if it's just one search and replace on the codebase, I don't think we need to worry about this. Just document the change to a migration guide.

yogeshGroony:
@VividLemon Just wanted to save time and effort to change lot of codebase.
Please do it, from product point of view it would be great.
If you don't want to make internal alias, then you can keep short ones
Like everyone will prefer BInput over BFormInput
From product point please make the change,
there is no other alternate available for bootstrap-vue in market.

VividLemon:
I was thinking about the purpose of the internal alias the other day, and came to the realization that it's a bit silly not to include it by default... It doesn't necessarily need to be a default, but could be an opt in feature (one that would hopefully require a simple one line change perhaps). Or simply just be made the default. That being said, the inclusion of default aliases would also require the ability to add custom aliases. So, in conclusion, once I finish the alias system, I will include plans for default aliases.

PR #1889, opened 2024-05-14
feat: implmenet a use defaults system WIP fixes #1607 

### LINKED ISSUE #1607: Global options
A requirement of the library is to have a mechanism for creating "global options". Such as setting the default for props. 

How exactly this would work idk yet.

But for instance this could be applied to having the nuxt module convert BLinks to NuxtLink (instead of RouterLink) by default, and other Nuxt specific options

VividLemon:
# Describe the PR

## Small replication

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

### DIFF
--- a/apps/docs/src/docs/configurations/global-options.md
+++ b/apps/docs/src/docs/configurations/global-options.md
@@ -1,0 +1,28 @@
+# Global options
+
+Global configurations allow you to set default prop values for all Vue components in your application. This is achieved using the createBootstrap plugin. Here's an example:
+
+```ts
+createBootstrap({
+  components: {
+    global: {
+      type: 'grow', // Caution: This impacts BFormInput, which does not support this value!
+    },
+    BAccordion: {
+      flush: true,
+    },
+    BFormText: {
+      text: 'foobar!',
+    },
+    BSpinner: {
+      type: 'grow',
+    },
+  },
+})
+```
+
+In the above example, each key of the components object corresponds to a Vue component name, and the value is an object that defines the default prop values for that component.
+
+The special key global can be used to set a default value for a prop across all components. For example, if you set type: 'grow' under global, all components that have a type prop will default to 'grow', regardless of whether the component's prop definition accepts it.
+
+Please note that using global is a significant decision that can have wide-ranging effects on your application, so use it judiciously. For example, one use case is the Nuxt plugin, which sets `teleportsTo` to the Nuxt `#teleports` target, providing a designated area for teleporting components.
--- a/apps/docs/src/docs/configurations/global-options.md
+++ b/apps/docs/src/docs/configurations/global-options.md
@@ -26,3 +26,11 @@ In the above example, each key of the components object corresponds to a Vue com
 The special key global can be used to set a default value for a prop across all components. For example, if you set type: 'grow' under global, all components that have a type prop will default to 'grow', regardless of whether the component's prop definition accepts it.
 
 Please note that using global is a significant decision that can have wide-ranging effects on your application, so use it judiciously. For example, one use case is the Nuxt plugin, which sets `teleportsTo` to the Nuxt `#teleports` target, providing a designated area for teleporting components.
+
+## Limitations with `modelValues`
+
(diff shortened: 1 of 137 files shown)


packages/bootstrap-vue-next/src/types/BootstrapVueOptions.ts
```diff
@@ -63,5 +71,6 @@ export interface BootstrapVueOptions {
      * @default true
      */
     toast?: boolean
+    components?: Partial<ComponentProps>
```
VividLemon, line 74: Duplicate key `components`

VividLemon (commented):
Remember to revert main.ts

xvaara:
Just heading to sleep. 5min look seems nice. I'll have time next week to look at this and other stuff.

Should we add the handeling of modelValue to useDefaults? pass emit to it, revert back to vueuse style handeling of it... Just an idea, not really sure about it.

VividLemon:
> Should we add the handeling of modelValue to useDefaults? pass emit to it, revert back to vueuse style handeling of it... Just an idea, not really sure about it.

There are some examples of how its used. Its a bit strange code, all boilerplate. But its just a computed setter.

VividLemon:
@xvaara Something that I didn't think about is the fact that we don't know whether or not the modelvalue is bound. This causes a problem:

```ts
const props = useDefaults<typeof _props>(_props, 'BAccordion')
const _modelValue = defineModel<string | undefined>({
  default: undefined,
})
const modelValue = computed({
  get: () => props.modelValue,
  set: (v) => {
    _modelValue.value = v
  },
})
```
When we bind to props.modelValue when it's using the "passive" version of the modelValue, it's then bound to the wrong value. 

```html
<BAccordion /> 
```
This won't work, because internally its supposed to be using the ref. However, since we bounded to the props, it fails. 

I'm not sure if there's a way to detect this (I don't think there is). Meaning, I think we may just have to refuse global options for modelValues

apps/docs/src/docs/configurations/global-options.md
```diff
@@ -0,0 +1,28 @@
...
+    BSpinner: {
+      type: 'grow',
+    },
+  },
+})
+```
+
+In the above example, each key of the components object corresponds to a Vue component name, and the value is an object that defines the default prop values for that component.
+
+The special key global can be used to set a default value for a prop across all components. For example, if you set type: 'grow' under global, all components that have a type prop will default to 'grow', regardless of whether the component's prop definition accepts it.
```
VividLemon, line 26:
```suggestion
The special key `global` can be used to set a default value for a prop across all components. For example, if you set type: 'grow' under global, all components that have a type prop will default to 'grow', regardless of whether the component's prop definition accepts it.
```

packages/bootstrap-vue-next/src/components/BAccordion/BAccordion.vue
```diff
@@ -21,6 +21,8 @@ defineSlots<{
   default?: (props: Record<string, never>) => any
 }>()
 
+const props = useDefaults<typeof _props>(_props, 'BAccordion')
```
VividLemon, line 24:
```suggestion
const props = useDefaults(_props, 'BAccordion')
```

VividLemon:
feat: implement a use defaults system fixes #1607

feat: implement global alias system fixes #1753

refactor: add bvn:: prefix to symbol keys

refactor: fix typo in 'BCardImageProps' change to 'BCardImgProps'

feat(BFormInput)!: move props lazy, trim & number to v-model modifier. Modifiers should behave more like native Vue modifiers (number modifier with unparseable value returns value as is)

feat(BFormInput): type number performs the same as modifier

feat(BCarouselSlide): allow individual interval for slides

chore: update deps to fix some issues (nuxt useId, vue-tsc etc)

docs: global options, aliases

feat(nuxt): automatically globally set teleportTo to `#teleports`

feat(BreakpointProps)!: strongly type breakpoints. Instead of weak string | number => 1,2,3...'1','2','3'...

refactor(BRow): use script setup instead of options api

refactor(BCol): use script setup instead of options api

Issue #1897, opened 2024-05-20
BTabs error with single BTab


makroxyz:
### Describe the bug

inject(...) is not a function when BTabs has only one BTab

### Reproduction

https://stackblitz.com/edit/github-vh7tfb?file=src%2Fcomponents%2FComp.vue,src%2FApp.vue

### Used Package Manager

npm

### Validations

- [X] Have tested with the latest version. This is still alpha version and sometime things change rapidly.
- [X] Follow our [Code of Conduct](https://github.com/bootstrap-vue-next/bootstrap-vue-next/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guide](https://github.com/bootstrap-vue-next/bootstrap-vue-next/blob/main/CONTRIBUTING.md).
- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A, please open a GitHub Discussion instead.
- [X] The provided reproduction is a [minimal reproducible](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

xvaara:
Super weird. Might be some optimization that vue does when single children component.

PR #1899, opened 2024-05-20
single tab error workaround

makroxyz:
# Describe the PR

This PR is a workaround for #1897 

Please have a look because I cannot understand what really happens


VividLemon:
Should make a issue with Vue core

xvaara:
It happens because tab is a computed that runs the buttonId which is useId the first time in the provide context, in this fix it's  run in the setup context. I fixed this in another pr.

PR #1903, opened 2024-05-21
feat(BOffcanvas): allow specifying a width

someonewithpc:
# Describe the PR

Allow specifying the width of the sidebar with a prop. This could be achieved with passing the `style` attribute, but it would be simpler for migrations if it could use the same props as `bootstrap-vue` (I think the aliases should be built in, too)

## Small replication

```vue
<template>
  <BContainer>
    <BButton @click="click">Show OffCanvas</BButton>
    <BOffcanvas v-model="show" width="20em" />
  </BContainer>
</template>

<script setup lang="ts">
import {BContainer, BOffcanvas, BButton} from './components'
import {ref} from 'vue'

const show = ref(false)

const click = () => {
  show.value = !show.value
}
</script>
```

## PR checklist

- [x] Feature - `feat(...)`
- [ ] Other (please describe)

### DIFF
--- a/packages/bootstrap-vue-next/src/types/ComponentProps.ts
+++ b/packages/bootstrap-vue-next/src/types/ComponentProps.ts
@@ -494,6 +494,7 @@ export interface BOffcanvasProps extends TeleporterProps {
   title?: string
   // responsive?: Breakpoint
   // TODO responsive doesn't work
+  width: string
 }
 
 export interface BOverlayProps extends RadiusElementExtendables {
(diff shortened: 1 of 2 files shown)


packages/bootstrap-vue-next/src/types/ComponentProps.ts
```diff
@@ -494,6 +494,7 @@ export interface BOffcanvasProps extends TeleporterProps {
   title?: string
   // responsive?: Breakpoint
   // TODO responsive doesn't work
+  width: string
```
VividLemon, line 497:
```suggestion
  width?: string
```

someonewithpc: Oops, forgot to push the version I had that fixed that, thanks for handling it :)

PR #1925, opened 2024-05-28
feat(BDropdown): add prop for max height and width for the dropdown menu

someonewithpc:
# Describe the PR

Add `max-height` and `max-width` props to the `BDropdown` component, to allow overriding the size calculated by `@floating-ui/vue`

## Small replication

An example was added to the docs

## PR checklist

- [x] Feature - `feat(...)`
- [ ] Other (please describe)

### DIFF
--- a/apps/docs/src/data/components/dropdown.data.ts
+++ b/apps/docs/src/data/components/dropdown.data.ts
@@ -105,6 +105,18 @@ export default {
           type: 'number | string | {mainAxis?: number; crossAxis?: number; alignmentAxis?: number | null',
           default: 0,
         },
+        {
+          prop: 'maxHeight',
+          type: 'number | string',
+          default: undefined,
+          description: 'The max-height to use for the dropdown menu',
+        },
+        {
+          prop: 'maxWidth',
+          type: 'number | string',
+          default: undefined,
+          description: 'The max-width to use for the dropdown menu',
+        },
       ],
       emits: [
         {
--- a/apps/docs/src/docs/components/dropdown.md
+++ b/apps/docs/src/docs/components/dropdown.md
@@ -548,6 +548,60 @@ const show4 = ref(false)
 
 **Note:** _changing the size of the button(s) does not affect the size of the menu items!_
 
+### Menu sizing
+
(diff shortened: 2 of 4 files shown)


someonewithpc:
I'll note that part of my motivation for this was that I couldn't get the `offset` to work, but it's probably something in my setup (old app, getting updated). However, from reading the code, it seems like it uses `@floating-ui/vue` which hard-codes using the CSS `top` and `left`, but it feels like for `placement: end`, it should use `right`, and so on, no?

xvaara:
What is the use case for this feature? you want to limit the size more than the available space and make dropdown scrollable?

Why couldn't you just use css to make this happen? 
Maybe we should but the floating-ui height to css var so it could be used in a class to do the calculation on css side.

xvaara:
> I'll note that part of my motivation for this was that I couldn't get the `offset` to work, but it's probably something in my setup (old app, getting updated). 

Can you provide a minimal reproduction of that?

> However, from reading the code, it seems like it uses `@floating-ui/vue` which hard-codes using the CSS `top` and `left`, but it feels like for `placement: end`, it should use `right`, and so on, no?

no

someonewithpc:
> What is the use case for this feature? you want to limit the size more than the available space and make dropdown scrollable?

Yes, that's it

> Why couldn't you just use css to make this happen? Maybe we should but the floating-ui height to css var so it could be used in a class to do the calculation on css side.

I can't (easily) do it in CSS because the `max-height` gets applied as an in-line style, so it overrides my class (would have to use `!important`)

someonewithpc:
> Can you provide a minimal reproduction of that?

I'll try, but I'm currently trying to reproduce an error I'm getting with tooltips which isn't reproducing outside the app I'm working on migrating

xvaara:
> > What is the use case for this feature? you want to limit the size more than the available space and make dropdown scrollable?
> 
> Yes, that's it
> 
> > Why couldn't you just use css to make this happen? Maybe we should but the floating-ui height to css var so it could be used in a class to do the calculation on css side.
> 
> I can't (easily) do it in CSS because the `max-height` gets applied as an in-line style, so it overrides my class (would have to use `!important`)

so you're saying !important is hard to add to css? 

I'm gonna close this since it's easily solvable with css and your code isn't working to begin with. I'll create a issue for myself to convert the height to a variable.

PR #1935, opened 2024-06-01
docs(BFormRadio): Get docs to parity

dwgray:
# Describe the PR

- Update BFormRadio and BFormRadioGroup docs to get them to parity with BSV and BS5.
- Changed one type interface to a type for consistency between BFormRadio and BFormCheckbox
- Small updates to BFormChecbox docs to be more consistent

## Small replication

N/A

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

### DIFF
--- a/apps/docs/src/docs/components/form-checkbox.md
+++ b/apps/docs/src/docs/components/form-checkbox.md
@@ -269,6 +269,68 @@ const checkEx2Options = [
   </template>
 </HighlightCard>
 
+## Checkbox values and `v-model`
+
+By default, `BFormCheckbox` value will be true when checked and false when unchecked. You can customize the checked and unchecked values by specifying the `value` and `unchecked-value` properties, respectively.
+
+When you have multiple checkboxes that bind to a single data state variable, you must provide an array reference (`[ ]`) to your v-model.
+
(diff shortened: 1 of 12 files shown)


apps/docs/src/docs/components/form-checkbox.md
```diff
@@ -269,6 +269,68 @@ const checkEx2Options = [
   </template>
 </HighlightCard>
 
+## Checkbox values and `v-model`
```
dwgray, line 272: This is just a move to make the order consistent with BFormRadio - it puts the meatier discussion about how model works about descriptions of variations on style, which seems appropriate.

packages/bootstrap-vue-next/src/types/RadioTypes.ts
```diff
@@ -6,7 +6,7 @@ export type RadioValue =
   | number
   | null
 
-export interface RadioOption {
```
dwgray, line 9: I changed this to be consistent with CheckboxOption.  According to the [typescript docs](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#differences-between-type-aliases-and-interfaces):

> Type aliases and interfaces are very similar, and in many cases you can choose between them freely. Almost all features of an interface are available in type, the key distinction is that a type cannot be re-opened to add new properties vs an interface which is always extendable.

I don't have a real preference here, but think they should be consistent. Happy to change them both to interfaces if anyone has a preference.

VividLemon (dismissed):

Issue #1584, opened 2023-11-12
docs: show extended values through the component reference


VividLemon:
### Clear and concise description of the problem

In situations where components extend the values of other components, there needs to be an extension to the component reference. 

I believe something like (BAvatar example)

| BAvatar props                            |                                         |                                       |
|------------------------------------------|-----------------------------------------|---------------------------------------|
| badge                                    | default: false                          | badge desc                            |
| Omit<BLinkProps, 'event' \| 'routerTag'> |                                         |                                       |
| variant                                  | 'secondary' (Not a BLink default)       | (desc taken automatically from BLink) |
| active                                   | undefined -- (default taken from BLink) | (desc taken automatically from BLink) |

The purpose is to make this automatic. Without requiring you to re-declare props all over the place.

VividLemon:
related: https://github.com/bootstrap-vue-next/bootstrap-vue-next/pull/1580#discussion_r1389474061

Issue #1957, opened 2024-06-15
Reduce redundancy in component data


dwgray:
### Clear and concise description of the problem

There are quite a few properties that are used by many components. I'm not talking about things that are necessarily related in the code, but just basic properties that are the same throughout the product. Things like `aria-role`, `name` and `variant`. It's cumbersome to maintain these for all of the different components that use them.

### Suggested solution

The way that BSV solves this problem is by having a [common-props.json](https://github.com/bootstrap-vue/bootstrap-vue/blob/576e67b3af434037a5ee17533a232465527d5edd/docs/common-props.json) file that specifies props that can be used by any of the documentation. And then it uses generic names like 'the component' in the description rather than the name of the component.  This doesn't seem like a loss of information - I didn't even realize BSV was doing this until I dug into how they generate the docs, and I've been using BSV for years.

In our case, I propose adding a common-properties.json file that has an array of properties (we'd have to factor out ComponentReference props), then make any prop in the concrete component's ComonentReference that just has undefined `type`, `description` or `default` fields pull from the equal named `prop` in the common-properties.json file. Fairly straightforward and let's us override properties that have the same name but different types or descriptions while reducing the copy/paste and maintenance overhead of the common cases.

There is some additional magic involved in BSV where it seems to be using the underlying code to know which properties to populate. At some future point I might dive into trying some variation of that, possibly based on get the typescript compiler to spit out the type info for the component props, but I think the proposal outlined above is still a significant win.

### Additional context

@VividLemon I'm happy to tackle this - I'd like to spend part of my time reducing the overhead of filling out the documentation as I move forward, cleaning up things, and filling out the parity report. I just want to get an initial reaction as to whether you think something like this makes sense before diving in.

### Validations

- [X] Follow our [Code of Conduct](https://github.com/bootstrap-vue-next/bootstrap-vue-next/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guide](https://github.com/bootstrap-vue-next/bootstrap-vue-next/blob/main/CONTRIBUTING.md).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

VividLemon:
If you're talking about for the documentation, the files are just TS files. You can import values, spread arrays, and do anything else you want to share common prop definitions

dwgray:
I've got PR #1964 up with an initial implementation of what I meant by the straightforward approach.

[...]

PR #1964, opened 2024-06-18
docs: Implement the concept of 'common-props' to reduce redundancy in component references

dwgray:
# Describe the PR

This addresses #1957. I pulled the idea from BSV, where they keep a flat file of prop definitions that are re-used in multiple. components.

- @VividLemon reworked the props field to be in object format and type-checked.
- He also added the ability to incorporate 'extensions' in a way that pulls in the property from a different component and puts those definitions in a sub-table (this is demonstrated in `table.md`
- I used `BFormSelect*`, `BFormCheckbox*` and `BFormRadiobutton*` as a trial, it reduced the component reference footprint significantly (although the `BForm*` compnenets are probably the best case for this).

## Small replication

See the `BFormSelect*`, `BFormCheckbox*`, `BFormRadiobutton*` and `BTable` documentation

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)


apps/docs/src/docs/components/form-radio.md
```diff
@@ -734,9 +734,10 @@ Supported `aria-invalid` values are:
 
 `aria-invalid` is automatically set to `true` if the `state` prop is `false`.
 
-<ComponentReference :data="data" />
+<ComponentReference :data="data" :common-props="commonProps" />
```
dwgray, line 737: I'm think I may be able to inject this object from a root component so that I don't have to add this property and the import to every *.md file, but I'd like to get feedback on the approach before diving into that.

apps/docs/src/data/components/formCheckbox.data.ts
```diff
@@ -206,121 +133,58 @@ export default {
...
           default: false,
           description: 'When set, adds the Bootstrap class `was-validated` to the group wrapper',
         },
-        {
-          prop: 'valueField',
-          type: 'string',
-          default: 'value',
-          description: 'Field name in the `options` array that should be used for the value',
-        },
+        {prop: 'valueField'},
```
VividLemon, line 187: Rather than this, I'd actually like a schema more like the following

```ts
const commonProps = [
  // Props definitions here
]

// In different file
import {commonProps} from '../'
export default {
load: (): ComponentReference[] => [
  // Previous data
  ...pick(commonProps, ['disabled']) // A pick function exists in the main package I think. It's essentially a TS equivalent to Pick<>
]
```

Then there's also extensions. 

```ts
// For BOverlayProps extends RadiusElementExtendables example
const RadiusElementExtendables = [] 

export default {
  load: () => [
    RadiusElementExtendables
  ]
}
```
Then RadiusElementExtendables appears as a table subsection. 

```md
| Main props          |
|--------------|
| bar          |
| buzz         |
| RadiusElementExtendables  |
|--------------|
| Section Prop |
```

VividLemon: Load looks like `load: (): (ComponentReference | ComponentReference[])[] => [` 
and pick exists in `object.ts`

VividLemon: Alternatively the commonProps can exist as an extension like so, and would not need to be spreaded
```ts
load: () => [
  pick(commonProps, ['disabled'])
]
```

VividLemon: (or if BTable can't do this, multiple concurrent tables would be the same)
```md
| RadiusElementExtendables |
|---------------------|
| bar                 |
| buzz                |

| RadiusElementExtendables |
|---------------------|
| Section Prop        |
```

dwgray: Thanks, @VividLemon - I think I've got the gist of what you're proposing. I'll turn around something based on that next time I've got some time to play with this (probably this week).

VividLemon: A large undertaking for sure, especially [`https://github.com/bootstrap-vue-next/bootstrap-vue-next/issues/1957#issuecomment-2177246378](https://github.com/bootstrap-vue-next/bootstrap-vue-next/issues/1957#issuecomment-2177246378)

VividLemon: Alternatively a discord call could clear up any confusion if you had any

VividLemon:
@dwgray That should be all the changes I planned on making. You should just need to fill in the blanks

dwgray:
@VividLemon thanks for working through this. I agree that the object form reads better for properties and the type-checking actually caught one error in my original work. I updated the *.data.md files that I touched, fixed some build/linting issues in a couple of other files and updated common props.

I think this is ready to go and will unblock me to spend part of my time verifying and filling out the component data for other components as I work through the parity validation. I'm finding this to be slow, but I might as well do all of the validation and get the docs/component data up-to-date all at the same time.

Out of curiosity, did you find a clean way to update the props in the *.data.ts files from array to object format? I got github copilot to do that on a per file basis, but didn't find anything that would let me do a global search and replace equivalent...

VividLemon:
fixes  #1584

Issue #1703, opened 2023-12-31
Docs build "document not defined"


VividLemon:
### Describe the bug

In the docs there is a series of errors declaring that document is not defined. This is likely an error in some code that is not SSR friendly. I'm not sure when this started

### Reproduction

f

### Used Package Manager

pnpm

### Validations

- [X] Have tested with the latest version. This is still alpha version and sometime things change rapidly.
- [X] Follow our [Code of Conduct](https://github.com/bootstrap-vue-next/bootstrap-vue-next/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guide](https://github.com/bootstrap-vue-next/bootstrap-vue-next/blob/main/CONTRIBUTING.md).
- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A, please open a GitHub Discussion instead.
- [X] The provided reproduction is a [minimal reproducible](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #1975, opened 2024-06-26
docs: get dark mode working again

dwgray:
# Describe the PR

Code blocks are currently not respecting dark mode (the text in the code block is using the light-mode colors).

The current solution is to go back to manually setting/removing `class='dark'` on the `HTML` tag, but to set it up in the `onMounted` hook, which only runs on the client.

----

I'm keeping the notes below in case I (or someone else) wants to take a stab at a more elegant solution in the future.

@VividLemon - This fixes dark mode, but regresses the documentation warning fix, I've done a bunch of exploration and haven't found a better fix. If you've got a chance to take a look and point me in the right direction, that would be great; otherwise, I'll sleep on it and take another look later.

I broke this with #1703 - specifically the code that pulled out setting/clearing class="dark" on the document element.
But I think this code was only coincidentally working and must have covered a change in Vitepress that broke us.

I can't figure out the mechanism that does this, but it appears that either [shiki](https://github.com/shikijs/shiki)  or vitepress' use of shiki required that the <HTML> element have class='dark' to render in dark mode. Setting class='dark' further down in the document doesn't work, but manually setting `class='dark'` on the document node makes everything work.
Vitepress' default theme's use of `useColorScheme` sets `class='dark', and it works. We use `data-bs-theme` instead of adding a class and it fails. 

I went down a path of exploring setting `color-scheme:dark`, since vitepress has some css that does that

I also tried calling `useColorMode` with `attribute='class'` and `selector = 'html'` to change the class attribute rather than the `data-bs-theme` attribute on `body`, but then most of the BSVN components just ignore dark mode. I'm still seeing color-`scheme:dark` show up in the debug tools, so I guess there is some other mechanism that I'm missing?

I don't know if there is a way to get the call in the onChanged handler to execute only on the client? That would fix both the warning and dark mode - it still feels like a kludge, but I'd take that at this point.

Otherwise, the only path I can think of is to continue to figure out how BSVN and VitePress handle dark/light mode and figure out how to get them to play well together.

## Small replication

Set dark mode in docs and see code block.

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

### DIFF
--- a/apps/docs/.vitepress/theme/Layout.vue
+++ b/apps/docs/.vitepress/theme/Layout.vue
@@ -265,12 +265,10 @@ const map = {
 
 const options = Object.keys(map) as (keyof typeof map)[]
 
-const currentIcon = computed(() => map[colorMode?.value ?? 'light'])
+const currentIcon = computed(() => map[colorMode.value])
 
 const set = (newValue: keyof typeof map) => {
-  if (colorMode) {
-    colorMode.value = newValue
-  }
+  colorMode.value = newValue
 }
 
 watch(isLargeScreen, (newValue) => {
--- a/apps/docs/.vitepress/theme/Layout.vue
+++ b/apps/docs/.vitepress/theme/Layout.vue
@@ -239,20 +239,22 @@ const headerExternalLinks = [
   },
 ]
 
-let colorMode: UseColorModeReturn<BasicColorMode> | null = null
+const colorMode = useColorMode({
+  persist: true,
+})
 
 onMounted(() => {
-  colorMode = useColorMode({
-    persist: true,
-    onChanged(mode, defaultHandler) {
-      defaultHandler(mode)
-      if (mode === 'dark') {
+  watch(
+    colorMode,
+    (newValue) => {
+      if (newValue === 'dark') {
         document.documentElement.classList.add('dark')
       } else {
         document.documentElement.classList.remove('dark')
       }
     },
-  })
+    {immediate: true}
+  )
 })
 
 const map = {
--- a/apps/docs/.vitepress/theme/Layout.vue
+++ b/apps/docs/.vitepress/theme/Layout.vue
@@ -241,6 +241,14 @@ const headerExternalLinks = [
 
 const colorMode = useColorMode({
   persist: true,
+  onChanged(mode, defaultHandler) {
+    defaultHandler(mode)
+    if (mode === 'dark') {
+      document.documentElement.classList.add('dark')
+    } else {
+      document.documentElement.classList.remove('dark')
+    }
+  },
 })
 
 const map = {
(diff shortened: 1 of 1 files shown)


VividLemon:
I wonder if an onMounted watcher with immediate true may work

dwgray:
> I wonder if an onMounted watcher with immediate true may work

Thanks! I pushed a change using onMounted, which works and doesn't cause the warnings. I think this is at least a viable short-term solution. I'd like to poke around a little more and see if I can understand what's going on so that we can get VitePress and BSVN's use of color scheme to work together more naturally, but I can leave that for another day.

apps/docs/.vitepress/theme/Layout.vue
```diff
@@ -239,8 +239,20 @@ const headerExternalLinks = [
...
+    onChanged(mode, defaultHandler) {
+      defaultHandler(mode)
+      if (mode === 'dark') {
+        document.documentElement.classList.add('dark')
+      } else {
+        document.documentElement.classList.remove('dark')
+      }
+    },
+  })
 })
```
VividLemon, line 256: I meant more like
```suggestion
useColorMode({
    persist: true,
    onChanged(mode, defaultHandler) {
      defaultHandler(mode)
      if (mode === 'dark') {
        document.documentElement.classList.add('dark')
      } else {
        document.documentElement.classList.remove('dark')
      }
    },
  })
})

onMounted(() => {
  watch(colorMode, (newValue) => {
     if (mode === 'dark') {
        document.documentElement.classList.add('dark')
      } else {
        document.documentElement.classList.remove('dark')
      }
  }, {immediate: true})
})
```

VividLemon: When built this way watchers are added like normal, but only started after the mount phase -- which would be client side. Not sure how vitepress deals with that tho.

Ig another option would be to make a renderless component, wrap it in client only, then make the result of color mode a prop and init the watcher in there. But this is not the most ideal solution, just a back pocket option

dwgray: I see. Yes, that's definitely cleaner than what I did and works. I just pushed a commit incorporating those changes, verified that they worked and don't re-introduce the build warnings.

VividLemon (changes requested):

VividLemon (approved):

apps/docs/.vitepress/theme/Layout.vue
```diff
@@ -251,10 +265,12 @@ const map = {
 
 const options = Object.keys(map) as (keyof typeof map)[]
 
-const currentIcon = computed(() => map[colorMode.value])
+const currentIcon = computed(() => map[colorMode?.value ?? 'light'])
```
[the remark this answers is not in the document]
dwgray: Thanks. Yep, pulled the null checks as colorMode is now always created in setup.

PR #1991, opened 2024-07-01
fix(BFormFloatingLabel): Clean up props

dwgray:
# Describe the PR

- The `text` prop on `BFormFloatingLabel` used the value of that property as the fallback content of the `default` slot. This doesn't work, that slot has to contain one of select few form controls. Putting plain text there just generates ugliness.
- Make slots required where appropriate (`BForm`, `BFormFloatingLabel` and `BFormRow`)

## Small replication

[stackblitz](https://stackblitz.com/edit/github-llmgfm?file=src%2Fcomponents%2FComp.vue)

```
<template>
  <Container>
    <BRow
      ><BCol>
        <BForm class="floating-label">
          <BInput
            id="floatingEmail"
            type="email"
            class="form-control"
            placeholder="Email address"
            label="Email address"
          />
          <BInput
            id="floatingPassword"
            type="password"
            placeholder="Password"
            label="Password"
          />
        </BForm> </BCol
    ></BRow>
    <BRow
      ><BCol>
        <BForm>
          <BFormFloatingLabel
            label="Email address"
            label-for="floatingEmail"
            class="my-2"
            text="Testing floating label text"
          >
          </BFormFloatingLabel>
        </BForm> </BCol
    ></BRow>
  </Container>
</template>
```
## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

### DIFF
--- a/packages/bootstrap-vue-next/src/components/BForm/BForm.vue
+++ b/packages/bootstrap-vue-next/src/components/BForm/BForm.vue
@@ -17,6 +17,7 @@ import {computed, ref} from 'vue'
 
 const _props = withDefaults(defineProps<BFormProps>(), {
   id: undefined,
+  floating: false,
   novalidate: false,
   validated: false,
 })
@@ -34,6 +35,7 @@ defineSlots<{
 }>()
 
 const computedClasses = computed(() => ({
+  'form-floating': props.floating,
   'was-validated': props.validated,
 }))
 
--- a/packages/bootstrap-vue-next/src/components/BForm/BForm.vue
+++ b/packages/bootstrap-vue-next/src/components/BForm/BForm.vue
@@ -16,7 +16,6 @@ import type {BFormProps} from '../../types'
 import {computed, ref} from 'vue'
 
 const _props = withDefaults(defineProps<BFormProps>(), {
-  floating: false,
   id: undefined,
   novalidate: false,
   validated: false,
@@ -35,7 +34,6 @@ defineSlots<{
 }>()
 
 const computedClasses = computed(() => ({
-  'form-floating': props.floating,
   'was-validated': props.validated,
 }))
 
--- a/packages/bootstrap-vue-next/src/components/BForm/BFormFloatingLabel.vue
+++ b/packages/bootstrap-vue-next/src/components/BForm/BFormFloatingLabel.vue
@@ -1,8 +1,6 @@
 <template>
   <div class="form-floating">
-    <slot>
-      {{ props.text }}
-    </slot>
+    <slot />
     <label :for="props.labelFor">
       <slot name="label">
         {{ props.label }}
@@ -24,7 +22,7 @@ const props = useDefaults(_props, 'BFormFloatingLabel')
 
 defineSlots<{
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  default?: (props: Record<string, never>) => any
+  default: (props: Record<string, never>) => any
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   label?: (props: Record<string, never>) => any
 }>()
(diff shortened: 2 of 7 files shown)


packages/bootstrap-vue-next/src/components/BForm/BForm.vue
```diff
@@ -16,7 +16,6 @@ import type {BFormProps} from '../../types'
 import {computed, ref} from 'vue'
 
 const _props = withDefaults(defineProps<BFormProps>(), {
-  floating: false,
```
dwgray, line 19: I have notes in the PR description and an example as to why I believe this property should be removed.

packages/bootstrap-vue-next/src/components/BForm/BFormFloatingLabel.vue
```diff
@@ -24,7 +22,7 @@ const props = useDefaults(_props, 'BFormFloatingLabel')
 
 defineSlots<{
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  default?: (props: Record<string, never>) => any
```
dwgray, line 27: I have notes in the PR description and an example as to why I believe this property should be removed.

VividLemon:
Not sure why you think you can't have form-floating on a form element
image

Idrk why one would want to use form over a div in this case though

dwgray:
I think those examples in BS5 are misleading. They only work if you have a single control/label combination in the form.  See [this stackblitz](https://stackblitz.com/edit/aqveji?file=index.html) for a counter-example in native BS5.

Us putting a prop called `floating` implies to me that all of the controls in the form will somehow be floating (maybe we could implement that, but it would be complicated and is not accomplished by merely putting `form-floating` on the form element. Someone who wanted to mimic the single control example could add the class on the `BForm` themselves and would hopefully know enough not to shoot themselves in the foot.

If you really want to keep that prop in, let me know.  In that case, I'd also like to add some stuff to the docs so people don't assume it does more than it does. Alternately, I could take it out and include a single form control example. Let me know which you prefer.

VividLemon:
Regardless on if the class is misleading, it is native bootstrap form code. There's no reason to remove it

packages/bootstrap-vue-next/src/components/BForm/BForm.vue
```diff
@@ -35,7 +34,6 @@ defineSlots<{
 }>()
 
 const computedClasses = computed(() => ({
-  'form-floating': props.floating,
```
VividLemon, line 38: This is probably my only issue with this PR. I suppose some components could benefit from making their slots not optional as shwon in BFormFloatingLabel. But this could be something else, or not added

VividLemon (changes requested):

packages/bootstrap-vue-next/src/components/BForm/form.spec.ts
```diff
@@ -28,15 +28,6 @@ describe('form', () => {
     expect(wrapper.attributes('novalidate')).toBeUndefined()
   })
```
VividLemon, line 30:
```suggestion

it('has class form-floating when prop floating', async () => {
    const wrapper = mount(BForm, {
      props: {floating: true},
    })
    expect(wrapper.classes()).toContain('form-floating')
    await wrapper.setProps({floating: false})
    expect(wrapper.classes()).not.toContain('form-floating')
  })

```

packages/bootstrap-vue-next/src/components/BForm/form.spec.ts
```diff
@@ -28,7 +28,7 @@ describe('form', () => {
...
 
-  it('has class form-floating when prop floating', async () => {
+it('has class form-floating when prop floating', async () => {
```
VividLemon, line 31:
```suggestion
  it('has class form-floating when prop floating', async () => {
```

VividLemon (approved):

dwgray:
> Regardless on if the class is misleading, it is native bootstrap form code. There's no reason to remove it

Okay, I will defer to you on this. Added that property back in. I have #2003 that was in progress when I saw this that fixes some other docs issues. I've added some documentation about the `floating` label that I hope will make its limited value clear to users.

PR #2007, opened 2024-07-09
fix(BOffcanvas): Fix regression in docs caused  by #1992

dwgray:
# Describe the PR

This fixes an issue caused by #1992 where the overlay in `BOffCanvas` was not behaving as expected (see repro).

Revert to using `BOverlay` rather than manually setting up an overlay in `BOffcanvas`

I didn't dig further to see how hard it would be to get this working with out using the `BOverlay` component, I'm not sure that would be useful (possibly slightly better performance?)

## Small replication

See the docs site against main:
image

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

### DIFF
--- a/apps/docs/.vitepress/theme/Layout.vue
+++ b/apps/docs/.vitepress/theme/Layout.vue
@@ -94,11 +94,10 @@
         <BOffcanvas
           id="sidebar-menu"
           v-model="sidebar"
-          :teleport-disabled="true"
-          :backdrop="false"
+          responsive="lg"
           title="Browse docs"
+          header-class="d-flex"
           class="h-100 border-0"
-          :body-scrolling="isLargeScreen"
         >
           <TableOfContentsNav />
         </BOffcanvas>
@@ -128,13 +127,11 @@
               <BOffcanvas
                 id="otp-menu"
                 v-model="onThisPage"
-                :teleport-disabled="true"
-                :backdrop="false"
+                responsive="lg"
                 placement="end"
                 title="On this page"
                 class="h-100 border-0"
-                :body-scrolling="isLargeScreen"
-                header-class="pb-0"
+                header-class="pb-0 d-flex"
                 body-class="py-2"
               >
                 <div class="bd-toc" />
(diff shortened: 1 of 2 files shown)


VividLemon:
I am not seeing an issue. How are you able to reproduce this?

dwgray:
Clean build from main. I've tried this on two different machines and done a git clean -d -f -x to clean the branch.  I guess I could try a new enlistment. But it reliably reproduced without these changes and these changes fix it, so I am struggling to figure out what I could be doing locally to induce the bug.

On the other hand, your changes are released, right? So one would expect the bug to show up on [Introduction | BootstrapVueNext (bootstrap-vue-next.github.io)](https://bootstrap-vue-next.github.io/bootstrap-vue-next/)?

VividLemon:
The issue is mainly related to `backdrop` becoming `hide-backdrop`. But its underlying system could be reworked to use the responsive prop, rather than the manual stuff its doing now

VividLemon:
This looks like it fixes everything, probably.
image

dwgray:
> This looks like it fixes everything, probably.

That gets the overlay stuff working, but it messes up the right sidebar

image

VividLemon (approved):

VividLemon:
Try now

dwgray:
> Try now

That gets everything but the styling of the right-nav toc. It looks like we don't have .offcanvas.offcanvas-start or .offcanvas.offfcanvas-end as parents of the toc.

image

I can fix this by adding .offcanvas.body to the list, but that feels kludgy.

dwgray:
> I can fix this by adding .offcanvas.body to the list, but that feels kludgy.

Actually, I was making an assumption based on some previous exploration. The thing that's missing is `offcanvas` because there is an `offcanvas-lg` instead? Possible fix:

image

VividLemon:
> > I can fix this by adding .offcanvas.body to the list, but that feels kludgy.
> 
> Actually, I was making an assumption based on some previous exploration. The thing that's missing is `offcanvas` because there is an `offcanvas-lg` instead? Possible fix:
> 
> image
> 
> I mean - this fixes the scenario I'm looking at, not sure about the robustness...

Yes. offcanvas-lg is the best substitute for what I did. Idrc about robustness in this case as it's unlikely the breakpoint will change

dwgray:
> Yes. offcanvas-lg is the best substitute for what I did. Idrc about robustness in this case as it's unlikely the breakpoint will change

Okay - I pushed that change.  Thanks!

VividLemon (approved):

PR #2009, opened 2024-07-09
docs: get markdown table formatting working with bootstrap

dwgray:
# Describe the PR

The markdown tables in the docs are unformatted. The default theme has custom styles for tables. It seems preferable to use bootstrap table stylings. This means adding the `table` and `table-striped` classes to the tables. The only way I could find to do that was to use a markdown-it extension. This seems to work fairly well, but I'm happy to take another approach if someone has a better idea.

Also, it looks like `turbo` is set to `latest` in the main `package.json` - I _think_ that's what's causing the cascade of changes in the `lock.yaml` file. I believe that also forced the change of `pipeline` to `tasks` in the turbo.config  and required that the main package have a name. I'm happy to back all those things out if I can get some direction on how to do that (I'd try to set turboto 1.^, but maybe there is a cleaner way?) Or happy to just move forward.

## Small replication

Before:
image

After:

image

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,4 +1,5 @@
 {
+  "name": "bootstrap-vue-next-monorepo",
   "private": true,
   "workspaces": [
     "packages/*",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1,4 +1,4 @@
-lockfileVersion: '6.0'
+lockfileVersion: '9.0'
 
 settings:
   autoInstallPeers: true
@@ -16,7 +16,7 @@ importers:
         version: 15.2.2
       turbo:
         specifier: latest
-        version: 1.10.7
+        version: 2.0.6
 
   apps/docs:
     devDependencies:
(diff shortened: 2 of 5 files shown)


turbo.json
```diff
@@ -1,6 +1,6 @@
 {
   "$schema": "https://turborepo.org/schema.json",
-  "pipeline": {
```
dwgray, line 3: This is a change in turbo 2.0 (see main comment)

pnpm-lock.yaml
dwgray, line 1: I think the cascading changes are due to turbo upgrade 2.0 (see main comment) - Obviously adding the one new package would cause some changes here.

package.json
```diff
@@ -1,4 +1,5 @@
 {
+  "name": "bootstrap-vue-next-monorepo",
```
VividLemon, line 2: Why

dwgray, line 2: Turbo 2.0 enforces a name on the main package (I think - I was certainly getting an error until I added this).

VividLemon (approved):

PR #2014, opened 2024-07-10
doc(BFormGroup): Update docs and component data

dwgray:
# Describe the PR

Complete the docs and component data for BFormGroup, I've also updated the parity report.

I believe this component may need a rework before we can call it complete -  at minimum, we should consider using script setup and creating a typescript interface. @VividLemon - Do you agree? If so, should I create an issue for this? I do have a note in the parity report, but I think it would be a good idea to track significant changes to the code with issues...

## Small replication

N/A

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)


apps/docs/src/data/components/formGroup.data.ts
```diff
@@ -7,99 +8,207 @@ export default {
...
         },
       },
       emits: [],
-      slots: [],
+      slots: [
+        {
+          name: 'default',
+          description: 'Content (form checkboxes) to place in the form checkbox group',
+          scope: [
+            {
```
dwgray, line 168: @VividLemon I'm wondering if it's worth extending the typing work that we (well, mostly you) did for props to cover emits and slots. It feels weird to have the scoped slots in the old format even though they are shaped the same as properties. If so, I'll take that on soon (I want to be judicious about how much time I'm spending on infrastructure vs. getting through the parity assessment/updates - so I'm going to aim for getting a few components covered between diving into infrastructure - we'll see how that goes)

VividLemon: Slot scopes and emits aren't really shared, they're more commonalities

VividLemon (approved):

VividLemon:
> I believe this component may need a rework before we can call it complete - at minimum, we should consider using script setup and creating a typescript interface. @VividLemon - Do you agree? If so, should I create an issue for this? I do have a note in the parity report, but I think it would be a good idea to track significant changes to the code with issues...

I'll do this later. It's not a trivial refactor

dwgray:
> I'll do this later. It's not a trivial refactor. At the time of looking at it I didn't know about `createReusableTemplate`, which simplifies things

Yeah - I'm not planning on doing major reworks of components as part of the docs/parity pass, nor am I expecting you to. 

This is more a question of how you'd like them tracked. Issues or just notes in the parity report?

The secondary question is, where is this component wrt blocking a v1 release? My thought is that for v1, the most important thing is that the API is stable to the best of our ability. e.g. we wouldn't expect sweeping breaking changes until some point (hopefully far in the future) when we would consider a v2. On that level, the only thing that I'd consider changing to get to v1 is the typing of the properties that are currently generically strings but in other places in the code are types eg: `export type Breakpoint = 'sm' | 'md' | 'lg' | 'xl' | 'xxl'` - I think I can fix those without any kind of refactor, but haven't really played with that yet. If you agree, I'll queue that up and handle it once I've got a bit more of the parity report under my belt and have a better idea of the patterns.

PR #2019, opened 2024-07-11
fix(BFormCheckbox): Handle contextual state === false correctly (Fixes #2017)

### LINKED ISSUE #2017: :state="false" on BFormCheckbox
### Describe the bug

Shouldn't the first checkbox on https://bootstrap-vue-next.github.io/bootstrap-vue-next/docs/components/form-checkbox.html#contextual-state-and-validation-example be red since the `state` prop is set to `false`?

I have a required checkbox that I want to highlight in red under certain circumstances.

This code didn't produce a checkbox at all on StackBlitz for some reason:

```html
<BFormCheckbox :state="false">Should be red</BFormCheckbox>
```

### Reproduction

N/A

### Used Package Manager

npm

### Validations

dwgray:
# Describe the PR

Checkbox `state === false` was resolving to undefined (which is effectively the null state) if there was no parent. Instead of using the parent value for all `falsey` values, I'm checking that the direct prop is `true` or `false` and only falling back to the parent if it's not.

## Small replication

See #2017 

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

### DIFF
--- a/packages/bootstrap-vue-next/src/components/BFormCheckbox/BFormCheckbox.vue
+++ b/packages/bootstrap-vue-next/src/components/BFormCheckbox/BFormCheckbox.vue
@@ -115,7 +115,8 @@ const classesObject = computed(() => ({
   inline: props.inline || (parentData?.inline.value ?? false),
   reverse: props.reverse || (parentData?.reverse.value ?? false),
   switch: props.switch || (parentData?.switch.value ?? false),
-  state: props.state || parentData?.state.value,
+  state:
+    props.state === true || props.state === false ? props.state : parentData?.state.value ?? null,
   size: props.size ?? parentData?.size.value ?? 'md', // This is where the true default is made
   buttonVariant: props.buttonVariant ?? parentData?.buttonVariant.value ?? 'secondary', // This is where the true default is made
   hasDefaultSlot: hasDefaultSlot.value,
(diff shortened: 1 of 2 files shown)


packages/bootstrap-vue-next/src/components/BFormCheckbox/BFormCheckbox.vue
```diff
@@ -115,7 +115,8 @@ const classesObject = computed(() => ({
   inline: props.inline || (parentData?.inline.value ?? false),
   reverse: props.reverse || (parentData?.reverse.value ?? false),
   switch: props.switch || (parentData?.switch.value ?? false),
-  state: props.state || parentData?.state.value,
+  state:
+    props.state === true || props.state === false ? props.state : parentData?.state.value ?? null,
```
VividLemon, line 119: Why not `props.state ?? parentData?.state.value ?? null`

dwgray:
When `props === false` and `parent === undefined`, `props.state ?? parentData?.state.value ?? null` evaluates to `null`, we want it to evaluate to `false` (that's what this bug is about)

So, if we did `props.state ?? parentData?.state.value ?? false`, that would catch the case that the bug is about, but if `props.state === null`, then we have the same problem (we get `false` when we want `null`)

Semantically, I think what we want to say is that if `props.state` is a boolean, use that, otherwise, use `parentData.state` and fall back to `null` if `parentData` doesn't exist. I don't see a cleaner way to state that, but I'm open to suggestions.

VividLemon (approved):

Issue #33, opened 2021-08-12
npm install error


bobaoapae:
> npm install
npm ERR! code ERESOLVE
npm ERR! ERESOLVE unable to resolve dependency tree
npm ERR!
npm ERR! While resolving: bootstrap-vue-3@0.0.1-alpha.3
npm ERR! Found: eslint@7.32.0
npm ERR! node_modules/eslint
npm ERR!   dev eslint@"^7.32.0" from the root project
npm ERR!
npm ERR! Could not resolve dependency:
npm ERR! peer eslint@">= 1.6.0 < 7.0.0" from @vue/cli-plugin-eslint@4.5.13
npm ERR! node_modules/@vue/cli-plugin-eslint
npm ERR!   dev @vue/cli-plugin-eslint@"~4.5.0" from the root project
npm ERR!
npm ERR! Fix the upstream dependency conflict, or retry
npm ERR! this command with --force, or --legacy-peer-deps
npm ERR! to accept an incorrect (and potentially broken) dependency resolution.
npm ERR!
npm ERR! See C:\Users\jvbor\AppData\Local\npm-cache\eresolve-report.txt for a full report.

npm ERR! A complete log of this run can be found in:
npm ERR!     C:\Users\jvbor\AppData\Local\npm-cache\_logs\2021-08-12T17_36_46_149Z-debug.log

Seems to be a issue with incompatible version of cli-serve and eslint, eslint 7 it's only supported on cli-service beta 5

mariusa:
vue-cli isn't required anymore, since vitejs works out of the box now and
it's faster. I propose to remove it, I don't see any advantage?
(Already switched my projects to vite)


Marius

bobaoapae:
I want to try do develop some things missing on this component for i can use in my project migrating to vue3 but i can't start due to error on run npm install

cdmoro:
Please, try using `yarn` to install the dependencies.

bobaoapae:
yarn install work, but will be nice to work with npm install to, i never used yarn before

mariusa:
@cdmoro Would you agree to remove vue-cli support? This would reduce dependencies (webpack is huge) and also solve this bug. Vitejs is the preferred way to run & build Vue 3 apps and it already works fine with bootstrap-vue-3 (I added the support). It's much faster too.

I can send a PR if you agree.

cdmoro:
The thing is we use vue-cli to build this project as a library.

[...]

PR #2036, opened 2024-07-13
feat(BFormDatalist): Initial implementation

dwgray:
# Describe the PR

## Small replication

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)


packages/bootstrap-vue-next/src/composables/useFormSelect.ts
```diff
@@ -25,13 +25,14 @@ export default (options: MaybeRefOrGetter, props: Record<string, unknown>) => {
     const html: string = get(option, props.htmlField as string)
     const disabled: boolean = get(option, props.disabledField as string)
 
-    const opts: undefined | (unknown | Record<string, unknown>)[] = props.optionsField
-      ? get(option, props.optionsField as string)
-      : undefined
+    const opts: undefined | (unknown | Record<string, unknown>)[] = get(
```
dwgray, line 28: I'm using this composable in Datalists, which don't support options groups, so they don't have the optionsField property.  There are probably other ways to get around this, but it seems like this is a more robust way to handle the default case for other uses of this composable anyway, unless I'm missing something?

packages/bootstrap-vue-next/src/composables/useFormSelect.ts
```diff
@@ -25,13 +25,14 @@ export default (options: MaybeRefOrGetter, props: Record<string, unknown>) => {
...
+      option,
+      (props.optionsField as string) ?? 'options'
+    )
 
     if (opts !== undefined) {
       return {
-        label: get(option, props.labelField as string) || text,
+        label: get(option, (props.labelField as string) ?? 'label') || text,
```
dwgray, line 35: same comment as with optionsField

packages/bootstrap-vue-next/src/components/BForm/BFormDatalist.vue
```diff
@@ -0,0 +1,56 @@
...
+
+const computedId = useId(() => props.id, 'datalist')
+
+const {normalizedOptions, isComplex} = useFormSelect(() => props.options, props)
+
+const normalizedOptsWrapper = computed(() => {
+  let opts = normalizedOptions.value
+  if (opts.find((opt) => isComplex(opt))) {
+    // eslint-disable-next-line no-console
+    console.warn('Datalist may not contain complex options', computedId.value)
```
dwgray, line 50: I'd like to use the same mechansisms we have in place for options lists, but datafields don't allow for options groups. I'm not seeing a consistent way to manage runtime errors for a case like this, happy with any guidance for managing this differently.

dwgray:
@VividLemon This is the first component I've written from scratch, so I'd appreciate any early feedback you have on this before I document it and write tests (also interested in what you'd test on this - this is basically a very paired-down version of `BFormSelect` and that doesn't have direct testing - most of the complex code is shared code that is under test).

BTW - I'm not planning on filling in significant sized holes as I'm going through the parity/documentation pass - I think it's more important that I get through a first round. However, this helper component is actually something that I use and is blocking my next round of page conversions - it just happened to be related to stuff that I'm touching with the parity work.

dwgray:
> @VividLemon This is the first component I've written from scratch...

This is ready to go as far as I'm concerned. But as I noted, this is my first component from scratch...

packages/bootstrap-vue-next/src/composables/useFormSelect.ts
VividLemon, line 28: `(unknown | Record<string, unknown>)[] = get(`
This is bad code on my part, as type 'unknown' already is inclusive of `Record<string, unknown>`

VividLemon: Seemingly the best option would be to pass it in the parent component. 

`useFormSelect(() => props.options, computed(() => (...props, optionsField: props.optionsField ?? 'options'}))`

etc etc

VividLemon: With this change this file would be pretty unchanged

packages/bootstrap-vue-next/src/components/BForm/BFormDatalist.vue
```diff
@@ -0,0 +1,56 @@
...
+  disabled: false,
+  disabledField: 'disabled',
+  htmlField: 'html',
+  id: undefined,
+  // eslint-disable-next-line vue/require-valid-default-prop
+  options: () => [],
+  textField: 'text',
+  valueField: 'value',
+})
+const props = useDefaults(_props, 'BFormSelect')
```
VividLemon, line 33: This seems like it needs a different key `BFormSelect`

packages/bootstrap-vue-next/src/components/BForm/BFormDatalist.vue
VividLemon: You'd have to show some code for what you tried.

I don't like this code here. 

I think if this component never has a situation where isComplex should be true, then it should never be an issue and this code should be removed. Ie if it does ever happen somehow, we should simply ignore it as its user error

Hypothetically this is something that would be validated by TS. Don't know how strongly typed that is though

VividLemon (changes requested):

packages/bootstrap-vue-next/src/components/BForm/BFormDatalist.vue
dwgray: Here's a fiddle https://jsfiddle.net/ohdwg/6tzu4nqc/2/ showing that options groups aren't supported in data lists.  They're also not documented on Mozilla for datalist: https://jsfiddle.net/ohdwg/6tzu4nqc/2/ while they are for  select: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select

I agree that typing the property would be the preferred way to go. But, the types have to be loose enough to allow substituting key names for the different fields (diabledField, htmlField, textField, valueField, labelField, optionsField).

I reworked this to silently ignore complex options and addressed the limitation in the documentation. I considered flattening the list (which is closer to what the native control does if you add in complex options), but then I'd have to add back in `labelField` and `optionsField` which seems like ugly complexity for something that is getting flattened back out. Short of pulling off some typescript voodoo, that's the best I can come up with.

packages/bootstrap-vue-next/src/composables/useFormSelect.ts
dwgray: Thanks - I like that better.

packages/bootstrap-vue-next/src/components/BForm/BFormDatalist.vue
dwgray: Fixed.

packages/bootstrap-vue-next/src/composables/useFormSelect.ts
dwgray: Sorry - yeah, I can change it to just unknown if you like. I can also take one more run at trying to more strictly type this - but I'd like you take on whether you think it's even doable before I spend more time on it (with the issue of the field names being fungible at runtime that I noted below).

dwgray: On closer examination, I believe `undefined | unknown[]` is a slight improvement to this declaration. I think that's the best we can do in that case.

packages/bootstrap-vue-next/src/components/BForm/BFormDatalist.vue
VividLemon: No, I mean it's currently BFormSelect, but seems to need `BFormDatalist`

dwgray: Sorry - something I don't understand is going on with github's view of this. If you look at [this commit](https://github.com/bootstrap-vue-next/bootstrap-vue-next/pull/2036/commits/9145882f360bfaed693326c4098f22221930668a) - I've changed line #33 of BFormDatlist.vue to 
`const props = useDefaults(_props, 'BFormDatalist')`, which is what you want, right? This comment is marked as outdated, so it's not updating the diff here? Not sure why.

packages/bootstrap-vue-next/src/components/BForm/BFormDatalist.vue
```diff
@@ -0,0 +1,54 @@
...
+  default?: (props: Record<string, never>) => any
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  first?: (props: Record<string, never>) => any
+}>()
+
+const computedId = useId(() => props.id, 'datalist')
+
+const {normalizedOptions, isComplex} = useFormSelect(
+  () => props.options,
+  computed(() => ({...props, optionsField: 'options', labelField: 'label'})).value
```
VividLemon, line 46: This is wrong. You should be passing in a Ref, not its .value =>

ie. if props changes, useFormSelect wont update

```suggestion
  computed(() => ({...props, optionsField: 'options', labelField: 'label'}))
```

VividLemon (approved):

packages/bootstrap-vue-next/src/composables/useFormSelect.ts
```diff
@@ -25,7 +25,7 @@ export default (options: MaybeRefOrGetter, props: Record<string, unknown>) => {
     const html: string = get(option, props.htmlField as string)
     const disabled: boolean = get(option, props.disabledField as string)
 
-    const opts: undefined | (unknown | Record<string, unknown>)[] = props.optionsField
+    const opts: undefined | unknown[] = props.optionsField
```
VividLemon, line 28: Some of the code for useFormSelect needs to be rewritten to support reactive inputs

This function
image
shadows this parameter
image
(Two duplicate variable names)

```ts
export default (options: MaybeRefOrGetter, props: Record<string, unknown>) => {
```
Options needs a generic input `MaybeRefOrGetter<unknown>` would suffice, unless there's some more strict input we can use

props should be `props: MaybeRefOrGetter<Record<string, unknown>>` which will likely need to be satisfied with

```ts
 const propsValue = readonly(toRef(props))
```

or

```ts
  const normalizeOption = (
    option: any
  ): ComplexSelectOptionRaw | SelectOption => {
    const propsValue = toValue(props) // <-- then use this in the function
    if (typeof option === 'string') {
```

I prefer the first though. Perhaps `normalizeOptions(toValue(options)))` should be

```ts
const optionsValue = readonly(toRef(options))
const normalizedOptions = computed(() => normalizeOptions(optionsValue.value))
```
Here as well though

VividLemon (changes requested):

packages/bootstrap-vue-next/src/composables/useFormSelect.ts
```diff
@@ -1,15 +1,20 @@
...
+) => {
   const isComplex = (option: unknown): option is ComplexSelectOptionRaw =>
     typeof option === 'object' && option !== null && 'label' in option
 
   const normalizeOption = (
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     option: any
   ): ComplexSelectOptionRaw | SelectOption => {
+    const propsValue = readonly(toRef(props)) // as Readonly<Record<string, unknown>>
+
```
VividLemon, line 17:
```suggestion
export default (
  options: MaybeRefOrGetter<Readonly<unknown[]>>,
  props: MaybeRefOrGetter<Record<string, unknown>>
) => {
  const propsValue = readonly(toRef(props)) // as Readonly<Record<string, unknown>>
  const optionsValue = readonly(toRef(options))

  const isComplex = (option: unknown): option is ComplexSelectOptionRaw =>
    typeof option === 'object' && option !== null && 'label' in option

  const normalizeOption = (
    // eslint-disable-next-line @typescript-eslint/no-explicit-any
    option: any
  ): ComplexSelectOptionRaw | SelectOption => {
```

packages/bootstrap-vue-next/src/composables/useFormSelect.ts
```diff
@@ -45,11 +50,12 @@ export default (options: MaybeRefOrGetter, props: Record<string, unknown>) => {
...
 
   const normalizeOptions = (
-    options: (unknown | Record<string, unknown>)[]
+    opts: Readonly<unknown[]>
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  ): (ComplexSelectOptionRaw | SelectOption)[] => options.map((option) => normalizeOption(option))
+  ): (ComplexSelectOptionRaw | SelectOption)[] => opts.map((option) => normalizeOption(option))
 
-  const normalizedOptions = computed(() => normalizeOptions(toValue(options)))
+  const optionsValue = readonly(toRef(options))
```
VividLemon, line 57:
```suggestion
```

VividLemon (changes requested):

packages/bootstrap-vue-next/src/composables/useFormSelect.ts
```diff
@@ -1,8 +1,14 @@
-import {computed, type MaybeRefOrGetter, toValue} from 'vue'
+import {computed, type MaybeRefOrGetter, readonly, toRef} from 'vue'
 import {get} from '../utils'
 import type {ComplexSelectOptionRaw, SelectOption} from '../types'
 
-export default (options: MaybeRefOrGetter, props: Record<string, unknown>) => {
+export default (
+  options: MaybeRefOrGetter<Readonly<unknown[]>>,
```
VividLemon, line 6:
```suggestion
  options: MaybeRefOrGetter<unknown[]>,
```

VividLemon, line 6:
```suggestion
  options: MaybeRefOrGetter<ReadonlyArray<unknown>>,
```

packages/bootstrap-vue-next/src/composables/useFormSelect.ts
```diff
@@ -45,11 +51,11 @@ export default (options: MaybeRefOrGetter, props: Record<string, unknown>) => {
   }
 
   const normalizeOptions = (
-    options: (unknown | Record<string, unknown>)[]
+    opts: Readonly<unknown[]>
```
VividLemon, line 54:
```suggestion
    opts: ReadonlyArray<unknown>
```

VividLemon (approved):

PR #2037, opened 2024-07-14
doc(BFormSpinbutton): Fill out component reference and validate docs

dwgray:
# Describe the PR

Fill out the component reference for spinbutton. In the process I validated the existing docs (did not find any issues) and updated the parity report.

## Small replication

N/A
## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

### DIFF
--- a/apps/docs/src/docs/migration-guide.md
+++ b/apps/docs/src/docs/migration-guide.md
@@ -96,6 +96,11 @@ Datalist and disabling mousewheel events are not yet implemented.
 
 [Options as an object](https://bootstrap-vue.org/docs/components/form-select#options-as-an-object) was deprecated in BootstrapVue and never implemented in BootstrapVueNext
 
+## BFormSpinbutton
+
+The locale property in BSVN only allows a for a single locale, while BSV allows for an array of locales. If this is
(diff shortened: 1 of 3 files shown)


apps/docs/src/docs/migration-guide.md
```diff
@@ -96,6 +96,11 @@ Datalist and disabling mousewheel events are not yet implemented.
 
 [Options as an object](https://bootstrap-vue.org/docs/components/form-select#options-as-an-object) was deprecated in BootstrapVue and never implemented in BootstrapVueNext
 
+## BFormSpinbutton
+
+The locale property in BSVN only allows a for a single locale, while BSV allows for an array of locales. If this is
```
dwgray, line 101: I'm not entirely certain what the expected behavior should be when setting multiple locales. Maybe it's a fallback list? In any case, I don't think this parity issue should be v1 blocking, since adding to the API shouldn't be a breaking change. But if there is a quick fix to get this to full parity, that would be another solution (I am just trying to limit the time that I dive into random issues).

VividLemon: I don't know, nor do I really know much about the spinbutton. It's a bit odd because the dev that implemented this pretty much copied the bv implementation. So idk, it's fine

VividLemon (approved):

Issue #2039, opened 2024-07-14
BFormGroup missing class="form-group" and bottom margin


jacobg:
### Missing feature is from

bootstrap-vue

### Docs

https://bootstrap-vue-next.github.io/bootstrap-vue-next/docs/components/form-group.html#component-reference

### Describe the missing feature

In original bootstrap-vue, the BFormGroup component created a <div> element with class="form-group", and that style had a 1rem bottom margin:
https://bootstrap-vue.org/docs/components/form-group

But in the current bootstrap-vue-next, no "form-group" class is declared, nor is there a bottom margin:
https://bootstrap-vue-next.github.io/bootstrap-vue-next/docs/components/form-group.html#component-reference

### Validations

- [X] Have tested with the latest version. This is still alpha version and sometime things change rapidly.
- [X] Follow our [Code of Conduct](https://github.com/bootstrap-vue-next/bootstrap-vue-next/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guide](https://github.com/bootstrap-vue-next/bootstrap-vue-next/blob/main/CONTRIBUTING.md).
- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A, please open a GitHub Discussion instead.
- [X] The provided link is to the documentation of the missing feature.

VividLemon:
I decided against this https://github.com/bootstrap-vue-next/bootstrap-vue-next/pull/2042#issuecomment-2229087517

PR #2042, opened 2024-07-15
fix(BFormGroup): Add form-group class + styling for compatibility addresses #2039

dwgray:
# Describe the PR

We were no longer adding the `form-group` class to the main `BFormGroup` element, since BS5 dropped this class `https://getbootstrap.com/docs/5.3/migration/#forms-1`

This PR adds `form-group` back and defines the class with the margin that BS4 used. This address #2039. I'm not 100% convinced that we want to do this in every case that comes up, but it's pretty straightforward in this case. @VividLemon?

## Small replication

Examples in the docs replicate this

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

### DIFF
--- a/packages/bootstrap-vue-next/src/components/BFormGroup/BFormGroup.vue
+++ b/packages/bootstrap-vue-next/src/components/BFormGroup/BFormGroup.vue
@@ -370,6 +370,7 @@ export default defineComponent({
         {
           'was-validated': props.validated,
         },
+        'form-group',
       ],
       'id': useId(() => props.id).value,
       'disabled': isFieldset ? props.disabled : null,
@@ -396,3 +397,9 @@ export default defineComponent({
   },
 })
 </script>
+
+<style lang="scss">
+.form-group {
+  margin-bottom: 1rem;
+}
+</style>


VividLemon:
> We were no longer adding the form-group class to the main BFormGroup element, since BS5 dropped this class https://getbootstrap.com/docs/5.3/migration/#forms-1

> This PR adds form-group back and defines the class with the margin that BS4 used. This address https://github.com/bootstrap-vue-next/bootstrap-vue-next/issues/2039. I'm not 100% convinced that we want to do this in every case that comes up, but it's pretty straightforward in this case. @VividLemon?

No, I do not want to do this

jacobg:
How would you suggest then we style the form group component? Currently it renders in the dom as a `<fieldset>` element with no class name at all.

VividLemon:
> How would you suggest then we style the form group component? Currently it renders in the dom as a `<fieldset>` element with no class name at all.

Are you unable to use a `class` attribute? (This component is old and badly written, so it might be impossible wothout an update)

jacobg:
I guess I could create a wrapper FormGroup class that wraps BFormGroup. I just thought that it was the customary practice to give every bootstrap component a class name? e.g., BButton is `<button>` element with class name "btn", etc.

PR #2066, opened 2024-07-20
fix(BFormInput,BFormTextArea)!: Deprecate trim, lazy, number props in favor of modifiers

dwgray:
# Describe the PR

These three properties were implemented in BSV because modifiers wouldn't work (at least according to the documentation). In BSVN they were partially declared and only `number` was actually implemented.

This PR removes all three properties from `BFormInput` and `BFormTextarea` and reactors the implementation of the modifiers to be shared between the two.

@VividLemon I ran into this inconsistency when doing the parity pass on `BFormTextarea` - I missed it on `BFormInput`. I don't have a strong opinion on deprecating or implementing these three props - slightly leaning towards deprecating, but if you prefer, I can easily just implement them. I do believe that we should treat all three the same way though. And I'd like your call on this before I go and fix up the documentation and parity reports for these two components.

## Small replication

```
<template>
  <BFormTextarea
    id="textarea"
    v-model.trim.lazy="textSelectEx"
    placeholder="Enter something..."
    rows="3"
  />

  *{{ textSelectEx }}*

  <BFormInput id="num" v-model.number="num" placeholder="Enter something..." />
  *{{ num }}*

  <BFormInput id="lazy-trim" v-model.lazy.trim="text" placeholder="Enter something..." />
  *{{ text }}*
</template>

<script setup lang="ts">
import {ref} from 'vue'
import {BFormTextarea} from './components'
const textSelectEx = ref('')

const text = ref('')
const num = ref(0)
</script>
```

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)


packages/bootstrap-vue-next/src/components/BFormInput/BFormInput.vue
```diff
@@ -44,39 +45,23 @@ const _props = withDefaults(defineProps<BFormInputProps>(), {
   form: undefined,
   formatter: undefined,
   id: undefined,
-  lazy: false,
```
dwgray, line 47: Notably, these three props weren't declared in `BFormInputProps` but also weren't throwing an error, so we're not getting type-checking at this level. Possibly a bug in vue? I'm not to concerned about that, it's more important that the type-checking works for the consumer of the component, which I believe it does. Is this related to what @anrolmar is asking about on discord?

packages/bootstrap-vue-next/src/utils/normalizeInput.ts
```diff
@@ -0,0 +1,15 @@
+import type {Numberish} from 'src/types'
+
+export default (
```
dwgray, line 3: I'm open to feedback on a better way to share this code. I couldn't add this function to useFormInputs because it takes the result of defineModel as inputs. The implementations in `BFromInput` and `BFormTextarea` were subtly different, so it seems like it is justifiable to share the implementation so they don't diverge again.

packages/bootstrap-vue-next/src/utils/normalizeInput.ts
```diff
@@ -0,0 +1,15 @@
+import type {Numberish} from 'src/types'
```
VividLemon, line 1: Make this a relative path and not `src/types`

VividLemon (changes requested):

packages/bootstrap-vue-next/src/components/BFormInput/BFormInput.vue
VividLemon: > Notably, these three props weren't declared in BFormInputProps but also weren't throwing an error, so we're not getting type-checking at this level

Were they being not checked in our lib, or in user libs?

Our lib makes sense, as eslint plugin vue lacks the ability to check for imported files.

User packages make sense because the object likely is consumed in the runtime. Both are bad things

dwgray: I was referring to the checking in our lib. We were setting defaults for properties that weren't defined by the type and there was no error or warning. Agreed that this is expected. Not a big deal, just have to be careful when modifying props/defaults to keep them in sync.

For external, I believe we are in general getting both design-time (in ts) and run-time type checking (in ts and js).

My understanding is that the vue compile uses typescript to infer the runtime types and in 3.3 they've removed the limitation on types defined in external files and handle most complex types, including the types we use for our components.  See: https://vuejs.org/guide/typescript/composition-api.html#syntax-limitations

image

In this example, we're correctly getting type errors for `id` and `lazy-formatter` `('null' is not assignable to 'string | undefined'` and `'string' is not assignable to 'boolean | undefined''`) and I get a runtime error for `lazy-formatter`
image
Unsurprisingly the runtime type-check doesn't warnon the `undefined`/`null` discrepancy of `id`

We are not getting any errors or warnings at either design or runtime when adding a property that's not declared in the typescript. e.g. `snap="foobar"` in the example above. But that seems to be vue behavior in general:

image

That's somewhat annoying in the case of having a typo in a property name - e.g. `lazy-formater="true"` would not show any errors or warning, but would just not set the property you're expecting. But I don't think there is anything we can do about that...

packages/bootstrap-vue-next/src/utils/normalizeInput.ts
dwgray: Done.

PR #2074, opened 2024-07-23
docs: kebab-case properties

dwgray:
# Describe the PR

While the [official recommendation](https://vuejs.org/style-guide/rules-strongly-recommended.html#prop-name-casing) for use of properties in SFCs is just that it be consistent, we're using kebab-case and BSV used kebab case, so this seems more readable and more likely what folks are searching for - especially since that's how we're using properties in the examples.

I did not touch slot names or event names since they are strings and could be addressed by being consistent in the data.

I would be interested in changing `update:modelValue` to `update:model-value` globally if you're all right with that @VividLemon 

## Small replication

See the component reference in docs

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

### DIFF
--- a/apps/docs/src/components/ComponentReference.vue
+++ b/apps/docs/src/components/ComponentReference.vue
@@ -154,7 +154,7 @@
                           v-for="scope in d.item.scope as SlotScopeReference[]"
                           :key="scope.prop"
                         >
-                          <code>{{ scope.prop }}</code>
+                          <code>{{ kebabCase(scope.prop) }}</code>
                           <code>: {{ scope.type }}</code>
                           <span v-if="!!scope.description"> - {{ scope.description }}</span>
                         </div>
@@ -162,7 +162,7 @@
                       <template #cell(args)="d">
                         <!-- eslint-disable-next-line prettier/prettier -->
                         <div v-for="arg in d.item.args as EmitArgReference[]" :key="arg.arg">
-                          <code>{{ arg.arg }}</code>
+                          <code>{{ kebabCase(arg.arg) }}</code>
                           <code>: {{ arg.type }}</code>
                           <span v-if="!!arg.description"> - {{ arg.description }}</span>
                         </div>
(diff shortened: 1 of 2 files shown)


apps/docs/src/utils/index.ts
```diff
@@ -45,3 +45,6 @@ export const omit = <
   Object.keys(objToPluck)
     .filter((key) => !keysToPluck.map((el) => el.toString()).includes(key))
     .reduce((result, key) => ({...result, [key]: objToPluck[key]}), {} as Omit<A, B[number]>)
+
```
dwgray, line 48: I pulled this from the BSV sources. It doesn't look like we've got testing infrastructure in place for docs code and I don't think this is worth doing that for...

apps/docs/src/components/ComponentReference.vue
```diff
@@ -154,15 +154,15 @@
...
+                          <code>{{ kebabCase(scope.prop) }}</code>
                           <code>: {{ scope.type }}</code>
                           <span v-if="!!scope.description"> - {{ scope.description }}</span>
                         </div>
                       </template>
                       <template #cell(args)="d">
                         <!-- eslint-disable-next-line prettier/prettier -->
                         <div v-for="arg in d.item.args as EmitArgReference[]" :key="arg.arg">
-                          <code>{{ arg.arg }}</code>
+                          <code>{{ kebabCase(arg.arg) }}</code>
```
dwgray, line 165: I'm being a bit inconsistent about what is addressed in code vs. data. I could pull this and scope.prop from the code and address via data to keep things consistent (or address the event name in code) happy to do either if there is a preference.

VividLemon:
@dwgray 

> I would be interested in changing update:modelValue to update:model-value globally if you're all right with that @VividLemon

Yes

apps/docs/src/components/ComponentReference.vue
VividLemon: I don't understand the issue. I have no preference

Issue #1913, opened 2024-05-23
feat!: props composition > prop inheritance


VividLemon:
### Clear and concise description of the problem

Many components extend other components. Instead, what we should do is allocate a prop that combines those other props all in one. Leading to more expected behavior (For example, BAvatar had an icon prop, BLink had an icon prop. Who's do you use? [Although BAvatar.icon is gone now, it's an example])

### Suggested solution

The extended BLink props turn into `linkProps: BLinkProps`. With exceptions to Exclude elements

cdmoro:
My two cents, I do not think the final user wants/needs to know that the link used in a component is a `BLink` under the hood. In any case, we can type the props in a way that is scalable for us, for example, if a component uses two props from other component we can have something like:

``` typescript
export interface MyComponent {
  otherProp: string
  href: BLinkProps["href"]
  rel: BLinkProps["rel"]
}
```

Following the approach you propose, components will look like this and I definitely think we should create the least object-like props.

``` vue
<BAvatar
  :linkProps="{ href: 'https://www.vuejs.org', rel: '_blank' }"
  alt="Vue.js"
  badge="foo"
/>
```

Apart from that, it would be nice to have a "BoostrapVueNext desition record" everytime we decided to make a breaking change, with this if we need to revisit a desition we have them documented in one place.

VividLemon:
The issue is that blink has like 40 props, which when mixed in a flattened way with other other components, it makes it difficult to find actual props from the component. 

However this doesn't extend to actual extended values. Typically blink is the root component, so it makes sense for it to be an extension rather than composited (is that the right word?). My issue is more with nested component structures like BProgress/BProgressBar or Bavatar/BBadge. Where its not the main focus of the component being consumed

@xvaara

PR #2093, opened 2024-08-07
docs(BAvatar): parity pass

dwgray:
# Describe the PR

- Fix doc bugs.
- Fill out the component reference
- Write migration guide
- Update the parity report

I have not updated the `BLink` related properties for two reasons
1. I think it's better to document BLink, and then we can backfill its usage here
2. I'm not entirely clear where we landed  #1913 @VividLemon - did you land on just keeping things as they are? If so, what happens with properties like `variant` that are relevant to both `BLink` and `BAvatar`?

## Small replication

N/A

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)


dwgray:
@VividLemon Do you know off the top of your head if router links are supported by BAvatar? In taking a quick look at BLink it seems like they should be. If that's the case I'll put this PR into draft mode and tackle BLink & https://bootstrap-vue.org/docs/reference/router-links next. Any thoughts/references you have beyond the code on what BSVN supports for router-links would be really helpful.  Thanks!

VividLemon:
> Do you know off the top of your head if router links are supported by BAvatar?

BAvatar uses BLink which can use RouterLink

VividLemon:
> m not entirely clear where we landed https://github.com/bootstrap-vue-next/bootstrap-vue-next/issues/1913 @VividLemon - did you land on just keeping things as they are? If so, what happens with properties like variant that are relevant to both BLink and BAvatar?

It kind of seems like too much. Idrk. Maybe it could be a good addition idrk.

dwgray:
> > Do you know off the top of your head if router links are supported by BAvatar?
> 
> BAvatar uses BLink which can use RouterLink

Thanks, I cleaned up the docs to reflect this and refactored, so that BAvatar uses the BLink property reference as an extension. I think that makes this good to go. I'll document the BLink soon, but I don't think having those props empty should block this PR.

dwgray:
> > m not entirely clear where we landed #1913 @VividLemon - did you land on just keeping things as they are? If so, what happens with properties like variant that are relevant to both BLink and BAvatar?
> 
> It kind of seems like too much. Idrk. Maybe it could be a good addition, I'm not sure

Yeah, I agree that it's a stretch. And with the ability to document extensions, I think this works reasonably well in general. In this particular case, the `variant` prop on `BAvatar` hides the `variant` prop on `BLink`, so for a text avatar with a link, the link is unreadable (take a look at the example in the docs). This is a pretty corner case, I've opened #2102 to address it in the future.

VividLemon (approved):

PR #2112, opened 2024-08-12
fix(BLink): fix link-opacity when no underline variant specified

dwgray:
# Describe the PR

`link-underline-opacity`  requires that either a `link-underline-{variant}` or `link-underline` is specified. This PR handles the latter case. https://getbootstrap.com/docs/5.3/utilities/link/#underline-opacity

Also, alphabetized properties.

## Small replication

See the docs

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)


packages/bootstrap-vue-next/src/components/BLink/BLink.vue
```diff
@@ -119,6 +119,7 @@ const computedClasses = computed(() => ({
   [`link-underline-${props.underlineVariant}`]: props.underlineVariant !== null,
   [`link-offset-${props.underlineOffset}`]: props.underlineOffset !== undefined,
   [`link-offset-${props.underlineOffsetHover}-hover`]: props.underlineOffsetHover !== undefined,
+  ['link-underline']: props.underlineVariant === null && props.underlineOpacity,
```
VividLemon, line 122: if props.underlineOpacity is 0, this seems like it's going to not be expected. But maybe that's the point?

dwgray: Yep - absolutely right. I also missed the hover case. Both fixed + a couple of typos in the related docs cleaned up.

VividLemon (approved):

Issue #2124, opened 2024-08-15
BForm component is unconditionally preventing default submit behavior


dwgray:
### Describe the bug

[BForm](https://github.com/bootstrap-vue-next/bootstrap-vue-next/blob/main/packages/bootstrap-vue-next/src/components/BForm/BForm.vue) is capturing the `submit` event and preventing default:

``` vue
<template>
  <form
    :id="props.id"
    ref="element"
    :novalidate="props.novalidate"
    :class="computedClasses"
    @submit.prevent="submitted"
  >
    <slot />
  </form>
</template>
```

It looks like this was added here:

[commit e0d25b1](https://github.com/bootstrap-vue-next/bootstrap-vue-next/commit/e0d25b116e1c759b93b7cee056f3b4725dbcfe40#diff-25c62253a4ce5738f3f910bb45a128c61eb4c91361e029a0ed5338e1f556bfae)

This seems wrong. The only way I could work around this was to implement a submit handler and explicitly call submit on the form. If we want the out of the box behavior to be to prevent default, we should provide a way to override that, but I would lean towards just getting rid of the `prevent` here. But I don't know what other implications that might have. I will note that I ran into this while converting a component from BSV, so while I haven't confirmed 100%, my guess is that this is a compatibility issue as well.

### Reproduction

https://stackblitz.com/edit/github-agbcoe?file=src%2Fcomponents%2FComp.vue

### Used Package Manager

pnpm

### Validations

- [X] Have tested with the latest version. This is still alpha version and sometime things change rapidly.
- [X] Follow our [Code of Conduct](https://github.com/bootstrap-vue-next/bootstrap-vue-next/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guide](https://github.com/bootstrap-vue-next/bootstrap-vue-next/blob/main/CONTRIBUTING.md).
- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A, please open a GitHub Discussion instead.
- [X] The provided reproduction is a [minimal reproducible](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

VividLemon:
> getting rid of the prevent here

This

PR #2125, opened 2024-08-15
fix(BForm): Remove confusing prevent default on form submit - Addresses #2124

dwgray:
# Describe the PR

I can't find any reason to override the submit event's default behavior on `BForm`. I think the best way to address #2124 is to remove the custom event handler and let the native event propagate. 

I might be missing something about how the event propagation in Vue3 works, but I tested with a simple form and the event is coming through the way I expect.

## Small replication

See issue

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

### DIFF
--- a/packages/bootstrap-vue-next/src/components/BForm/BForm.vue
+++ b/packages/bootstrap-vue-next/src/components/BForm/BForm.vue
@@ -1,11 +1,5 @@
 <template>
-  <form
-    :id="props.id"
-    ref="element"
-    :novalidate="props.novalidate"
-    :class="computedClasses"
-    @submit.prevent="submitted"
-  >
+  <form :id="props.id" ref="element" :novalidate="props.novalidate" :class="computedClasses">
     <slot />
   </form>
 </template>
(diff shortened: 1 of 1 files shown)


VividLemon (approved):

kennyki:
Is this entirely removing the submit event handling? The change is supposed to be `@submit.prevent="submitted"` -> `@submit="submitted"` right?

kennyki:
> Is this entirely removing the submit event handling? The change is supposed to be `@submit.prevent="submitted"` -> `@submit="submitted"` right?

I was facing an issue with form submission - it was performing HTML submission even with `<b-form @submit.prevent="handleSubmission">`

I figured out that it's an issue with using `@vue/compat`. Removing that or using`compatConfig: { MODE: 3 }` will work

gtbuchanan:
The incremental solution for the `@vue/compat` problem is to make sure `COMPILER_V_ON_NATIVE` is enabled and then use `@submit.native` (likely `@submit.native.prevent` in your case).

PR #2143, opened 2024-08-23
fix: pass attributes through in BDropdownItemButton

dword-design-honestly:
# Describe the PR

Currently, attributes passed to BDropdownItemButton are swallowed. But we want to add them to the list element like it was in BootstrapVue.

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

### DIFF
--- a/packages/bootstrap-vue-next/src/components/BDropdown/BDropdownItemButton.vue
+++ b/packages/bootstrap-vue-next/src/components/BDropdown/BDropdownItemButton.vue
@@ -29,6 +29,7 @@ const _props = withDefaults(defineProps<BDropdownItemButtonProps>(), {
   buttonClass: undefined,
   disabled: false,
   variant: null,
+  wrapperAttrs: undefined,
 })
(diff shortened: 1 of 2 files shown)


VividLemon (changes requested):
Similar to other components the `class` attribute gets applied to the `li` and the rest of the attrs get applied to the `button`

inherit attrs is still probably needed here, but you likely should do

```ts
const { class: wrapperClass, ...attrs } = useAttrs()

<li :class="wrapperClass"> <button v-bind="attrs>
```
Similar to other components that implement this, `buttonClass` is (already implemented) and applies to the button, but we should also add `wrapperAttrs` to v-bind to the li component. 

(Searching for usages of `useAttrs` should point you in the direction of the standard implementation for components like this)

Reference: 
image

dword-design-honestly:
@VividLemon I did the changes as suggested 👍.

dword-design-honestly:
@VividLemon Can we have a version published since this is blocking us shipping sth.? 😊

PR #2169, opened 2024-09-03
fix(BTabs): fix recursion on titleLinkAttrs

### LINKED ISSUE #2167: title-link-attrs in BTab leads to infinite recursive updates
### Describe the bug

When adding title-link-attrs to BTab, we get an infinite update recursion.

```
Uncaught (in promise) Maximum recursive updates exceeded. This means you have a reactive effect that is mutating its own dependencies and thus recursively triggering itself. Possible sources include component template, render function, updated hook or watcher source function.
```

### Reproduction

https://github.com/dword-design-honestly/demo-bootstrap-vue-next-title-link-attrs

### Used Package Manager

pnpm

### Validations

xvaara:
# Describe the PR

closes #2167 

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

### DIFF
--- a/packages/bootstrap-vue-next/src/components/BTabs/BTab.vue
+++ b/packages/bootstrap-vue-next/src/components/BTabs/BTab.vue
@@ -67,7 +67,7 @@ const tab = computed(
       title: props.title,
       titleComponent: slots.title,
       titleItemClass: props.titleItemClass,
-      titleLinkAttrs: props.titleLinkAttrs,
+      titleLinkAttrs: toRef(() => props.titleLinkAttrs),
       titleLinkClass: props.titleLinkClass,
       onClick,
       el: el.value,
--- a/packages/bootstrap-vue-next/src/components/BTabs/BTab.vue
+++ b/packages/bootstrap-vue-next/src/components/BTabs/BTab.vue
@@ -66,9 +66,9 @@ const tab = computed(
       disabled: props.disabled,
       title: props.title,
       titleComponent: slots.title,
-      titleItemClass: props.titleItemClass,
-      titleLinkAttrs: toRef(() => props.titleLinkAttrs),
-      titleLinkClass: props.titleLinkClass,
+      titleItemClass: () => props.titleItemClass,
+      titleLinkAttrs: () => props.titleLinkAttrs,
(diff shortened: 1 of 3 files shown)


packages/bootstrap-vue-next/src/components/BTabs/BTab.vue
```diff
@@ -67,7 +67,7 @@ const tab = computed(
       title: props.title,
       titleComponent: slots.title,
       titleItemClass: props.titleItemClass,
-      titleLinkAttrs: props.titleLinkAttrs,
+      titleLinkAttrs: toRef(() => props.titleLinkAttrs),
```
xvaara, line 70: tested, this works if this line is just `titleLinkAttrs: () => props.titleLinkAttrs` and v-bind is just `v-bind="tab.titleLinkAttrs?.()"`

xvaara: @VividLemon which would be better? it seems v-bind with and object causes all the childs props to update which updates v-bind object and loops.

xvaara:
> Note that this is also the case for other props like titleLinkClass if an object is passed.

Nice catch! I'll fix them too.

dword-design-honestly:
Let me know if I can help here 🙂.

packages/bootstrap-vue-next/src/components/BTabs/BTab.vue
xvaara:
```
<b-tabs>
    <b-tab title="Tab 1" :title-link-attrs="{ foo: 'bar'}" />
  </b-tabs>
 ```
does not work. And I think that is what most people will use.

xvaara: Maximum recursive updates exceeded in component <BTabs>. This means you have a reactive effect that is mutating its own dependencies and thus recursively triggering itself. Possible sources include component template, render function, updated hook or watcher source function.

https://stackblitz.com/github/dword-design-honestly/demo-bootstrap-vue-next-title-link-attrs

VividLemon (approved):
This doesn't seem like the best way to fix this issue. But seems fine

PR #2177, opened 2024-09-05
docs(BImg): Parity pass

dwgray:
# Describe the PR

This is the parity pass for `BImg` and B`ImgCard`

- Filled out component references
- Refactored docs to put examples in separate files
- Fixed some inconsistencies in the docs
- Updated the parity spreadsheet.
- Refactored the doc utilities to allow me to use `buildCommonProps` when defining `imageProps`

I also marked `BImgLazy` and `BImgCardLazy` as deprecated - I believe they aren't needed in modern browsers where the `loading` attribute is implemented. There is a note in the docs that we intend to implement placeholder slots, but from my reading of BSV, they weren't implemented in the `*Lazy` variations, so this would be an enhancement not a parity issue.

## Small replication

N/A

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)


apps/docs/src/data/components/image.data.ts
```diff
@@ -1,98 +1,24 @@
...
-            type: 'Numberish',
-            default: undefined,
-          },
-          roundedEnd: {},
-        } satisfies Record<keyof BvnComponentProps['BImg'], PropertyReference>,
+        '': imageProps,
       },
       emits: [
         {
+          event: 'load',
```
dwgray, line 13: I initially pulled this out based on my reading of https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/loading - there is no mention of a load event other than Window.load. But testing the code, we do get a load event fired for every image. Based on testing though, we do get a load event fired for every image load (both proactive and lazy). It looks like there is a native onLoad event on img tags, which I presume vue.js it using to fire a load event, but I'm not seeing that documented anywhere - this may just be a hole in my knowledge and I'm looking in the wrong places. In any case, I'm putting this back in, let me know if there are details that I'm missing...

VividLemon: There's a section detailing the load event on the page you posted.

Although practically the event doesn't need to be declared as it should bubble up naturally

dwgray: I must be missing something. The place I found on that page is in the [Timing of the load event](https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/loading#timing_of_the_load_event) section. I read that as describing the behavior of the [`Window.load`](https://developer.mozilla.org/en-US/docs/Web/API/Window/load_event) event, which is not what I'm interested in here. The bubble-up event is a load event on the `` element. Doesn't really matter though - the point is that the event is there and it's bubbling up.

That does bring up the question of if/when we document bubble-up events. I've been doing that when it's consistent (e.g. it doesn't depend on a wrapper component being in the middle or not) and core to the functionality of the component I'm documenting. The `load` event here meets my criteria for documenting even though we're not explicitly declaring it, but let me know if you have a different preference. Or, if we should flag bubble-up events as such in the documentation...

VividLemon (approved):

Issue #2122, opened 2024-08-14
BBadge incorrectly declares bg-variant and text-variant properties


dwgray:
### Describe the bug

I'm pretty sure the bootstrap badge component doesn't respect the `text-*` and `bg-*` classes (see the reproduction below). If I'm right, we shouldn't be exposing the `bg-variant` and `text-variant` properties. I think this means removing the us of `useColorVariantClasses` and manually implementing the handling of the `variant` property?

### Reproduction

https://stackblitz.com/edit/github-bserkl?file=src%2Fcomponents%2FComp.vue

### Used Package Manager

pnpm

### Validations

- [X] Have tested with the latest version. This is still alpha version and sometime things change rapidly.
- [X] Follow our [Code of Conduct](https://github.com/bootstrap-vue-next/bootstrap-vue-next/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guide](https://github.com/bootstrap-vue-next/bootstrap-vue-next/blob/main/CONTRIBUTING.md).
- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A, please open a GitHub Discussion instead.
- [X] The provided reproduction is a [minimal reproducible](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

VividLemon:
Not quite. This should be something written somewhere in the documentation among components with a default variant prop.

When you use prop variant, it will overwrite props textVariant & bgVariant. So when you have default props like BBadge's `variant: 'secondary'` you will only get `text-bg-secondary`

In order to get your 
`<BBadge text-variant="success" bg-variant="danger"> testing </BBadge>`
example to work, you would need to do
`<BBadge text-variant="success" bg-variant="danger" :variant="null"> testing </BBadge>`

There isn't a great way to fix this manually for users, I think the best is to just have some resource that tells why it doesn't work. BAvatar should suffer from the same issue

dwgray:
Thanks for the explanation. #2200 is my attempt to address this in documentation - and you're right `BAvatar` has a similar issue and there was a visible bug in the docs if you looked carefully enough. By my reading `BButton` also has this issue.

VividLemon:
https://github.com/bootstrap-vue-next/bootstrap-vue-next/pull/2200#issuecomment-2351273613

PR #2200, opened 2024-09-15
doc(variant): interaction between variant, bg-variant & text-variant

dwgray:
# Describe the PR

This PR addresses #2122 among other things. Since the bootstrap 5 behavior makes `bg-variant` and `text-variant` override `variant` (it's actually `bg-*` and `text-*` classes overriding `text-bg-*` classes) - the most straightforward was to solve this problem is to just all of the variant props set their respective classes and let bootstrap sort it out.

## Small replication

See #2122

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

### DIFF
--- a/apps/docs/src/docs/components/avatar.md
+++ b/apps/docs/src/docs/components/avatar.md
@@ -232,27 +232,28 @@ Use the `variant` prop to specify one of Bootstrap theme variant colors. The def
   </template>
 </HighlightCard>
 
-When displaying text inside the avatar, text colors are calculated based on the `variant` prop. The result is either `light` or `dark`. You can override the calculated text color by manually specifying the `bg-variant` and `text-variant` props. In order to do so, you must override the default for the variant prop by
-setting it to null - `:variant='null'`
+When displaying text inside the avatar, text colors are calculated based on the `variant` prop. The result is either `light` or `dark`. You can override the calculated text color by manually specifying the `bg-variant` and `text-variant` props. Note that
+`bg-variant` and `text-variant` take precedence over `variant`. See the
+[Color Variant Reference](/docs/reference/color-variants#variant-interactions) for details.
 
 <HighlightCard>
   <div class="d-flex" style="column-gap: 1%;">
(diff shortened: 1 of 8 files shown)


VividLemon:
On the other hand, using `variant` with textvariant or bgvariant wouldn't make sense. It could make sense to flip the `[`text-bg-${props.variant}`]: props.variant !== null,` to `[`text-bg-${props.variant}`]: props.variant !== null && (textVariant || bgVariant)`, check

dwgray:
> On the other hand, using `variant` with textvariant or bgvariant wouldn't make sense. It could make sense to flip the `[`text-bg-${props.variant}`]: props.variant !== null,` to `[`text-bg-${props.variant}`]: props.variant !== null && (textVariant || bgVariant)`, check

I agree that something like this could work better. I started to go down that path, and got tied up a bit because no matter what you do, we'll have some non-intuitive interactions.

Is what you're proposing that we set `text-bg-variant` if and only if the `variant` prop is set and neither the `text-variant` nor the `bg-variant` prop is set? Wouldn't that be?
```
[`text-bg-${props.variant}`]: props.variant !== null && !(textVariant || bgVariant),
```

If that is what you're thinking, then the only slightly odd thing is that if someone sets all three `variant` props, then `variant` will be ignored, but that seems like a lot more of a corner case than what we have now. Which is pretty much what you said above, I'm just processing...

I'll set this to draft and try going down that path (hopefully tomorrow)...

VividLemon:
Hypothetically one could remove the null check all together and just see what bootstrap does

dwgray:
> Hypothetically one could remove the null check all together and just see what bootstrap does

[Playing with BS5](https://stackblitz.com/edit/usb5of?file=index.html), `text-variant` and `bg-variant` take precedence over `bg-text-variant`. So I think the best course here is to just set the classes without trying to be smart and let BS5 handle it. I've probably over-documented the behavior, since it's fairly straightforward, but since I had most of this in place I just modified based on describing the BS5 behavior.

VividLemon (approved):

PR #2219, opened 2024-09-21
 fix(useFocusTrap): dont deactivate focus on esc fixes #2144 

### LINKED ISSUE #2144: BModal deactivates focus trap on esc key
### Describe the bug

When using `BModal` with `noCloseOnEsc` set to `true` or when one of the modal's child elements wants to handle esc key and prevent dialog closure manually, pressing esc key inside the modal dialog will always deactivate the focus trap.

In this scenario scenario, the modal dialog will not close but the focus will not remain within the modal element as intended. The focus will switch to the modal element and the currently focused element inside the modal (e.g input) will lose focus (only on first time esc is pressed, after that focus trap not active anymore).

Solution:
Pass `false` to `escapeDeactivates` option to _vueuse_'s `useFocusTrap()` (which will pass it to _focus-trap_ library). Since the modal dialog closes anyway when `noCloseOnEsc` is `false` and modal dialog should never lose focus otherwise, making it the default behavior makes sense.

### Reproduction

https://stackblitz.com/edit/github-cw3lum?file=src%2FApp.vue

### Used Package Manager

npm

### Validations

VividLemon:
# Describe the PR

## Small replication

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

### DIFF
--- a/packages/bootstrap-vue-next/src/components/BTable/BTableLite.vue
+++ b/packages/bootstrap-vue-next/src/components/BTable/BTableLite.vue
@@ -302,7 +302,7 @@ const generateDetailsItem = (item: TableItem): [object, boolean | undefined] => 
   item,
   item._showDetails,
 ]
-const detailsMap = ref(new WeakMap())
+const detailsMap = ref(new WeakMap<object, boolean | undefined>())
 watch(
   () => props.items,
   (items) => {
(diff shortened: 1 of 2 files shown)


packages/bootstrap-vue-next/src/components/BTable/BTableLite.vue
```diff
@@ -302,29 +302,16 @@ const generateDetailsItem = (item: TableItem): [object, boolean | undefined] =>
...
-        if (isTableItem(el)) {
-          acc.push(generateDetailsItem(el))
-        }
-        return acc
-      },
-      [] as [object, boolean | undefined][]
-    )
-  )
-)
+const detailsMap = ref(new WeakMap())
```
VividLemon, line 305:
```suggestion
const detailsMap = ref(new WeakMap<object, boolean | undefined>())
```

PR #2251, opened 2024-10-07
feat(useToast)!: rename useToast to useToastController

VividLemon:
# Describe the PR

## Small replication

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)


apps/docs/src/components/TableOfContentsNav.vue
```diff
@@ -147,7 +147,9 @@ const composablesList: {name: string}[] = [
   {name: 'useColorMode'},
   {name: 'useModal'},
   {name: 'useModalController'},
-  {name: 'useToast'},
+  {name: 'useToastController'},
+  {name: 'useTooltipController'},
+  {name: 'usePopoverController'},
```
VividLemon, line 152: @xvaara Is this better off as a single composable which has `tooltip()` `popover()` methods?

packages/bootstrap-vue-next/src/components/BDropdown/BDropdown.vue
```diff
@@ -83,22 +83,17 @@ import {BvTriggerableEvent} from '../../utils'
...
   boundary: 'clippingAncestors',
   boundaryPadding: undefined,
-  center: false,
   teleportTo: undefined,
   teleportDisabled: false,
   disabled: false,
-  dropend: false,
-  dropstart: false,
-  dropup: false,
-  end: false,
```
xvaara, line 101: This is a breaking change from previous. And change from bsv2. But it's clearer to have them all the same.

packages/bootstrap-vue-next/src/types/ComponentOrchestratorTypes.ts
```diff
@@ -36,38 +38,32 @@ export type ToastOrchestratorShowParam = {
   props?: MaybeRefOrGetter<PublicOrchestratedToast>
 }
 
-export type PrivateOrchestratedTooltip = Omit<BTooltipProps, 'modelValue' | 'reference'> & {
-  _reference: Exclude<BTooltipProps['reference'], undefined>
+export type PrivateOrchestratedTooltip = Omit<BTooltipProps, 'modelValue' | 'target'> & {
```
xvaara, line 41: Tooltip/popover can have different target and reference (ie trigger tooltip from one element and show it related to another)

VividLemon: Does one or both need to be set?

Which is which

packages/bootstrap-vue-next/src/components/BDropdown/BDropdown.vue
VividLemon: This is a fine breaking change imo that I added

apps/docs/src/docs/components/button-group.md
```diff
@@ -138,13 +138,13 @@ that split dropdown menus are not supported when prop `vertical` is set.
 <HighlightCard>
   <BButtonGroup>
     <BButton>Button</BButton>
-    <BDropdown right text="Menu">
+    <BDropdown placement="right"text="Menu">
```
tbl0605, line 141: There seems to be missing spaces between the 2 attributes ^^

packages/bootstrap-vue-next/src/types/ComponentOrchestratorTypes.ts
xvaara: if no reference is set, it's set to the same as target (the 99% use case). If it's changed, you need to rebind the popover.

VividLemon: @xvaara Can you work on pushing to this branch for any changes you'd like to see for the composable & orchestrator? One issue that popped in my mind is how do we handle when a target component is unmounted but still in the composable stack... Not sure if the overhead of checking if the component is removed is worth it just lingering in the stack

VividLemon: @xvaara On second thought, I'm just going to release 0.25 and not write any docs on it and keep it more or less "dark"

xvaara: I'll take a look tonight. 

Can we add a onUnmount(() => remove(self)), to the popover/tooltip functions? I think that would cover most cases. Or are you meaning it's added to a another component, basically using a templateRef or dom query? Then we could add a mutation observer to the dom element, and remove it from the stack when it's removed from dom.

PR #2311, opened 2024-10-28
feat!: remove html props -- use equivalent slots fixes #1930

### LINKED ISSUE #1930: we should sanitize v-html props in components or remove them
There are many places that accept v-html input and the input isn't sanitized for xss attacks, we leave this to the user. I think we should remove all the v-html places since we have slots everywhere that the user can use to shoot them self in the foot.

### LINKED ISSUE #2333: Add wrapper-class to BNavForm
### Missing feature is from

bootstrap-vue

### Docs

https://bootstrap-vue.org/docs/components/nav#comp-ref-b-nav-form-props

### Describe the missing feature

BSV has  `form-class` prop to explicitly add classes to the form object. In our case, we're binding the $attrs to the `BForm`, so for consistency/pairty with BSV I believe we should ad a `wrapper-class` prop to `BNavForm`.

I don't believe this is a blocker to release, since adding this should not be a breaking change, but it came up in the parity pass.

### Validations

- [X] The provided link is to the documentation of the missing feature.

VividLemon:
refactor: useTemplateRef

@xvaara Please double check the changes to popover html usage

fixes #2333


packages/bootstrap-vue-next/src/types/ComponentProps.ts
```diff
@@ -241,7 +240,6 @@ export interface BFormCheckboxGroupProps {
 
 export interface BFormDatalistProps {
   disabledField?: string
-  htmlField?: string
```
VividLemon, line 244: These are the major breaking changes. The rest are simply use the slot, but since these "groups" have some complex logic on how they v-for their sub-component, this would require them to essentially rebuild the sub component. Not sure how I feel about "groups" with html being removed. @xvaara @dwgray Thoughts?

dwgray: Isn't this similar to a table? Could we have a scoped slot passing in the data for each item. Then the use would be something like:
```
<BFormCheckboxGroup ...>
  <template #default="{ props, text }">
    {{ text }} (<b>{{props.value}}</b>)
  </template>
</BFormCheckboxGroup>
```

packages/bootstrap-vue-next/src/components/BForm/BFormDatalist.vue
```diff
@@ -3,11 +3,7 @@
     <slot name="first" />
     <template v-for="(option, index) in normalizedOptsWrapper" :key="index">
       <BFormSelectOption :value="option.value" :disabled="option.disabled">
-        <!-- eslint-disable-next-line vue/no-v-html -->
-        <span v-if="!!option.html" v-html="option.html" />
-        <template v-else>
-          {{ option.text }}
-        </template>
+        {{ option.text }}
```
xvaara, line 6:
```suggestion
        <slot v-bind="option">
          {{ option.text }}
        </slot>
```

and then change the slot below to name="last" Or this part to name="option"

packages/bootstrap-vue-next/src/types/ComponentProps.ts
xvaara:
```vue
  <template #option="{ option }">
    {{ option.text }} (<b>{{option.value}}</b>)
  </template>
  ```

Comment (VividLemon) [reply packages/bootstrap-vue-next/src/types/ComponentProps.ts:244]: Yes I think this would work

File: packages/bootstrap-vue-next/src/components/BFormSelect/BFormSelect.vue [hunk 3]
```diff
@@ -21,15 +21,10 @@
...
-        :html-field="props.htmlField"
         :disabled-field="props.disabledField"
       />
       <BFormSelectOption v-else :value="option.value" :disabled="option.disabled">
-        <!-- eslint-disable-next-line vue/no-v-html -->
-        <span v-if="!!option.html" v-html="option.html" />
-        <template v-else>
-          {{ option.text }}
-        </template>
+        {{ option.text }}
```
xvaara, line 27: Shouldn't this also have the option() slots logic?

packages/bootstrap-vue-next/src/components/BFormSelect/BFormSelectOptionGroup.vue
```diff
@@ -8,11 +8,7 @@
       :value="option.value"
       v-bind="$attrs"
     >
-      <!-- eslint-disable-next-line vue/no-v-html -->
-      <span v-if="!!option.html" v-html="option.html" />
-      <template v-else>
-        {{ option.text }}
-      </template>
+      {{ option.text }}
```
xvaara, line 11: Shouldn't this also have the option() slots logic?

packages/bootstrap-vue-next/src/components/BForm/BFormDatalist.vue
xvaara, line 6: Shouldn't this also have the option() slots logic?

packages/bootstrap-vue-next/src/components/BFormCheckbox/BFormCheckboxGroup.vue
```diff
@@ -9,19 +9,22 @@
     tabindex="-1"
   >
     <slot name="first" />
-    <BFormCheckbox v-for="item in normalizeOptions" :key="item.self" v-bind="item.props">
-      <!-- eslint-disable-next-line vue/no-v-html -->
-      <span v-if="!!item.html" v-html="item.html" />
-      <template v-else>
+    <BFormCheckbox v-for="(item, index) in normalizeOptions" :key="item.self" v-bind="item.props">
+      <slot
+        :name="slots[`option(${index})`] ? (`option(${index})` as 'option()') : 'option()'"
```
xvaara, line 14: I like the idea of the slots. Was wondering if we could have a slot for value also? option(${item.props.value}) so we could have `<template #option(foo)>`

xvaara (commented):
Mostly ok.

dwgray (commented):
LGTM in general. Agree with @xvaara on wanting the slots for the other similar components.

packages/bootstrap-vue-next/src/components/BFormCheckbox/BFormCheckboxGroup.vue
xvaara: ah, but then we have problem if the value is a number, does the index or the value take precedence...

Me myself, I don't see a case where I would use the index over the value in this. So I'd say use the value as option.foo and create a different for the index, ie option[1]... or something like that?

VividLemon: In all situations of any component, I am unsure if `value` is unique

VividLemon: @xvaara In some components `value` could also be a non-string representation. `<option>`'s `value` could be `{}` if desired

VividLemon: I toyed with the idea of using the components `id` but index seemed easier than ids

xvaara: > In all situations of any component, I am unsure if `value` is unique

Does it need to? It's a problem for the end user. I find if I need to figure which index the element is in more bothersome.  ie if a add an element to the array, I need to change the v-slot index value also... And if it's a dynamic list that I want to somehow change for some, it will be so hard. Or I'd need to have multiple v-if:s in there.

The more I think about this the more I think that this is wrong way of doing it. 

We should have a prop for the optionSlotField (or something like that) which defaults to 'value' and name the slot with that, or index. And pass the whole option to the slot:

```vue
        <slot
          :name="
            slots[`option(${option[prop.optionSlotField]})`]
              ? (`option(${option[prop.optionSlotField]})` as 'option()')
              : slots[`option(${index})`]
                ? (`option(${index})` as 'option()')
                : 'option()'
          "
          :option="option"
        >
           <option :value="option.value" :disabled="option.disabled">
            {{ option.text }}
          </option>
        </slot>
```
 
then we can have

```vue
<template #option={option}>
  <option :value="option.value" :disabled="option.disabled">
    <img :src="option.flag"> {{ option.name }} ({{ option.coutrycode }})
  </option>
</template>
<template #option(us)={option}>
  <option :value="option.value" :disabled="option.disabled" class="bg-danger">
    <img :src="//images/us-flag.svg"> USA! USA! USA!
  </option>
</template>
```

Limiting it to text and value kinda limits the functionality too much.

xvaara: > I toyed with the idea of using the components `id` but index seemed easier than ids

What does the components id have to do with the options v-for?

xvaara: On the other hand, do we need to have slots for other than basic? user can just have:
```vue
<template #option={option}>
  <template v-if="option.countrycode === 'us'>
    <img :src="/images/us-flag.svg"> USA! USA! USA!
  </template>
  <template v-else>
    <img :src="option.flag"> {{ option.name }} ({{ option.coutrycode }})
  </template>
</template>

```

Issue #2316, opened 2024-10-30
Margin issue with b-form-group in latest update


rademax:
### Describe the bug

After updating Bootstrap Vue Next to the latest version, I noticed that `b-form-group` elements no longer have the expected default margin (especially `margin-bottom).` This styling used to be applied to maintain proper spacing between form elements but seems to have been removed or altered in the latest release.

### Steps to Reproduce:
Update to the latest version of Bootstrap Vue Next.
Use b-form-group components in a Vue 3 project.
Check padding and margin on b-form-group components in DevTools.

**Expected Behavior:**
`b-form-group` elements should have consistent spacing (`margin-bottom`) applied by default, similar to previous versions.

**Current Behavior:**
`b-form-group` components lack `margin-bottom`, causing form elements to appear closely packed without spacing.

### Screenshots:
<img width="1632" alt="Screenshot at Oct 30 18-56-34" src="https://github.com/user-attachments/assets/14af922a-caf6-4fb8-996c-5c8839551240">

### Additional Context:
I’ve applied temporary CSS overrides to address this issue, but an official fix would help avoid maintaining custom CSS for each project.

### Reproduction

https://stackblitz.com/edit/github-amz8xq

### Used Package Manager

npm

### Validations

- [X] Have tested with the latest version. This is still alpha version and sometime things change rapidly.
- [X] Follow our [Code of Conduct](https://github.com/bootstrap-vue-next/bootstrap-vue-next/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guide](https://github.com/bootstrap-vue-next/bootstrap-vue-next/blob/main/CONTRIBUTING.md).
- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A, please open a GitHub Discussion instead.
- [X] The provided reproduction is a [minimal reproducible](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

VividLemon:
I kind of didn't know what I was doing with this refactor, someone else working on this issue would be appreciated as I don't quite know how to fix it

loicduong:
In Bootstrap 4, there is a class .form-group with the margin-bottom property. This class has been removed in Bootstrap 5, leading to issues like the one above.

https://getbootstrap.com/docs/5.3/migration/#forms-1

rademax:
@loicduong Ok, thanks.

[...]

PR #2322, opened 2024-11-01
fix(BFormGroup): fix missing margin bottom

loicduong:
# Describe the PR

fix https://github.com/bootstrap-vue-next/bootstrap-vue-next/issues/2316

## Small replication

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

### DIFF
--- a/packages/bootstrap-vue-next/src/components/BFormGroup/BFormGroup.vue
+++ b/packages/bootstrap-vue-next/src/components/BFormGroup/BFormGroup.vue
@@ -6,7 +6,7 @@
     :role="isFieldset ? null : 'group'"
     :aria-invalid="computedAriaInvalid"
     :aria-labelledby="isFieldset && isHorizontal ? labelId : null"
-    :class="[stateClass, {'was-validated': props.validated}]"
+    :class="['bv-form-group', stateClass, {'was-validated': props.validated}]"
   >
     <ContentTemplate.define>
       <BFormInvalidFeedback
--- a/packages/bootstrap-vue-next/src/components/BFormGroup/_form-group.scss
+++ b/packages/bootstrap-vue-next/src/components/BFormGroup/_form-group.scss
@@ -1,0 +1,3 @@
+.bv-form-group {
+  margin-bottom: 1rem;
+}
--- a/packages/bootstrap-vue-next/src/styles/styles.scss
+++ b/packages/bootstrap-vue-next/src/styles/styles.scss
@@ -8,6 +8,7 @@
 @import "../components/BPagination/pagination";
 @import "../components/BTable/table";
 @import "../components/BFormInput/form-input";
+@import "../components/BFormGroup/form-group";
 @import "../components/BFormFile/form-file";
 @import "../components/BFormSpinbutton/spinbutton";
 


xvaara:
I really dislike that we add classes and margins etc. to components. If the user wants a margin, they can add mb-3 to the components class. Do we need this by default?

@VividLemon should we define our guidelines about these things somewhere? I could write a little proposal and you could take a look at it... Or are we aiming at maximum compatibility with BV2?

loicduong:
@xvaara I think we can set up a default value, if we need to change it we can use class to override

VividLemon:
> I really dislike that we add classes and margins etc. to components. If the user wants a margin, they can add mb-3 to the components class. Do we need this by default?

I have no opinions. @xvaara You can manage this issue. If you want to remove it, put it in the migration guide

dwgray:
I'm with @xvaara on this one, I don't think we should be adding/maintaining extra CSS explicitly for BSV/boostrap4 compatibility _especially_ when it's related to something that BS5 has explicitly changed.

The way my head was going with this is that we could have a suggested configuration to pick up some of these things for folks that are looking for a smooth path migration from BSV - at minimum document in migration, but if there are more than one or two of these, we could just have a configuration that for people to grab. Note that I am not suggesting that we maintain a compatibility switch, just documentation.

I thought there was a smooth path to this - we basically just want to add the default class of "mb-1" to `BFormGroup`.
e.g. 

```ts
createApp(App)
  .use(createBootstrap({components: {BFormGroup: {class: 'mb-1'}}}))
  .use(router)
  .mount('#app')
```
But since the class attribute is a fall-through attribute this doesn't work because (1) typescript doesn't know about it and (2) the class from the default doesn't get merged. But if I explicitly set `class="mb-1"` on a `BFormGroup` the class gets merged correctly.

Is there a different way to do this? I think it would be generally useful to be able to set `class` defaults for components and we can for more complex components that have wrapper classes, etc. but not in this case. Unless I'm missing something?

xvaara:
@dwgray We don't want to add the class as a props, that causes so much headache. Maybe add wrapperClass to that component also? Or add it to everyclass? We could have extraAttrs for each component, bind it to the root (not clear on every component) then we could have more customising the html. But I'm not a fan of those. Just add the mb-3 to the component. Like bootstrap5 docs show input groups: `<div class="input-group mb-3">`.

Or this should at least have a class so that people can but the margin in their own style.css? Or let's say every our own custom component should have a unique class. That might be easiest way for this.

loicduong:
@xvaara What will our decision be? I also like the solution like Bootstrap 5, showing it in examples and adding it to the migration section.

dwgray:
> Or this should at least have a class so that people can but the margin in their own style.css? Or let's say every our own custom component should have a unique class. That might be easiest way for this.

I'd be good with any of the options @xvaara suggested, but I like having a unique class per component - that gives a great deal of flexibility without being opinionated. That combined with information in the migration guide should make it readily consumable by someone migrating from BSV.

xvaara:
I'm in favour of just adding the class for the components. if bootstrap is in basic form (ie modal) should we also be? Like class="form-group" ? do we need the bv-prefix?

VividLemon:
> do we need the bv-prefix?

All custom classes should have this prefix. Not necessarily by requirement, but to eliminate conflicts, it shows where the class came from, & is a convention started by bv

xvaara:
closing in favor of #2344

PR #2344, opened 2024-11-14
add some classes and ids for components

xvaara:
feat(BDropdown): id for split and menu elements
feat(BFromGroup): add class b-form-group
feat(BImg): add class b-img

# Describe the PR

@dwgray @VividLemon think of any other elements that would require a class or id for sub elements?

## Small replication

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

### DIFF
--- a/packages/bootstrap-vue-next/src/components/BDropdown/BDropdown.vue
+++ b/packages/bootstrap-vue-next/src/components/BDropdown/BDropdown.vue
@@ -26,6 +26,7 @@
     </BButton>
     <BButton
       v-if="props.split"
+      :id="computedId + '-split'"
       ref="button"
       :variant="props.variant"
       :size="props.size"
@@ -49,6 +50,7 @@
       <ul
         v-if="!props.lazy || modelValue"
         v-show="props.lazy || modelValue"
+        :id="computedId + '-menu'"
         ref="floating"
         :style="[floatingStyles, sizeStyles]"
         class="dropdown-menu overflow-auto"
--- a/packages/bootstrap-vue-next/src/components/BFormGroup/BFormGroup.vue
+++ b/packages/bootstrap-vue-next/src/components/BFormGroup/BFormGroup.vue
@@ -7,6 +7,7 @@
     :aria-invalid="computedAriaInvalid"
     :aria-labelledby="isFieldset && isHorizontal ? labelId : null"
     :class="[stateClass, {'was-validated': props.validated}]"
+    class="b-form-group"
   >
     <ContentTemplate.define>
       <BFormInvalidFeedback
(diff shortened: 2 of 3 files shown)


xvaara:
I added b-prefix since b-avatar etc was using it. Our own utility classes have bv-prefix.

For completeness sake, should we add the b-component-name class to every other component also? or rename our component classes without b-prefix to be inline with bootstraps names?

dwgray:
> I added b-prefix since b-avatar etc was using it. Our own utility classes have bv-prefix.
> 
> For completeness sake, should we add the b-component-name class to every other component also? or rename our component classes without b-prefix to be inline with bootstraps names?

Thinking this through a bit:

Bootstrap itself doesn't use a namespace prefix, so the pattern that many of our components have by the nature of implementing Bootstrap is that we have a class with a bare name that is basically the name of their 'component' on the main element. Just looking at a few, BAccordion = accordion, BAlert = alert, BBadge = badge, BBreadcrumb = breadcrumb, Button = btn, etc.
So I think we're just looking at the few that don't follow this pattern.
FormGroup is interesting because BS4 had a form-group class and then deprecated the class in BS5.
So I think it would be great to add a class to those that don't in order to fill in the pattern. I agree that it makes sense to use a prefix. And `b-*` seems fine for this.
I don't think that we should add an additional class to all of the components like the one cited above, since we'd not be getting any functionality. While I'm generally for consistency, adding the verbosity of an additional `b-*` class for every element that already has a `*` class seems like overkill. I'd be happy to stick a table in the docs somewhere to make it easy to find the class to add things to for each component (maybe we just add this to the `ComponentReference`?

Renaming our components to be in line with Bootstrap names—Are you suggesting that we rename Button to Btn? Are there other examples? I'm not sure I have a strong opinion on that.

VividLemon:
> Renaming our components to be in line with Bootstrap names—Are you suggesting that we rename Button to Btn? Are there other examples? I'm not sure I have a strong opinion on that.

BBtn can be an additional export. Similar to Binput

> And b-* seems fine for this

Prefer using this prefix for custom classes

> I don't think that we should add an additional class to all of the components like the one cited above, since we'd not be getting any functionality

I'd agree that useless classes like these should be avoided, even ones that are currently included now should probably be removed if they don't do anything

dwgray:
@VividLemon & @xvaara 

Okay - I think we have a path forward here. The plan is:
1. Add `b-*` to the relatively few components that don't have a bootstrap class already defined
1. Add aliases for those few components where the bootstrap class is not the same name as the component e.g. `btn` vs. `BButton` would prompt us to add a `BBtn` alias.
1. Document the class name for the 'primary' class for a component, since it may be either `b-name` or `name`, which won't be obvious to the developer without either some investigation or seeing it in the documentation.

I would be happy to handle 3.

Since the existing PR conforms to the above guidelines, I'd recommend that we let it through and I can catch the remaining exceptions in my parity pass (and since these are additive, getting every single one isn't urgent, while the ones that @xvaara covered here are cases that have come up - in at least one case, several times).

PR #2372, opened 2024-11-21
perf: use a composable rather than BTransition component, eliminating…

VividLemon:
… a useless fragment

# Describe the PR

## Small replication

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)


VividLemon:
> @VividLemon I'm against this... adding css again, css that doesn't even use bootstrap variables.

BTransition consumed components already use this CSS anyways. There is no real difference between before/after this change besides eliminating `BTransition` from the codebase

It could be a good idea to refactor to use actual bootstrap classes, if at all possible. But on the other hand to that, `Transition` is a useful abstraction for managing transitions

Issue #2378, opened 2024-11-22
BTable responsive ignore b-table-busy class


makroxyz:
### Describe the bug

BVN 0.26.2

When BTable has responsive property busy property class is ignored.

div table-responsive class  build around BTable **inherits all BTable props**

### Reproduction

https://stackblitz.com/edit/github-9yadcy?file=src%2Fcomponents%2FComp.vue,src%2FApp.vue

### Used Package Manager

npm

PR #2379, opened 2024-11-22
fixed BTable responsive

makroxyz:
See #2378


VividLemon:
https://github.com/bootstrap-vue-next/bootstrap-vue-next/pull/2380

Not quite the best way to fix this -- there's a prop that already duplicates that behavior. Thanks for the inspiration though.

Issue #2385, opened 2024-11-25
useShowHide `visible` attribute not behaving as documented


dwgray:
### Describe the bug

I'm working on updating the migration guide to reflect #1821 . Based on the description in `showhide-props.ts`, I would expect that a component with just `visible` would show on initial render (and with just `show` it would animate). `show` works as documented, but `visible` doesn't - see the example.

@xvaara Let me know I'm misunderstanding the intent of the documentation.

### Reproduction

https://stackblitz.com/edit/github-yymg1c?file=src%2FApp.vue

### Used Package Manager

pnpm

xvaara:
@dwgray I'm quite sure I missed that I broke that feature at some point. There were so many moving parts, forgot to test each of them after each change. It works when you toggle visible after the first render. I'll fix it.

Issue #2366, opened 2024-11-21
RFC: Unify BToast and BAlert (and maybe useShowHide Alert)


xvaara:
Hey, 

just noticed that BAlert is so close to BToast that meybe we should unify their functionality and props.

They could ne almost identical, only change the classes and few props maybe (like BTooltip & BPopover).

@VividLemon @dwgray Your thoughts on this matter?

dwgray:
> Hey,
> 
> just noticed that BAlert is so close to BToast that meybe we should unify their functionality and props.
> 
> They could ne almost identical, only change the classes and few props maybe (like BTooltip & BPopover).
> 
> @VividLemon @dwgray Your thoughts on this matter?

I'm always happy for things that reduce duplication. Taking a quick look, it seems like `BAlert` is a subset of `BToast` - I'd have a hard time consolidating them, but if you see a way to do that, I'd love to see it - I'm sure I'd learn a lot from it...

PR #2386, opened 2024-11-26
docs: migration guide for changes to show/hide related props

dwgray:
# Describe the PR

This is an attempt to encapsulate the prop changes in #1821. I will also get to the emit changes, but they were more straightforward so I'll tack them separately.

@xvaara In the process of gong through the code for #1821 I found a number of things that I either don't understand or are internally inconsistent (as well as the bug I filed yesterday #2385)

I'm going to list them here for now, but happy to pull them out as a bug or bugs if you agree they are issues:
- Is there a reason that transProps, noAnimation, and noFade aren't implemented on BAccordianItem and BAccordian? They're commented out and it feels like the should be implemented.
- BAlert and BCarousel both have `fade` rather than `noFade` - for carousel, this makes a lot of sense since this is referring to the transition within the component that defaults to `slide`, but for BAlert it seems like it might be better to have it follow the semantics of the other show/hide components. Certainly if we ever want the option to implement #2366, we'd want to line up the semantics now.
- `toggle-prevented` event is  missing on `BCollapse`, `BNavItemDropdown`
- `*-prevented` should have value arg of BvTriggerableEvent
- `BOffCanvas`, `BPopover`, and `BToast` don't declare events `toggle` or `toggle-prevented`, but they expose `toggle`, so I think they will emit those events?

## Small replication

N/A

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

### DIFF
--- a/apps/docs/src/components/ShowHideProps.vue
+++ b/apps/docs/src/components/ShowHideProps.vue
@@ -1,0 +1,95 @@
+<template>
+  <BTable :items="items" :fields="fields" hover small responsive bordered striped>
+    <template #cell(description)="d"> {{ d.item.description }} </template>
+    <template #cell(components)="d"
+      ><div v-for="component in d.item.components" :key="component">
+        <a :href="'#' + component.toLowerCase()">{{ component }}</a>
+      </div></template
+    >
+  </BTable>
+</template>
+
+<script setup lang="ts">
+import {computed} from 'vue'
+import {buildCommonProps, dropdownProps, kebabCase, pick, showHideProps} from '../utils'
+
+interface MigrationMap {
+  [key: string]: {oldProp: string; components: string[]}
+}
+
+interface ComponentMap {
+  [key: string]: string[]
+}
+
+const mappedComponents: ComponentMap = {
+  showHideProps: [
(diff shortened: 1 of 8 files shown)


VividLemon (approved):

xvaara:
I'm quite sure there are some mistakes there. The whole pr got out of hands and I didn't have the time or the energy to go through everything in it.

> I'm going to list them here for now, but happy to pull them out as a bug or bugs if you agree they are issues:
> 
> * Is there a reason that transProps, noAnimation, and noFade aren't implemented on BAccordianItem and BAccordian? They're commented out and it feels like the should be implemented.

On BAccordianItem it passes down the attrs to BCollapse. So they aren't defined as props, but work. Create A issue to keep track.

> * `toggle-prevented` event is  missing on `BCollapse`, `BNavItemDropdown`
> * `*-prevented` should have value arg of BvTriggerableEvent
> * `BOffCanvas`, `BPopover`, and `BToast` don't declare events `toggle` or `toggle-prevented`, but they expose `toggle`, so I think they will emit those events?

I'll create a pull request for these. Check out if I missed anything.

Issue #2418, opened 2024-12-05
Get Docs Error Clean & Type Check Docs Build


dwgray:
### Describe the bug

We keep introducing random type errors in the docs build since there is no build-time check. AFAIK we're just depending on annotations in VSCode.

I would ultimately like to get the docs build to fail if typescript emits errors just as we do with the main library build, but there are a number of non-trivial things to clean up before turning that on. So this is a tracking issue as I work through that process.

- [x] Add a `type-check` command to docs package, get it running in turbo
- [x] Figure out a work-around for the vite typescript errors
- [x] Fix errors in ComponentReference
- [x] Fix Errors in OppenCollective
- [x] Fix/suppress errors in examples
- [x] Fix errors in *.data file
- [x] Add `build-only` to docs and include type check in the default build

### Reproduction

run `vue-tsc --noEmit -p tsconfig.app.json` on docs

### Used Package Manager

pnpm

PR #2419, opened 2024-12-05
docs: Enable optional type-checking and clean up several issues

dwgray:
# Describe the PR

See #2418 

This PR sets up a type-check command for docs and cleans up the errors in component reference data and samples.

I'll chip away at the other errors as time permits

## Small replication

N/A

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

### DIFF
--- a/apps/docs/package.json
+++ b/apps/docs/package.json
@@ -6,6 +6,7 @@
   "scripts": {
     "dev": "vitepress dev --port 8000",
     "build": "cross-env NODE_OPTIONS=--max-old-space-size=4096 vitepress build",
+    "type-check": "vue-tsc --noEmit -p tsconfig.app.json",
     "preview": "vitepress preview",
     "lint": "eslint --fix src",
     "test:lint": "eslint src",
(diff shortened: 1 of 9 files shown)


apps/docs/package.json
```diff
@@ -6,6 +6,7 @@
   "scripts": {
     "dev": "vitepress dev --port 8000",
     "build": "cross-env NODE_OPTIONS=--max-old-space-size=4096 vitepress build",
```
VividLemon, line 8: Add build-only variant

package.json
```diff
@@ -32,6 +32,7 @@
     "dev": "turbo run dev --parallel",
     "build": "turbo run build",
     "lint": "turbo run lint",
+    "type-check": "turbo run type-check",
```
VividLemon, line 35: Add a build-only variant

VividLemon (changes requested):

apps/docs/package.json
dwgray: I think I have this set up the way you're asking. We can now do `pnpm build-only` from the root and it will do builds without type-check. Once I have `type-check` running in `docs` without errors, `build` will become type-check && build-only.

package.json
dwgray: See above.

VividLemon (approved):

PR #2425, opened 2024-12-07
Changes to public composables

xvaara:
# Describe the PR

Lot's of changes to composables.

## Small replication

## PR checklist

- [x] Feature - `feat(...)`
- [ ] Other (please describe)

## Summary by CodeRabbit

- **New Features**
  - Introduced `useToggle` composable with documentation and Table of Contents integration.
  - Added popover orchestration component and enhanced popover/tooltip controller APIs with promise-based control.
  - Enhanced modal and toast controllers with unified `create` methods and richer lifecycle management.
  - Expanded BAlert, BModal, BToast, and BPopover components with new props for customization and accessibility.
  - Added new slot props and standardized event typings for improved slot usage and event handling.
- **Bug Fixes**
  - Updated directives and components to use `body` prop consistently, replacing deprecated `content`.
  - Improved accessibility and ARIA attribute handling for toggle triggers.
- **Documentation**
  - Added and updated documentation for new composables, controllers, and migration guidance.
  - Updated usage examples to reflect new APIs and prop changes.
- **Refactor**
  - Unified and refactored orchestrator plugins and component APIs for consistency, extensibility, and promise-based workflows.
  - Improved internal typing and slot/event definitions for better type safety and developer experience.
  - Simplified event handling and reactive state management in components.
- **Tests**
  - Updated and expanded test suites to align with new component behaviors, prop changes, and plugin integration.
- **Chores**
  - Added new launch configuration for unit testing.
  - Updated TypeScript and plugin configurations for broader test coverage.


xvaara:
@VividLemon @dwgray any ideas why the docs build hangs forever?

packages/bootstrap-vue-next/src/plugins/toastController/index.ts
```diff
@@ -19,73 +19,124 @@ export const toastPlugin: Plugin = {
...
+      }
+    }
+
     /**
      * @returns {ControllerKey} If `id` is passed to props, it will use that id, otherwise, a symbol will be created that corresponds to its unique id.
      * You can pass this id to the hide function to force a Toast to hide
      */
-    const show = (obj: ToastOrchestratorShowParam = {}): ControllerKey => {
-      const resolvedProps = toRef(obj.props)
+    const show = (obj: ToastOrchestratorShowParam = {}): Promise<boolean | null> => {
```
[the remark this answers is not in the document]
VividLemon: I think a util method would look like

```ts
type Destructurable<T extends Record<string, any>> = T & Array<T[keyof T]>;

const makeDestructurable = <T extends Record<string, any>>(obj: T): Destructurable<T> => {
  const entries = Object.values(obj);
  return Object.assign(entries, obj) as Destructurable<T>;
}
```

In regards to the thenable, `then: async () => { await promise; return { id } `, it doesn't quite make sense to return the id here, as the toast (or other structure) would be closed. The id makes no use in this context. Regardless, it could be useful for the modals and whatnot.

packages/bootstrap-vue-next/src/directives/toggle.spec.ts
```diff
@@ -3,9 +3,10 @@ import {afterEach, describe, expect, it, vi} from 'vitest'
 import {nextTick} from 'vue'
 import {vBToggle as VBToggle} from './BToggle'
 import {asyncTimeout} from '../../tests/utils'
+import BCollapse from '../components/BCollapse/BCollapse.vue'
+import {createBootstrap} from '../plugins'
```
VividLemon, line 7: 🥱 A couple of tests need this. I haven't ran through everything but could be useful to fix up those cases as well where the terminal bloats up with "failed to find plugin". But it's never been a bother so I haven't done it. Still though, would be a useful addition, if you would.

VividLemon (changes requested):

VividLemon:
> @VividLemon @dwgray any ideas why the docs build hangs forever?

We'll figure it out after I'm done reviewing. Still not done yet, but I've submitted the first set of changes I'd like to see

VividLemon (commented):
I wonder if it could be preferable to combine useModal into useModalController. But maybe not, they serve different purposes 💭

xvaara:
> I wonder if it could be preferable to combine useModal into useModalController. But maybe not, they serve different purposes 💭 plus the naming of methods would seem like it would cause confusion

I was thinking of creating a useToggle that can control all the useShowHide components, that would make useModal kinda reduntant. Thoughts?

The documentation build was due to setInterval in script setup. Which is weird because I didn't add them 😄 
I'm a little under the weather at the moment, so might not continue this right away.

packages/bootstrap-vue-next/src/plugins/toastController/index.ts
```diff
@@ -19,73 +19,124 @@ export const toastPlugin: Plugin = {
...
+      const {component, slots} = toValue(obj)
+      if (component) {
+        if (isRef(obj)) obj.value.component = markRaw(component)
+        else if (typeof obj === 'object') obj.component = markRaw(component)
```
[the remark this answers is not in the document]
xvaara: Can't get typescript happy about any options using ||, and if I ask copilot to fix it, it just reverts to these if statements. Could you share an example?

packages/bootstrap-vue-next/src/plugins/toastController/index.ts
```diff
@@ -19,73 +19,124 @@ export const toastPlugin: Plugin = {
...
+      }
+      if (slots) {
+        if (isRef(obj)) obj.value.slots = markRaw(slots)
+        else if (typeof obj === 'object') obj.slots = markRaw(slots)
+      }
```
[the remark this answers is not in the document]
xvaara: same as above

packages/bootstrap-vue-next/src/utils/floatingUi.ts
```diff
@@ -23,33 +23,39 @@ export const resolveActiveStatus = (values: DirectiveBinding['value']): boolean
...
     return {
-      content: values,
+      body: values,
     }
   }
+
+  // TODO: deprication remove warning in 2025-02
+  if (values?.content)
+    // eslint-disable-next-line no-console
+    console.warn('v-b-popover/v-b-tooltip: `content` is deprecated, use `body` instead')
```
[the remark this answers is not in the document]
xvaara: What part of the documentation should it go to? Migration yes, but this might be used a lot by our current users, and directives don't have that good of a type safety so there is no deprecation warning there. It's in the changelog, but people might miss it and I don't want to leave the fix in forever.

packages/bootstrap-vue-next/src/plugins/toastController/index.ts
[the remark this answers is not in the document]
xvaara: I was thinking the same, but kinda just went with this. I'll implement the suggestion.

xvaara:
@VividLemon I'm thinking of removing the support for "component" their slots don't work at the moment, we need to support all the slots for that to work, now we just render what ever is in modal.slots for example. I couldn't find a way to list components available slots, other that just putting them to a variable, so I did the current version. 

Only thing I think it's useful for is to have a custom default for the bmodal and you don't want to list them in the function args.

packages/bootstrap-vue-next/src/plugins/toastController/index.ts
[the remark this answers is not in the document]
VividLemon: my mistake

packages/bootstrap-vue-next/src/utils/floatingUi.ts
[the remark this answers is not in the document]
VividLemon: Migration guide mostly. Ideally the popover documentation would get written, of both component usage and directive.

VividLemon (changes requested):
How do you share many same props between two instances of `show`?

```ts
const mySharedProps = computed(() => ({ foo: bar.value, buzz: aef.value }))
show({ ...mySharedProps.value, a: 1 })
show({ ...mySharedProps.value, b: 2 })
```
This would required `show(computed(() => ({ ... `
Is this valid syntax? Ideally both `show({ title: ref() })` and `show(computed(() => ({ title: ref() })))` would be valid

xvaara:
> How do you share many same props between two instances of `show`?
> 
> ```ts
> const mySharedProps = computed(() => ({ foo: bar.value, buzz: aef.value }))
> show({ ...mySharedProps.value, a: 1 })
> show({ ...mySharedProps.value, b: 2 })
> ```
> 
> This would required `show(computed(() => ({ ... ` Is this valid syntax? Ideally both `show({ title: ref() })` and `show(computed(() => ({ title: ref() })))` would be valid

They should all work, I haven't tested all. But the { var: ref() } part is only typed for title and body at the moment and it goest through toValue inside the watch. Maybe I'll loop all the props to be able to be refs... I'll test next week

VividLemon:
> > How do you share many same props between two instances of `show`?
> > ```ts
> > const mySharedProps = computed(() => ({ foo: bar.value, buzz: aef.value }))
> > show({ ...mySharedProps.value, a: 1 })
> > show({ ...mySharedProps.value, b: 2 })
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This would required `show(computed(() => ({ ... ` Is this valid syntax? Ideally both `show({ title: ref() })` and `show(computed(() => ({ title: ref() })))` would be valid
> 
> They should all work, I haven't tested all. But the { var: ref() } part is only typed for title and body at the moment and it goest through toValue inside the watch. Maybe I'll loop all the props to be able to be refs... I'll test next week

```ts
show(ref({ foo: 'bar' })
show({foo: ref('bar') })
show(() => ({ foo: props.bar }))
show(computed(() => ({ ...computedValue.value }))
```
I think for simplicity just the root should be `MaybeRef` `show({foo: ref('bar') })` Might be too much overhead.

How this would work with values to be updated, I'm not sure yet. 

`MaybeRefOrGetter` has the limitation of it's implicitly readonly. So we should probably do `MaybeRef` with `toRef` so if it's not a ref we can mutate the modelValue, and when it is, we can still update the modelValue

xvaara:
MaybeRefOrGetter<T = any> = MaybeRef<T> | ComputedRef<T> | (() => T)

it's not readonly

VividLemon:
> ```ts
> export type MaybeRefOrGetter<T = any> = MaybeRef<T> | ComputedRef<T> | (() => T);
> ```
> 
> it's not readonly

ComputedRefs will be. Which means if someone passes a computed, it breaks. MaybeRef is better in this case

packages/bootstrap-vue-next/src/components/BPopover/BPopover.vue
```diff
@@ -79,31 +77,42 @@ import {
...
+import {useShowHide} from '../../composables/useShowHide'
 
 defineOptions({
   inheritAttrs: false,
 })
 
+const attrs = useAttrs()
+
+// TODO: deprication remove warning in 2025-02
+if (attrs.content)
```
VividLemon, line 102: This

packages/bootstrap-vue-next/src/components/BPopover/BPopoverOrchestrator.vue
```diff
@@ -1,50 +1,34 @@
+<!-- eslint-disable vue/no-unused-vars -->
 <template>
-  <PopoverList />
-  <TooltipList />
+  <component
+    :is="_component ?? BPopover"
+    v-for="[key, {component: _component, slots, stop, ...val}] in tools.popovers?.value.entries()"
```
VividLemon, line 5: I'm not sure how destructuring behaves when popovers is undefined or null. The ts here is weird, because I'm pretty sure the behavior is js throws an error

packages/bootstrap-vue-next/src/components/BPopover/BPopoverOrchestrator.vue
```diff
@@ -1,50 +1,34 @@
...
+    :key="key"
+    v-bind="val"
+  >
+    <template v-for="(comp, slot) in slots" #[slot]="scope" :key="slot">
+      <component :is="comp" v-bind="scope" />
+    </template>
+  </component>
+  <component
+    :is="_component ?? BTooltip"
+    v-for="[key, {component: _component, slots, stop, ...val}] in tools.tooltips?.value.entries()"
```
VividLemon, line 15: same as above

packages/bootstrap-vue-next/src/components/BToast/BToastOrchestrator.vue
```diff
@@ -6,20 +6,30 @@
...
           <span
-            v-for="toast in tools.toasts?.value.filter((el) => el.props.pos === key)"
-            :key="toast.props._self"
+            v-for="{
+              _self,
+              slots,
+              promise,
+              component: _component,
+              ...val
+            } in tools.toasts?.value.filter((el) => el.position === key)"
```
VividLemon, line 19: above

packages/bootstrap-vue-next/src/directives/BToggle/index.ts
```diff
@@ -33,84 +33,84 @@ const getTargets = (
...
-      return
-    }
-    const target = document.getElementById(targetId)
+  targets.forEach(async (targetId) => {
+    let count = 0
+    while (count < 5) {
+      const showHide = showHideMap?.[targetId]
+      if (!showHide) {
+        count++
+        await new Promise((resolve) => setTimeout(resolve, 100))
```
VividLemon, line 68: async/await in Array.forEach doesn't run as typically expected. You may want to double check this behavior. (I think forEach doesn't await the callback)

packages/bootstrap-vue-next/src/plugins/modalController/index.ts
```diff
@@ -28,63 +28,86 @@ export const modalControllerPlugin: Plugin = {
...
+            promise,
+          })
+        },
+        {
+          immediate: true,
+          deep: true,
+        }
+      )
+
+      return promise.value
```
VividLemon, line 96: return { then() { return promise.value }, id: computedId.value }

Something like this

packages/bootstrap-vue-next/src/plugins/modalController/index.ts
```diff
@@ -28,63 +28,86 @@ export const modalControllerPlugin: Plugin = {
...
+          }
+          modals.value.set(_self, {
+            ...v,
+            ...(v.modelValue === undefined && {modelValue: true}),
+            // 'component': !v.component ? undefined : markRaw(v.component),
+
+            'onUpdate:modelValue': (val: boolean) => {
+              newValue['onUpdate:modelValue']?.(val)
+              const {modelValue} = toValue(obj)
+              if (isRef(obj) && !isRef(modelValue)) obj.value.modelValue = val
```
VividLemon, line 77: I think you could do toRef here and just always use it as a ref

packages/bootstrap-vue-next/src/plugins/modalController/index.ts
```diff
@@ -28,63 +28,86 @@ export const modalControllerPlugin: Plugin = {
...
+            v[key as keyof ModalOrchestratorShowParam] = toValue(
+              newValue[key as keyof ModalOrchestratorShowParam]
+            )
+          }
+          modals.value.set(_self, {
+            ...v,
+            ...(v.modelValue === undefined && {modelValue: true}),
+            // 'component': !v.component ? undefined : markRaw(v.component),
+
+            'onUpdate:modelValue': (val: boolean) => {
```
VividLemon, line 74: If you use toref and assume its always a ref, I don't know how useful this is

packages/bootstrap-vue-next/src/plugins/popoverController/index.ts
VividLemon, line 1: The changes of modalController are likely to be reflected here

packages/bootstrap-vue-next/src/plugins/toastController/index.ts
VividLemon, line 1: The above changes discussed before

VividLemon (changes requested):

packages/bootstrap-vue-next/src/directives/BToggle/index.ts
xvaara: I don't understand what you're meaning. Are you saying promises don't work inside forEach in some way?

Or are you meaning that the forEach returns immediately? if so, it doesn't matter, all the logic is inside the function.

VividLemon: https://play.vuejs.org/#eNp9Uk1v2zAM/SuELlUAw8nW9RKkBbYhA7rDVmy9VT3IDpOo1YehjySF4f9eyo7TFCh6kvj4SL1HqmXfm6bcJWRztgi1V02EgDE1N8Iq0zgfofW47mDtnYELIl4IWzsbIkjv4RoevhRfi8tHYUe4IpAqeCvn8PBYQJWPbjKmfbJE4BO4voFWWMhtyrXzS1lvuQwvtgaO+i0NMJ3CrTG4UjKifoEmhS0GkFoDajRoIwURnK1x4FflTuqEZVVmam424HIvVQSLe7gjLyog5x7D+UuQrd8rgy5FPkgkBo8+4aSAq9lsdmyV7Ryl/VJWnQQpC3GLEKRBiNQH+BW1JOOrcKwYxcl34nK/TliaDafbYjosglZAQUTTaLJOEUDbVh0RF9MzlBUsBnpjrTblU3CWNtn7Eax2plEa/d8mKhq+YPPRqWAk1+1/91j2V4x4vcX6+QP8KRwyJtgdjQT9DgU75aL0G4xDevn/Dx7ofkoat0qa2J8k/2FwOmWNA+1HsiuSfcbr1d72/1HZzX1YHiLaMJrKQvsx9nzB6Jf+/MT6m9zL8ltfR9Nn3SsGY/0B

They resolve at the same time. They are concurrent (forEach does not await the promise, it is `void`)

VividLemon:
The reason for using `MaybeRef` or `MaybeRefOrGetter` is not necessarily because a Ref passed to `MaybeRefOrGetter` is not readonly, but rather a convention to start to denote that if we are expecting `MaybeRefOrGetter`, we expect to mutate the value. If it is `MaybeRefOrGetter` we expect to not mutate the value, and should be treated as readonly

(If someone passed a computed, or a getter to the function it would break... A person passing `readonly(ref(())` would have an additional warning, but that's beyond our control (One could check `isReadonly()`, but this is overhead, and the user would experience a warning from vue, so it's a non-issue)

packages/bootstrap-vue-next/src/directives/BToggle/index.ts
xvaara: Yes, I know. That's what I want it to do. Did you read what it actually does? It loops throug the targets, if it doesn't find one it loops again 4 times waiting 100ms between them. Since sometimes directive runs before the child component is registered to the showHideMap.

VividLemon: Mb

asika32764:
I got 2 idea about merging these 2 composable.

One is keep the useModalController, and make everyone use controller object:

```ts
const modalController = useModalController();

modalController.create();
modalController.hide();
modalController.hideAll();

modalController.current(); // To replace `const { modal } =  useModal()`
```

Another is keep `useModal()`, and make people destruct functions:

```ts
const { create, hide, hideAll } = useModal();
```

That can make the code more consistency.

xvaara:
@asika32764 haha, damn you, you are right. I don't like the fact that we create with show function. 

useModal is more of a "toggleModal" composable anyway. and we now have useToggle composable that works on every component that can be toggled. I'd remove useModal, since it's not needed.

@VividLemon any objections? Since we are breaking the api anyway with this...

VividLemon:
I agree. Although a createandshow utility method would be useful. I often find myself using `if(await confirm()) deleteEntity()` style

xvaara:
also could we combine the orchestrators? now we need to add
  <BToastOrchestrator />
  <BModalOrchestrator />
  <BPopoverOrchestrator />
  
  for every app? Maybe
<BHelper :modal="false" \> if you don't want to load modals?

Maybe we should add BApp as a root component like vuetify has VApp as root, that creates the plugins, injections, defaults, orchestrators. That way user could have different defaults for different layouts for example.

xvaara:
I have a lot of promise test in my local code, I'll try to clean them up and move this pr forward tomorrow.

packages/bootstrap-vue-next/src/plugins/modalController/index.ts
xvaara: so how does the user then trigger stuff on change? pass a ref in and watch it? I thought the point of these hacks was to try to be like a component interface?

xvaara:
> @VividLemon I'm thinking of removing the support for "component" their slots don't work at the moment, we need to support all the slots for that to work, now we just render what ever is in modal.slots for example. I couldn't find a way to list components available slots, other that just putting them to a variable, so I did the current version.
> 
> Only thing I think it's useful for is to have a custom default for the bmodal and you don't want to list them in the function args.
> 
> -- Added later:
> 
> I figured why the slots of `component` wasn't rendering. In orchestrator if we have -tag inside the component it overrides all the default slots. Se we need to loop in extra template-tag, double the logic and add v-if="_component" to the first without template tags inside and another one with BModal and the template tags inside. Or do a render function, which you were against.

@VividLemon could you comment on this? Do you want to keep the component functionality in these? If you want to keep it, I'd suggest we convert to render function so we don't need to double the functionality.

xvaara:
a little demo and test for the composables using the controller: https://stackblitz.com/edit/a8wsb4ap?file=src%2FApp.vue

xvaara:
@VividLemon @asika32764 we remove the modal and toast after it is hidden and display toast automatically with 5s countdown. Should there be a parameter to "keep-alive" the component if one want's to do something with it? Or just create a new one. Now if you do 
```vue
const modal = create({title: 'foo'})
// something
modal.hide() // or close the modal

modal.set({title: 'bar'})
modal.show() // what should these two do? create a new modal with the same key/id? or error? Now they just return
```

Not really that common case, since why wouldn't you just make a new one.

VividLemon:
Are you saying that the way youve built it is that when you call hide, the modal will eventually be fully disposed of 'eventually' and you cant just show it again?

xvaara:
it is fully disposed at the moment. When someone does `confirm({body: 'Are you sure'}).then(i =>{ i && dataService.delete(id)})` we don't want the modal to stay in the registry. But someone might do something more complex... I'm just looking what your opinions are about how we should handle toast and modal here?

dwgray:
> it is fully disposed at the moment. When someone does `confirm({body: 'Are you sure'}).then(i =>{ i && dataService.delete(id)})` we don't want the modal to stay in the registry. But someone might do something more complex... I'm just looking what your opinions are about how we should handle toast and modal here?
> 
> We could just leave them. They now have `onScopeDispose(() => remove(_self), true)` so they are removed when the current scope is disposed. Maybe add removeAll to go with hideAll

@xvaara I thought I saw a question on this come through from you this morning, but I can't dig it up now.

I strongly agree that the default behavior should be to fully dispose of/deregister modals and other pop-up objects when they disappear. I believe that is the most likely use case (they are shown infrequently or possibly only once, or they are light enough weight that reconstructing them is no big deal). The issue with leaving them registered is that they act like a memory leak - so this might be a wrong knee-jerk from too much managing memory in C code and tracking down obscure garbage collection issues in C# code, but that's where I land.

That said, the ability to have some keep-alive flag for the remaining cases seems important. If I have a SPA what has a limited number of very complex modals that the user is frequently accessing, I'd want to keep them alive.

xvaara:
> 
> @xvaara I thought I saw a question on this come through from you this morning, but I can't dig it up now.
> 
 @dwgray It was on discord #dev channel.

> I strongly agree that the default behavior should be to fully dispose of/deregister modals and other pop-up objects when they disappear. I believe that is the most likely use case (they are shown infrequently or possibly only once, or they are light enough weight that reconstructing them is no big deal). The issue with leaving them registered is that they act like a memory leak - so this might be a wrong knee-jerk from too much managing memory in C code and tracking down obscure garbage collection issues in C# code, but that's where I land.

They are destoryed anyway when the current scope is destroyd, so not a big problem if not happening on the same component multiple times.

> 
> That said, the ability to have some keep-alive flag for the remaining cases seems important. If I have a SPA what has a limited number of very complex modals that the user is frequently accessing, I'd want to keep them alive.

to quote myself:
I was thinking we go the simple way: if it's created using `create`, we mark keep: true, if created with `show` (or `confirm`) then we mark keep: false.

If someone is creating anything complex enough, I'd hope they'd go for the component route. I'm mainly thinking of the "correct way" if we have `create` + `show` + `hide` + `delete` then `hide` shouldn't  also delete. Or we need to remove `hide` and `delete` and just put in `close`.

 but then again, why do we have `create` and `show` ... :smile: since it's so light to just create the modal/toast on the fly, I'm second guessing the fact that we give the user the option here. Since I don't see people using this instead of components to do anything complex. And you can control the modals/toasts with the useToggle that comes with this PR.

Also you can do: 
```js
const modal = mc.create({....})
modal.show()
modal.hide().delete()
// or
const { id } = mc.show({...})
mc.hide(id)
mc.delete(id)
```
I don't know if the api with the id is required if you can just use the object/promise returned from create/show. since there isn't that established way for these we need to decide what we want to to with these? I'd like not to offer multiple different ways of using these controller composables.
@VividLemon what is your opinion? do we need both of these apis for this?


I don't really know what the "correct way" would be here. So I guess I'm gonna implement all the ways to use these?

VividLemon:
>  @VividLemon what is your opinion? do we need both of these apis for this?

This is what I would prefer. 

This is the abstract behavior. It can be used to have a little bit of more control over the process. In what case one would want to use this, I don't know. I agree, any situation in where one would want to create a modal before showing is better served by using the component instead of this method
```ts
const modal = mc.create({})
modal.show()
modal.hide()
```

Most of the time people will tend to create/show modals on the fly. Say if they're using a `if(await confirm())` before saving to an api or something -- at least this is the most common use case _I_ have, and I'd argue that it's generally the most common. This is what should be promoted in the documentation, as it's probably best to use in terms of code cleanliness.
```ts
const modal = mc.show({}) // This method is a wrapper around mc.create() and mc.show(), it returns all from the abstract `create` method
modal.hide() // Does not destroy the modal. It is a pass through of mc.create() from above
```

> I don't know if the api with the id is required if you can just use the object/promise returned from create/show

This is where my stronger opinions lie. I think we completely scrap returning an id to plug into `hide()`. Instead I maintain the strong opinion that we implement the IDisposable interface from TypeScript. This dispose method is likely the following

```ts
async [Symbol.asyncDispose]() {
  await modal.hide() // Hide is a method that returns a promise that resolves when the animations for hiding have finished. If the modal has been hidden already, it resolves immediately (most common)
  modal.delete() // This method wipes the modal from the registry
}
```

Essentially, this asyncDispose method is equivalent to a `delete` method, or a cleanup method.

> They are destoryed anyway when the current scope is destroyd

When implemented, we will handle no auto cleanup of the modal from the registry. This automagic removal is something that should be removed. Instead it should be up to the user.

Why I say this is because we should prefer people to do 

```ts
await using modal = modal.show({})
```

The using feature will on scope destroy the modal. We provide no magic deletion or scheduled deletion. It is simply managed by the runtime. In cases where users are not using typescript or >5.2, they should use try/finally with `modal./* dispose() */` (or whatever the way to dispose is. 

I strongly encourage using this syntax. I believe it is a great addition to js (typescript only for now) and serves our situation perfectly. It simplifies the code as well. 

//
I could be convinced that there is no reason for `mc.create({....})` syntax. It just seems as a temporary intermediate to getting to `.show()`

packages/bootstrap-vue-next/src/types/ComponentOrchestratorTypes.ts
```diff
@@ -1,76 +1,206 @@
...
+   */
+  'pos'?: ContainerPosition
+  /**
+   * Position
+   * @default 'top-end'
+   */
+  'position'?: ContainerPosition
+  /**
+   * Sets whether or not the toast should be appended to the container
+   * @default undefined Implicitly defualts to the BToastOrchestrator's appendToast value
```
tbl0605, line 80: small typos in 2 different places: defualts => defaults

VividLemon:
Anything that you mark as "deprecated" should just be removed. If one is going to update to this version, then they should just have to update it once

xvaara:
> Anything that you mark as "deprecated" should just be removed. If one is going to update to this version, then they should just have to update it once

I'm mostly concerned if someone has `confirm().then((ok) => ok && deleteSystem())`
I think not everyone read the release notes, so this might help them.

xvaara:
@coderabbitai full review

VividLemon:
Once you resolve the conversations from the bot, I will look at this code and hopefully get it done for 0.29

packages/bootstrap-vue-next/src/plugins/toastController/index.ts
```diff
@@ -15,85 +33,193 @@ export const toastPlugin: Plugin = {
...
+          deep: true,
+          immediate: true,
+        }
+      )
+      onScopeDispose(() => {
+        const toast = toasts.value.find((el) => el._self === _self)
+        if (toast) {
+          toast.promise.value.destroy?.()
+        }
+      }, true)
```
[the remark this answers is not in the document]
xvaara: onScopeDispose takes two arguments, it's type in vue 3.5+ is:
function onScopeDispose(fn: () => void, failSilently?: boolean): void

packages/bootstrap-vue-next/src/plugins/modalController/index.ts
```diff
@@ -1,105 +1,217 @@
...
+          deep: true,
+        }
+      )
+      onScopeDispose(() => {
+        const modal = modals.value.get(_self)
+        if (modal) {
+          modal.promise.value.destroy?.()
+        }
+      }, true)
+      return promise.value
```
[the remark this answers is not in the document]
xvaara: @coderabbitai onScopeDispose takes two arguments, it's type in vue 3.5+ is:
function onScopeDispose(fn: () => void, failSilently?: boolean): void

packages/bootstrap-vue-next/src/plugins/popoverController/index.ts
```diff
@@ -1,134 +1,210 @@
...
+      if (slots) {
+        if (isRef(obj)) obj.value.slots = markRaw(slots)
+        else if (typeof obj === 'object') obj.slots = markRaw(slots)
+      }
       const resolvedProps = toRef(obj)
-      const _self = resolvedProps.value?.id || Symbol('Tooltip controller')
+      const _self = resolvedProps.value?.id || Symbol('Popover controller')
 
-      watch(
+      const promise = buildPromise(_self, popovers)
```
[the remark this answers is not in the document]
xvaara: PopoverOrchestratorParam does declate id field.

packages/bootstrap-vue-next/src/plugins/toastController/index.ts
```diff
@@ -15,85 +33,193 @@ export const toastPlugin: Plugin = {
...
+              nextTick(() => {
+                toast['onUpdate:modelValue']?.(false)
+              })
+            })
+          }
+          toasts.value.splice(
+            toasts.value.findIndex((el) => el._self === _id),
+            1
+          )
+        },
```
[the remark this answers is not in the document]
xvaara: splice isn't called else where that removes a toast.

VividLemon (approved):
@xvaara If you think this looks fine, please double check to review, then please squash merge with the correct conventional commits message with all of the breaking changes here. Other than that reservation, I think this looks good.

VividLemon:
Lmk when you're done

PR #2459, opened 2024-12-14
docs: Fix type errors in  OpenCollective components

dwgray:
# Describe the PR

General clean-up of types in the OpenCollective components, mostly to do with grouping resulting in unknown types and how to deal with the possibility of null members in the resulting groups.

The other is that for some reason, while [vitepress build time data loading](https://vitepress.dev/guide/data-loading#build-time-data-loading) is working, whatever magic they are using to lie to typescript to tell it that the `*.data.ts` file contains a named export called `data` isn't working. I wasn't able to find an existing but on vitepress about this. However, I was able to work around it by stubbing the `data` export. This fixes typescript, and doesn't break the data loading - would be delighted with a better solution if anyone has one.

```
Module '"../data/opencollective.data"' has no exported member 'data'. Did you mean to use 'import data from "../data/opencollective.data"' instead?
```

## Small replication

N/A

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

### DIFF
--- a/apps/docs/src/data/opencollective.data.ts
+++ b/apps/docs/src/data/opencollective.data.ts
@@ -4,11 +4,6 @@ const collectiveSlug = 'bootstrap-vue-next'
 const openCollectiveBaseURL = 'https://opencollective.com'
 const openCollectiveMembersFetchUrl = `${openCollectiveBaseURL}/${collectiveSlug}/members/all.json`
 
-// This is a stubbed export to satisfy typescript, for some reason
-//  the generated export from vitepress data isn't being recognized in this case
-//  but vitepress is generating the correct export at runtime
-export const data: CollectivePartialResponse = {members: []}
-
 export default {
   load: async (): Promise<CollectivePartialResponse> => {
     const response = await fetch(openCollectiveMembersFetchUrl)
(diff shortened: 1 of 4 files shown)


apps/docs/src/data/opencollective.data.ts
```diff
@@ -1,33 +1,16 @@
...
-  github: string | null
-  website: null | string
-}
+import {type CollectivePartialResponse} from '../types'
 
 const collectiveSlug = 'bootstrap-vue-next'
 const openCollectiveBaseURL = 'https://opencollective.com'
 const openCollectiveMembersFetchUrl = `${openCollectiveBaseURL}/${collectiveSlug}/members/all.json`
 
+// This is a stubbed export to satisfy typescript, for some reason
```
dwgray, line 7: Also see my comment on the PR - would be delighted if there is a better solution for this - but I can't seem to get typescript to recognise this without declaring it or to suppress the error.

VividLemon (approved):

apps/docs/src/data/opencollective.data.ts
VividLemon: suppress the error

dwgray: > suppress the error

Thanks. For some reason, I was blocking on the fact that the error didn't have an error number associated with it - but I used ts-expect-error for the line; that way, if the issue gets fixed downstream, this should start failing, and we can pull the suppression.

VividLemon (approved):

PR #2461, opened 2024-12-17
docs: use page header data to build "on this page" table of contents

dwgray:
# Describe the PR

Rather than manually adding [[toc]] to every page, I've turned on the option to add the page headers into the page data and then I can just put a vue component in the Layout.vue sidebar to generate the "On the page" table of contents.

My medium term goal for this is to move to using a vertical nav for this and adding in scrollspy to keep it in sync with the core page + expand/collapse like the BSV and bootstrap TOCs do.

I believe this step is a significant enough improvement on its own to get it in before I muck with scrollspy, etc.

- This removes a lot of boilerplate code from the *.md files
- It fixes the issue that I kludged up a while ago where the component reference item in the toc is in a separate <ul>, which looks fine visually but probably messes with accessibility.

## Small replication

N/A

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

### DIFF
--- a/apps/docs/.vitepress/config.mts
+++ b/apps/docs/.vitepress/config.mts
@@ -56,5 +56,6 @@ export default defineConfig({
       md.use(markdownItClass, {table: ['table', 'table-striped']})
       md.use(demoContainer, 'src')
     },
+    headers: true,
   },
 })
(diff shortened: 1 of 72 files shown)


VividLemon (commented):
Some places don't seem to delete their now unused component

dwgray:
> Some places don't seem to delete their now unused component

I did a global search for `ContentsSidebar` and `ComponentSidebar` and they no longer exist in this branch.

I think the thing that you're seeing in the diffs is that *.md files that I've already done the example refactor on, which includes making use of auto importing, didn't have an explicit reference to these components, so you won't see a delete in this PR.

VividLemon (approved):

PR #2467, opened 2024-12-19
feat(BTable): custom filter functions

kevinkosterr:
# Describe the PR
This PR aims to bring over the ability to pass a custom filter function to the `BTable` component as seen in `bootstrap-vue`. This gives developers more control over how exactly they want to filter their data inside tables.

## Small replication
Hard to properly convey in a video, but here is a small replication of what this could do. Code is displayed underneath the table.

https://github.com/user-attachments/assets/232fbd06-ed5b-4087-9f37-17081af03b2c

## PR checklist

- [x] Feature - `feat(...)`
- [ ] Other (please describe)

### DIFF
--- a/packages/bootstrap-vue-next/src/components/BTable/BTable.vue
+++ b/packages/bootstrap-vue-next/src/components/BTable/BTable.vue
@@ -661,7 +661,7 @@ const computedItems = computed<Items[]>(() => {
               return false
 
             if (props.filterFunction && typeof props.filterFunction === 'function') {
-              return props.filterFunction(item)
+              return props.filterFunction(item, props.filter)
             }
 
             const realVal = (): string => {
--- a/packages/bootstrap-vue-next/src/components/BTable/BTable.vue
+++ b/packages/bootstrap-vue-next/src/components/BTable/BTable.vue
@@ -242,6 +242,7 @@ const _props = withDefaults(
     noSortableIcon: false,
     perPage: Number.POSITIVE_INFINITY,
     filter: undefined,
+    filterFunction: undefined,
     mustSort: false,
     filterable: undefined,
     provider: undefined,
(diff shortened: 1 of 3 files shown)


packages/bootstrap-vue-next/src/components/BTable/BTable.vue
```diff
@@ -658,6 +659,11 @@ const computedItems = computed<Items[]>(() => {
               (!props.filterable?.includes(key) && !!props.filterable?.length)
             )
               return false
+
+            if (props.filterFunction && typeof props.filterFunction === 'function') {
+              return props.filterFunction(item)
```
VividLemon, line 664: The second param according to the bootstrap-vue docs is `the content of the filter prop (could be a string, RegExp, array, or object)`, so `(item, prop.filter)`

VividLemon (changes requested):

packages/bootstrap-vue-next/src/components/BTable/BTable.vue
VividLemon: Idrk why it exists this way

packages/bootstrap-vue-next/src/types/ComponentProps.ts
```diff
@@ -1092,10 +1092,10 @@ export interface BTableProps<Items> extends Omit<BTableLiteProps<Items>, 'tableC
   perPage?: Numberish
   currentPage?: Numberish
   filter?: string
+  filterFunction?: (item: Items) => boolean
```
VividLemon, line 1095: I think this may be `(item: Readonly<Item>, filter: string | undefined) => boolean`

VividLemon:
I'm also not quite sure how I feel about this prop in general. Why not do `filter?: string | ((item: Readonly<Item>) => boolean)`? Why need the extra prop?

kevinkosterr:
@VividLemon I'm not sure why we'd use the `filter` prop for this, personally it seems more readable and clear being its own prop. I feel it would add more unnecessary complexity to cram it in with the existing `filter` prop.

PR #2480, opened 2024-12-25
docs: use BNav for "on this page" contents

dwgray:
# Describe the PR

This is another incremental step towards getting the "On the Page" contents working like in the BSV documentation.
- I've converted the raw `<ul>` `<li>` styling to use <BNav> and <BNavItem> and update the CSS to work correctly.
- In order to allow for nested navs I've added a slot after the link in the <BNavItem>
- With this is place we should be able to start using ScrollSpy to keep the nav and the body in sync (but an initial attempt failed, so I'll need to dig into ScrollSpy a bit deeper before I get that working)

## Small replication

See the docs On this Page sidebar

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

### DIFF
--- a/packages/bootstrap-vue-next/src/components/BNav/BNavItem.vue
+++ b/packages/bootstrap-vue-next/src/components/BNav/BNavItem.vue
@@ -10,6 +10,7 @@
     >
       <slot />
     </BLink>
+    <slot name="after" />
   </li>
 </template>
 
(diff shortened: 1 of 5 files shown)


packages/bootstrap-vue-next/src/components/BNav/BNavItem.vue
```diff
@@ -23,6 +24,8 @@ import {useDefaults} from '../../composables/useDefaults'
 defineSlots<{
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   default?: (props: Record<string, never>) => any
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  after?: (props: Record<string, never>) => any
```
dwgray, line 28: Something like this is useful to do nested vertical navs like the ones in our "On this page" contents. I'm struggling with the name. Other thoughts were `body`, `after-link` and `postfix`. Happy to change it if anyone has an opinion...

VividLemon (approved):

PR #2496, opened 2025-01-06
docs: Fix typescript errors for showhide

dwgray:
# Describe the PR

I'm not at all sure that this is the cleanest way to fix this issue. See comments in code.

## Small replication

N/A

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

### DIFF
--- a/apps/docs/src/utils/showhide-props.ts
+++ b/apps/docs/src/utils/showhide-props.ts
@@ -48,4 +48,4 @@ export const showHideProps = {
     default: false,
     description: "When 'true', open without animation",
   },
-} as const satisfies Record<keyof BvnComponentProps['showHide'], PropertyReference>
+} as const satisfies Record<keyof BvnComponentProps['showHide'] | 'modelValue', PropertyReference>
--- a/packages/bootstrap-vue-next/src/types/BootstrapVueOptions.ts
+++ b/packages/bootstrap-vue-next/src/types/BootstrapVueOptions.ts
@@ -294,5 +294,5 @@ type UnmappedComponentProps<BFormSelectOption = any, BTableLite = any, BTable = 
 }
 
 export type BvnComponentProps = {
-  [K in ComponentType]: UnmappedComponentProps[K]
+  [K in ComponentType | 'showHide']: UnmappedComponentProps[K]
 }


apps/docs/src/utils/showhide-props.ts
```diff
@@ -48,4 +48,4 @@ export const showHideProps = {
     default: false,
     description: "When 'true', open without animation",
   },
-} as const satisfies Record<keyof BvnComponentProps['showHide'], PropertyReference>
+} as const satisfies Record<keyof BvnComponentProps['showHide'] | 'modelValue', PropertyReference>
```
dwgray, line 51: In order for `showhide-props.ts` to reference `ShowHideProps` it must be exported, which it hadn't been. The way the other ComponentProps are exported is because their associated component is mapped in  `ComponentType` and the props are defined in `UnmappedComponentTypes`.  But there isn't really a BShowHide component. A think the alternative might be to directly export ShowHideProps? @VividLemon or @xvaara any thoughts on this?

packages/bootstrap-vue-next/src/types/BootstrapVueOptions.ts
```diff
@@ -294,5 +294,5 @@ type UnmappedComponentProps<BFormSelectOption = any, BTableLite = any, BTable =
 }
 
 export type BvnComponentProps = {
-  [K in ComponentType]: UnmappedComponentProps[K]
+  [K in ComponentType | 'showHide']: UnmappedComponentProps[K]
```
dwgray, line 297: In order for showhide-props.ts to reference ShowHideProps it must be exported, which it hadn't been. The way the other ComponentProps are exported is because their associated component is mapped in ComponentType and the props are defined in UnmappedComponentTypes. But there isn't really a BShowHide component. A think the alternative might be to directly export ShowHideProps? @VividLemon or @xvaara any thoughts on this?

VividLemon: Why aren't those props being automatically inferred with `ComponentProp extends ShowHideProps { }` ? 

I don't quite understand why this is an issue

VividLemon (changes requested):

packages/bootstrap-vue-next/src/types/BootstrapVueOptions.ts
dwgray: Because `showHide` isn't a component so doesn't get included in ExportedComponents - if I understand it correctly, it's basically the API surface for the showHide composable to be consumed by components that expose this functionality (but @xvaara may be able to shed more light on this). I didn't think adding a stubbed component called showHide in the components directory was any cleaner than what I did, but that's another possible solution.

And, of course, I may be completely missing something here. The core issue is that I would like to get the docs typescript error clean so I can turn on typechecking and not keep shooting myself in the foot. The error I'm fixing with this PR is

```Property 'showHide' does not exist on type 'BvnComponentProps'.ts(2339)```

here https://github.com/dwgray/bootstrap-vue-next/blob/docs-table-overview/apps/docs/src/utils/showhide-props.ts#L51

VividLemon: Export showHide props.

dwgray: Done

VividLemon (approved):

PR #2500, opened 2025-01-09
docs: Fix typescript errors in search components

dwgray:
# Describe the PR

This is the last set of typescript errors in docs. Once this and the other PRs in flight are merged, I'll push a change to add type-check to the main build, which should prevent new errors from being checked in.

I went down a bunch of paths to try to suppress these errors, since they aren't ours. However, it appears that the typescript team is opposed to suppressing errors in the dependency tree (you can exclude top-level files, but not things that are included by a file that you're checking.  See https://github.com/microsoft/TypeScript/issues/40426 for info.

I've opened a feature request on VitePress to see if they have other ideas on how to include their search control in a custom theme https://github.com/vuejs/vitepress/issues/4476 - but I'd like to get type checking turned on sooner then they're likely to react.

The other possibility is to rework our theme to be an extended default theme so that we can use their Layout component. But that seems like it would be a mess, and one of the nice things about our doc site as it stands is that we're using BSVN for most of the styling which I'm not sure we could do if we went down this route.

## Small replication

pnpm type-check

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

### DIFF
--- a/apps/docs/.vitepress/theme/Layout.vue
+++ b/apps/docs/.vitepress/theme/Layout.vue
@@ -177,10 +177,9 @@ import MoonStarsFill from '~icons/bi/moon-stars-fill'
 import SunFill from '~icons/bi/sun-fill'
 import ChevronRight from '~icons/bi/chevron-right'
 import CircleHalf from '~icons/bi/circle-half'
-import {useData, useRoute, withBase} from 'vitepress'
+import {useData, useRoute, VPNavBarSearch, withBase} from 'vitepress'
 import {appInfoKey} from './keys'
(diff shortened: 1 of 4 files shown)


dwgray:
> @VividLemon It looks like they fixed this (amazing turnaround) - between now and the next release, do you have a preference between letting this go through or putting up a new PR against the VP nightly build?

@VividLemon  On further though, I think this method is clearly better. I did a bit of poking around to check for unexpected regressions and things look fine. I'm going to move #2501 to draft in case you think otherwise, will close one or the other of these once the other one goes through.

Issue #2510, opened 2025-01-14
The formatter property on BTable's field definition is incomplete


dwgray:
### Missing feature is from

bootstrap-vue

### Docs

https://bootstrap-vue.org/docs/components/table#field-definition-reference

### Describe the missing feature

In BSV, this property could take a function definition or `the name of a method in your component`, the latter isn't implemented. Not sure how important it is, but we should decide to either implemented it or document it as deprecated.

VividLemon:
Oh no, we aren't going to implement that ability. To the user there is no difference between that and passing it in as an argument to the function. We aren't going to go through the hassle of searching for instance properties on their components, they can just give what they want directly

dwgray:
> Oh no, we aren't going to implement that ability. To the user there is no difference between that and passing it in as an argument to the function. We aren't going to go through the hassle of searching for instance properties on their components, they can just give what they want directly

That's fine by me - I'll document that we're deprecating this in the migration guide in my next round on table docs and close this out with that PR.

PR #2526, opened 2025-01-25
docs(BTable): Document custom data rendering

dwgray:
# Describe the PR

- Port the docs from BSV for custom data rendering of BTables
- Update the examples
- Update migration guide
- Address #2510 in the migration guide

## Small replication

N/A

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

### DIFF
--- a/apps/docs/src/docs/components/table.md
+++ b/apps/docs/src/docs/components/table.md
@@ -345,7 +345,78 @@ information on the `busy` state.
 
 ## Custom data rendering
 
-To Be Completed
+Custom rendering for each data field in a row is possible using either
+[scoped slots](https://vuejs.org/guide/components/slots.html#scoped-slots) or a formatter callback
+function, or a combination of both.
+
+### Scoped field slots
+
+Scoped field slots give you greater control over how the record data appears. You can use scoped
+slots to provided custom rendering for a particular field. If you want to add an extra field which
+does not exist in the records, just add it to the [`fields`](#fields-column-definitions) array, and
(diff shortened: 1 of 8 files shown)


apps/docs/src/docs/components/table.md
```diff
@@ -345,7 +345,78 @@ information on the `busy` state.
...
+#### Displaying raw HTML
+
+By default `BTable` escapes HTML tags in items data and results of formatter functions, if you need
+to display raw HTML code in `BTable`, you should use `v-html` directive on an element in a in
+scoped field slot.
+
+<<< DEMO ./demo/TableRawHtml.vue
+
+::: danger WARNING
+Be cautious of using the <code>v-html</code> method to display user
```
dwgray, line 400: While we've deprecated our own use of `v-html`, I think this a case where the developer may still need to use `v-html` against vue components - we're taking data from a js table and rendering it. But I may be missing something - @xvaara or @VividLemon any thoughts on this?

VividLemon: The issue is not with downstream devs from using v-html. The only issue was _our_ wrappers around v-html. 

The issue that I had is a prop `html`'s explicitness. It's explicit in its name, yes, but it doesn't have the same eslint rules for warning against v-html. The dev would need (assuming they had eslint installed) to write <div v-html /> and do an eslint-disable. Thus being three steps of security against writing insecure code.

So the goal of that was to make any place where v-html in our code was used, and instead replace them with slots that one could achieve the same result.

VividLemon (approved):

PR #2556, opened 2025-02-14
feat(BModal): use css var for zindex, add helper vars and ontop class

### LINKED ISSUE #2402: Add option to set the default or starting z-index for components

Some 3rd party components have been known to interfere with bootstrap default z-indexes. So we might want to add an option to add a starting point for example BModal.

### Suggested solution

option to createBootstrap, or add a css var that can be added to our calculations with 
`z-index: calc(${ourZindex} + var(--bv-zindex-hoist, 0))`

### Alternative

_No response_

### Additional context

_No response_

xvaara:
closes #2402

# Describe the PR

## Small replication

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

### DIFF
--- a/packages/bootstrap-vue-next/src/components/BModal/BModal.vue
+++ b/packages/bootstrap-vue-next/src/components/BModal/BModal.vue
@@ -16,6 +16,7 @@
           {
             fade: !computedNoAnimation,
             show: isVisible,
+            ontop: (activeModalCount ?? 0) - 1 === (activePosition ?? 0),
           },
         ]"
         role="dialog"
@@ -114,6 +115,7 @@
           :class="{
             fade: !computedNoAnimation,
             show: backdropVisible || computedNoAnimation,
+            ontop: (activeModalCount ?? 0) - 1 === (activePosition ?? 0),
           }"
           @click="hide('backdrop')"
         />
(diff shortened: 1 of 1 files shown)


xvaara:
demo on how to use:

https://stackblitz.com/edit/github-etfudpj2-rwoeskjl?file=src%2FApp.vue

xvaara:
I was thinking about this. We need a few more css variables in the code. Maybe we should import them from bootstrap and add to the :root element in our scss. toast width in toast orchestrator is at least another one.

VividLemon:
> I was thinking about this. We need a few more css variables in the code. Maybe we should import them from bootstrap and add to the :root element in our scss. toast width in toast orchestrator is at least another one.

Sounds like a good addition

packages/bootstrap-vue-next/src/components/BModal/BModal.vue
```diff
@@ -395,7 +406,23 @@ watch(stackWithoutSelf, (newValue, oldValue) => {
...
 
-const defaultModalDialogZIndex = 1056
+onMounted(() => {
+  watch(
+    renderRef,
+    (v) => {
+      if (!v) return
+      nextTick(() => {
+        if (!element.value) return
+        defaultModalDialogZIndex.value = parseInt(
```
VividLemon, line 416:
```suggestion
        defaultModalDialogZIndex.value = Number.parseInt(
```

packages/bootstrap-vue-next/src/components/BModal/BModal.vue
```diff
@@ -16,6 +16,7 @@
           {
             fade: !computedNoAnimation,
             show: isVisible,
+            ontop: (activeModalCount ?? 0) - 1 === (activePosition ?? 0),
```
VividLemon, line 19: This could probably be a computed method considering it's usage multiple times

VividLemon (changes requested):

dwgray:
> > I was thinking about this. We need a few more css variables in the code. Maybe we should import them from bootstrap and add to the :root element in our scss. toast width in toast orchestrator is at least another one.
> 
> Sounds like a good addition

That brings up something that I was looking at recently - in our [getting started example](https://bootstrap-vue-next.github.io/bootstrap-vue-next/docs.html#installation-vue-js), we directly import the generated .css files for bootstrap and bsvn. In the equivalent BSV example, the scss is imported. If we want developers to change our scss variables they'd need to import the SCSS. But if this is purely about CSS variables and we are moving to customization that way, we wouldn't. If that's the case, maybe we should be more explicit about moving to CSS variables? A section in the migration guide?

PR #2557, opened 2025-02-14
fix(b-avatar): Add "!default" to $b-avatar-border-color

tobiasnitsche:
Add "!default" to $b-avatar-border-color, so its can be overwritten.


xvaara:
thanks

tobiasnitsche:
Thanks a lot too for the quick merge 🙏

Issue #2571, opened 2025-02-21
BDropdown button click event discrepancy with docs


Rocik:
### Clear and concise description of the problem

Dropdown docs say
![Image](https://github.com/user-attachments/assets/7225c54d-7e37-42d5-a91c-e469747936e1)

But looking at the component, only `click` is emitted, and it works as if it was actually the `split-click`.
https://github.com/bootstrap-vue-next/bootstrap-vue-next/blob/3ee3fd828bb4df50f5c3c3ad854e4604c919ea3c/packages/bootstrap-vue-next/src/components/BDropdown/BDropdown.vue#L150-L154

I would need the caret click (in non-split mode) to be emitted with event prop in order to stop its propagation.

### Suggested solution

Implement the events (`click`, `split-click`) mentioned in the docs.

### Alternative

Fix documentation to match current implementation and then add additional event emit or update `click` event to be always emitted in non-split mode.

### Additional context

I'm up to sending a PR when a single solution is accepted.

xvaara:
@dwgray you added that to the docs in pr #2243 
Where did it come from and how is it supposed to work?

Is `click` emited only when not split? and split-click only when clicking on the slip-button and click on the button  even when split?. 

I think it's there mainly to push the event back up to the parent if the the button is split (on the button that doesn't open the dropdown, since that you can bind to with `@open`). So I'd remove`split-click` from the docs and add clarification to the docs that `click` is emitted only when the button is split.

dwgray:
> split-click

Sorry—this PR was one of those where I was updating the docs based on the BSV docs and checking parity across BSV/BS5/BSVN, so I'm not sure where this came from. When I'm doing these batches, I'm not intentionally adding anything new. I rewound the code to that commit. The code doesn't implement it there, and BSV doesn't implement it either.

Happy to update the docs. @Rocik , would what @xvaara is suggesting solve your issue? If you listen to the `@show` event, you can suppress it whether or not the dropdown button is split. If so, I'm happy to go ahead and update the docs, or you can if you'd like @Rocik - happy to have more folks contributing.

Rocik:
`@show` would not help, `@toggle` could, but it sends `BvTriggerableEvent` which does not stop the native click event.

My issue is shown here: https://stackblitz.com/edit/github-j4zpcsys?file=src%2FApp.vue

I don't really mind any solution, so even adding stop and prevent here would be fine by me.
[...]

PR #2572, opened 2025-02-22
fix(BDdropdown)!: Replace click event on split button with split-click

dwgray:
# Describe the PR

Addresses #2571 

- replace the `click` event with `split-click` when the dropdown is in split mode. This prevent it from hiding the native click event and allows the user to access the native event to stop propagation
- replace the `click:prev` and `click:next` events on `BCarousel` with `prev-click` and `next-click` for consistency
- make this work for NavItemdropdown as well
- remove the documentation for `click`
- update the migration guide
- Add test cases to `TDropdown` and `TNav`
- fix up the main documentation for `BDropdown`

## Small replication

See #2571 

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

### DIFF
--- a/apps/docs/src/data/components/carousel.data.ts
+++ b/apps/docs/src/data/components/carousel.data.ts
@@ -151,7 +151,7 @@ export default {
           ],
         },
         {
-          event: 'click:prev',
+          event: 'prev-click',
           description: '',
           args: [
             {
@@ -162,7 +162,7 @@ export default {
           ],
         },
         {
-          event: 'click:next',
+          event: 'next-click',
           description: '',
           args: [
             {
--- a/apps/docs/src/docs/components/dropdown.md
+++ b/apps/docs/src/docs/components/dropdown.md
@@ -97,7 +97,7 @@ You can view the [Floating-ui docs](https://floating-ui.com/docs/middleware) to 
 
 ## Split button support
 
-Create a split dropdown button, where the left button provides a `click:split` event and link support, while the right-hand side is the dropdown menu toggle button.
+Create a split dropdown button, where the left button provides a `split-click` event and link support, while the right-hand side is the dropdown menu toggle button.
 
 <<< DEMO ./demo/DropdownSplitButton.vue#template{vue-html}
 
--- a/apps/docs/src/docs/components/dropdown.md
+++ b/apps/docs/src/docs/components/dropdown.md
@@ -97,7 +97,7 @@ You can view the [Floating-ui docs](https://floating-ui.com/docs/middleware) to 
 
 ## Split button support
 
-Create a split dropdown button, where the left button provides a `split-click` event and link support, while the right-hand side is the dropdown menu toggle button.
+Create a split dropdown button, where the left button provides a `click:split` event and link support, while the right-hand side is the dropdown menu toggle button.
 
 <<< DEMO ./demo/DropdownSplitButton.vue#template{vue-html}
 
--- a/apps/docs/src/docs/components/dropdown.md
+++ b/apps/docs/src/docs/components/dropdown.md
@@ -97,7 +97,7 @@ You can view the [Floating-ui docs](https://floating-ui.com/docs/middleware) to 
 
 ## Split button support
 
-Create a split dropdown button, where the left button provides standard `click` event and link support, while the right-hand side is the dropdown menu toggle button.
+Create a split dropdown button, where the left button provides a `split-click` event and link support, while the right-hand side is the dropdown menu toggle button.
 
 <<< DEMO ./demo/DropdownSplitButton.vue#template{vue-html}
 
--- a/apps/docs/src/docs/migration-guide.md
+++ b/apps/docs/src/docs/migration-guide.md
@@ -278,7 +278,7 @@ See [Show and Hide](#show-and-hide) shared properties.
 See the [v-html](#v-html) section for information on deprecation of the `html` prop.
 
 The `click` event that was emitted when clicking on the left or button side of a `split` dropdown
-has been replaced by a `click:split` which provides the native mouse event. This is because
+has been replaced by a `split-click` which provides the native mouse event. This is because
 naming the event 'click' was hiding the native `click` event so supressing the that event for
 parents that might have unexpected actions (such as a link navigating to a new page) was difficult.
 
--- a/apps/docs/src/docs/migration-guide.md
+++ b/apps/docs/src/docs/migration-guide.md
@@ -278,7 +278,7 @@ See [Show and Hide](#show-and-hide) shared properties.
 See the [v-html](#v-html) section for information on deprecation of the `html` prop.
 
 The `click` event that was emitted when clicking on the left or button side of a `split` dropdown
-has been replaced by a `split-click` which provides the native mouse event. This is because
+has been replaced by a `click:split` which provides the native mouse event. This is because
 naming the event 'click' was hiding the native `click` event so supressing the that event for
 parents that might have unexpected actions (such as a link navigating to a new page) was difficult.
 
--- a/apps/docs/src/docs/migration-guide.md
+++ b/apps/docs/src/docs/migration-guide.md
@@ -277,6 +277,11 @@ See [Show and Hide](#show-and-hide) shared properties.
 
 See the [v-html](#v-html) section for information on deprecation of the `html` prop.
 
+The `click` event that was emitted when clicking on the left or button side of a `split` dropdown
+has been replaced by a `split-click` which provides the native mouse event. This is because
+naming the event 'click' was hiding the native `click` event so supressing the that event for
+parents that might have unexpected actions (such as a link navigating to a new page) was difficult.
+
 <NotYetImplemented>`toggleAttrs`</NotYetImplemented>
 
 #### Dropdown sub-components
--- a/apps/docs/src/utils/dropdown-common.ts
+++ b/apps/docs/src/utils/dropdown-common.ts
@@ -191,7 +191,7 @@ export const dropdownEmits: ComponentReference['emits'] = [
     description: 'Emitted when the dropdown tried to open, but was prevented from doing so.',
   },
   {
-    event: 'click:split',
+    event: 'split-click',
     description: 'Emitted when split button is clicked in split mode',
     args: [
       {
--- a/apps/docs/src/utils/dropdown-common.ts
+++ b/apps/docs/src/utils/dropdown-common.ts
@@ -154,7 +154,7 @@ export const dropdownProps = {
 export const dropdownEmits: ComponentReference['emits'] = [
   {
     event: 'click',
-    description: 'Emitted when button is clicked',
+    description: 'Emitted when the main button is clicked only when in split mode.',
     args: [
       {
         arg: 'event',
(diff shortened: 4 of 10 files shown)


VividLemon:
> this prevents a dropdown embedded in a link from navigating to the link's page

I don't quite like this solution because I think (or feel) that stopping propagation and default may have unintended side effects (mostly because it tends to have done so in the past)

In the edge case that is dropdowns and links, I think the user should likely handle this by their @click.stop.prevent or whatever

VividLemon (changes requested):
> this prevents a dropdown embedded in a link from navigating to the link's page

I don't quite like this as stop.prevent has had unintended consequences in the past.

I think in the edge case that is combining links and dropdowns, the user should probably be responsible for doing .stop.prevent. But perhaps I'm wrong

dwgray:
> > this prevents a dropdown embedded in a link from navigating to the link's page
> 
> I don't quite like this as stop.prevent has had unintended consequences in the past.
> 
> I think in the edge case that is combining links and dropdowns, the user should probably be responsible for doing .stop.prevent. But perhaps I'm wrong

That fair - but the problem is that as far as I can tell, we're not exposing an event that the user can do the `.stop.prevent` on, which was @Rocik's original issue. Would you support exposing some kind of click event on both the main button and the split button as an alternative? Or is there an existing way for the consumer of `BDropdown` to stop the native click events from propagating without us exposing them explicitly?

xvaara:
For me also this doesn't seem like the right way forward. I don't like the split button in dropdown anyway. It should be just one button, and the user is responsible for the other buttons and wrapping them in input-group class. I understand that this is was this way in vue2 version, but should it be?

We could just bind the attrs to the the right button and leave it to the end user.

dwgray:
> For me also this doesn't seem like the right way forward. I don't like the split button in dropdown anyway. It should be just one button, and the user is responsible for the other buttons and wrapping them in BButtonGroup component. I understand that this is was this way in vue2 version, but should it be?
> 
> We could just bind the attrs to the the right button and leave it to the end user.
> 
> -- paste from the issue:
> 
> remove the click emit. Then all this will work. well if we have a split then the @click will be bind to the wrapper, but I don't think that is a problem. Then we'd need to maybe have the emits for the button presses, but not for @click.
> 
> In either case, we shouldn't try to emulate native emits. It only leads to problems like these. We should just bind the attrs to the "right" element.

Short of deprecating the split mode, I landed on just renaming the `click` event to `split-click` - this give the user access to the native `click` event for anything they want to do with it while leaving a clean path for migration for folks that are using this component the way (I think) it was intended.

After going through this I realize that I may not fully understand

> We could just bind the attrs to the the right button and leave it to the end user.

If that allows for a reasonable migration experience for folks that were binding to the `click` event in split mode, I'd be happy to give that a shot, but I don't see how that would work.

xvaara:
> After going through this I realize that I may not fully understand
> 
> > We could just bind the attrs to the the right button and leave it to the end user.
> 
> If that allows for a reasonable migration experience for folks that were binding to the `click` event in split mode, I'd be happy to give that a shot, but I don't see how that would work.

the "native click" is in $attrs and onClick, so we could check for that and bind it to the "correct button". But I'm not sure if that is wise, wanted or even needed.

VividLemon:
Split buttons are native bootstrap

I'd prefer `click:split` as the name, but I don't know if this is common in the lib.

xvaara:
All the other emits are with `-` instead of `:` so let's not change that.

And I don't know what you mean by native? It's just another button in the same .btn-group and one class. What I mean is that this does basically the same as split: https://bootstrap-vue-next.github.io/bootstrap-vue-next/docs/components/button-group.html#dropdown-menu-support

we already detect when we are inside BButtonGroup, so we'd just need to add a prop to make the dropdown to render as currently in split.

Or maybe a split slot... So many better options than the current one.

xvaara (dismissed):
Seems fine for me. Just fix the test error 😁

dwgray:
> Split buttons are native bootstrap
> 
> I'd prefer `click:split` as the name, but I don't know if this is common in the lib.

We have a mix of events that use dash vs. colon. I did a quick search to see what we're currently doing. I think the only comparables are `click:prev` and `click:next` in `BCarousel` and `page-click` in `BPagination`.

My plan is to change this on to `click:split` in this PR (and fix the tests) and file a bug against myself to change the `page-click`. I'll append my research here, FYI.

**Events with colon**
```
5 results - 4 files

apps\docs\src\data\components\button.data.ts:
  89          {
  90:           event: 'update:pressed',
  91            description: 'Emitted when the `pressed` prop is changed',

apps\docs\src\data\components\carousel.data.ts:
  153          {
  154:           event: 'click:prev',
  155            description: '',

  164          {
  165:           event: 'click:next',
  166            description: '',

apps\docs\src\data\components\formCheckbox.data.ts:
  101          {
  102:           event: 'update:indeterminate',
  103            description: 'Emitted when the indeterminaate state of the checkbox is changed',

apps\docs\src\data\components\table.data.ts:
  473              ],
  474:             event: 'update:sortBy',
  475              description:
```

**Events with dash**
```
21 results - 12 files

apps\docs\src\data\components\accordion.data.ts:
  176          {
  177:           event: 'hide-prevented',
  178            description:

  196          {
  197:           event: 'show-prevented',
  198            description:

apps\docs\src\data\components\alert.data.ts:
  87          {
  88:           event: 'close-countdown',
  89            description: 'Content to place in the alert',

apps\docs\src\data\components\avatar.data.ts:
  133          {
  134:           event: 'img-error',
  135            args: [

apps\docs\src\data\components\collapse.data.ts:
   88          {
   89:           event: 'hide-prevented',
   90            description: 'Emitted when the Collapse tried to close, but was prevented from doing so.',

  121          {
  122:           event: 'show-prevented',
  123            description: 'Emitted when the Collapse tried to open, but was prevented from doing so.',

apps\docs\src\data\components\formTags.data.ts:
  252          {
  253:           event: 'tag-state',
  254            description: 'Emitted when tags in the user input are parsed',

apps\docs\src\data\components\modal.data.ts:
  363          {
  364:           event: 'hide-prevented',
  365            args: [],

  369          {
  370:           event: 'show-prevented',
  371            args: [],

apps\docs\src\data\components\offcanvas.data.ts:
  152          {
  153:           event: 'show-prevented',
  154            description: '',

  179          {
  180:           event: 'hide-prevented',
  181            description: '',

apps\docs\src\data\components\pagination.data.ts:
  181          {
  182:           event: 'page-click',
  183            description: 'Emitted when a page button was clicked. Cancelable',

apps\docs\src\data\components\popover.data.ts:
  143          {
  144:           event: 'show-prevented',
  145            description: '',

  170          {
  171:           event: 'hide-prevented',
  172            description: '',

apps\docs\src\data\components\tabs.data.ts:
  215            description: 'Emitted just before a tab is shown/activated. Cancelable',
  216:           event: 'activate-tab',
  217            args: [

apps\docs\src\data\components\toast.data.ts:
  143          {
  144:           event: 'close-countdown',
  145            args: [

  163          {
  164:           event: 'hide-prevented',
  165            args: [],

  187          {
  188:           event: 'show-prevented',
  189            args: [],

apps\docs\src\utils\dropdown-common.ts:
  170    {
  171:     event: 'hide-prevented',
  172      description: 'Emitted when the dropdown tried to close, but was prevented from doing so.',

  189    {
  190:     event: 'show-prevented',
  191      description: 'Emitted when the dropdown tried to open, but was prevented from doing so.',

  193    {
  194:     event: 'split-click',
  195      description: 'Emitted when split button is clicked in split mode',
```

dwgray:
> Seems fine for me. Just fix the test error 😁

Sorry @xvaara  - I didn't see your comments until I already made the change. Based on the pattern of existing events (see above) - we could achieve some level of consistency with:
1. What I just did - this makes 'modified' click events be specified with a colon and all other events use a dash. (Except of course for update events, which per Vue use the `update:[modelname]` syntax - I think this is why the colon modifier seems natural for the `click:[descriptor]` events
2. Make all events except for model update events use a dash. Not a big change either way and I'm happy to do it - I think it's worth establishing consistency now and there are only a few events in question.

@VividLemon do you have a strong preference for the colon syntax? If not it sounds like @xvaara prefers the dash syntax. I'll wait on your reply before doing anything more on this one...

VividLemon:
Dash is probably fine

VividLemon:
My thoughts is that it's `[action]:[context]` so there's a strong argument to make it `close:onHide`

dwgray:
I'm leaning towards the dash syntax at this point. There isn't as clear a line as I thought between "modified" click or native events and the state change events as I thought, so explaining that line in the docs (or users intuiting it) seems less hard.

I also missed a bunch of existing events, since I was depending on the docs for my grep, and we don't enforce emits matching the code. I've appended a list of the dash and colon-colon events from BSV. We're implementing the bulk of the dash events, most specifically the table events in BSVN. The root instance `::` events are replaced by local dash syntax events already.

So moving to colon syntax events would require changing a whole bunch more code and documenting the change in the migration guide as well as fixing up the existing documentation. On the other hand, moving the couple of existing instances to dash syntax should be straightforward.

@xvaara - if I move to the dash events for this PR + the two colon syntax carousel events, that shouldn't interfere with your giant PR, right? If that's the case, I'd like to go ahead and do that - we'll then have consistency in event naming and a lower lift for migration.

BSV events with dashes
```
BAlert -- dismiss-count-down
BAvatar -- img-error
BCarousel -- sliding-end
BCarousel -- sliding-start
BFormTags -- tag-state
BPagination -- page-click
BPaginationNav -- page-click
BTable -- context-changed
BTable -- head-clicked
BTable -- row-clicked
BTable -- row-contextmenu
BTable -- row-dblclicked
BTable -- row-hovered
BTable -- row-middle-clicked
BTable -- row-selected
BTable -- row-unhovered
BTable -- sort-changed
BTableLite -- head-clicked
BTableLite -- row-clicked
BTableLite -- row-contextmenu
BTableLite -- row-dblclicked
BTableLite -- row-hovered
BTableLite -- row-middle-clicked
BTableLite -- row-unhovered
BTabs -- activate-tab
```

BSV - global state events using double colon syntax
```
BCollapse -- bv::collapse::state
BDropdown -- bv::dropdown::hide
BDropdown -- bv::dropdown::show
BLink -- bv::link::clicked
BModal -- bv::modal::hidden
BModal -- bv::modal::hide
BModal -- bv::modal::show
BModal -- bv::modal::shown
BPopover -- bv::popover::disabled
BPopover -- bv::popover::enabled
BPopover -- bv::popover::hidden
BPopover -- bv::popover::hide
BPopover -- bv::popover::show
BPopover -- bv::popover::shown
BTooltip -- bv::tooltip::disabled
BTooltip -- bv::tooltip::enabled
BTooltip -- bv::tooltip::hidden
BTooltip -- bv::tooltip::hide
BTooltip -- bv::tooltip::show
BTooltip -- bv::tooltip::shown
```

dwgray:
> @xvaara - if I move to the dash events for this PR + the two colon syntax carousel events, that shouldn't interfere with your giant PR, right? If that's the case, I'd like to go ahead and do that - we'll then have consistency in event naming and a lower lift for migration.

I just went ahead and did this, as it was almost trivial and put us in a consistent state. @VividLemon, if you feel that strongly about the colon syntax, we can revisit later. For now, I'm going to switch to a couple of other projects that I'm running deadlines on and meditate on the [two hard things in computer science](https://martinfowler.com/bliki/TwoHardThings.html) :-)

Issue #2590, opened 2025-03-05
BFormRadioGroup: options html prop missing


noehmeier:
### situation
I updated `bootstrap-vue-next` from version `0.24.1` to `0.27.0`.

I have a `BFormRadioGroup`, where i use the `options` prop, to load the options programmatically. 

```html
<b-form-group
      class="form-group col-sm-3"
      label="Condition"
      label-for="formTagRelCondition"
      label-class="mb-1"
>
      <b-form-radio-group
        id="formTagRelCondition"
        name="formTagRelCondition"
        button-variant="outline-secondary"
        :options="formTagRelConditionOptions"
        buttons
      />
</b-form-group>
```

### problem
In `0.24.1`, the `options` prop allowed an array, that included the attribute `html`, like i used it the following:

```ts
const formTagRelConditionOptions = [
  {
    html: `<span title="equals" class="material-symbols-rounded-equal" /> ICON A`,
    value: FormTagRelConditionType.equals,
  },
  {
    html: `<span title="not equals" class="material-symbols-rounded-code" /> ICON B`,
    value: FormTagRelConditionType.not_equals,
  },
  {
    html: `<span title="is empty" class="material-symbols-rounded-circle" /> ICON C`,
    value: FormTagRelConditionType.empty,
  },
  {
    html: `<span title="is not empty" class="material-symbols-rounded-motion_photos_off" /> ICON D`,
    value: FormTagRelConditionType.not_empty,
  },
];
```

In `0.27.0`, the `options` prop has no `html` prop anymore. [Doc](https://bootstrap-vue-next.github.io/bootstrap-vue-next/docs/components/form-radio.html#component-reference) says, it expects now the type `CheckboxOptionRaw[]`, which i found [here ](https://github.com/bootstrap-vue-next/bootstrap-vue-next/blob/f40abfd5a5fb862f3c4697f68b04454ed85f8482/packages/bootstrap-vue-next/src/types/CheckboxTypes.ts#L16), which expects `CheckboxOption`, which is an array with three attributes called `text`, `value` and `disabled`:

```ts
export type CheckboxOption = {
  text: string
  value: CheckboxValue
  disabled?: boolean
}
```

So there is no `html` anymore? Is this a bug or a feature? If it is a feature, how can i now include my html code based on a programmatically variable?

### reproduction

[...]

PR #2594, opened 2025-03-06
docs: point users to migration guide from getting started

dwgray:
# Describe the PR

Based on feedback from #2590, I've made a couple of changes to the migration guide

- point users to migration guide from getting started
- add a status section to the migration guide
- add an example for replacing `v-html` in a `BRadioGroup`

## Small replication

See the bug

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)


apps/docs/src/docs/demo/RadioGroupMigration.vue
```diff
@@ -0,0 +1,64 @@
...
+enum MyColors {
+  black = 'BLACK',
+  red = 'RED',
+  green = 'GREEN',
+  blue = 'BLUE',
+  purple = 'PURPLE',
+}
+
+const model = ref<RadioValue>(MyColors.black)
+const options = [
```
VividLemon, line 32: const myColors = ['Black', 'Red', 'Green', 'Blue', 'Purple']

const model = ref<RadioValue>(myColors[0])
const options = myColors.map((e) => ({ value: e, html: `<span style="color:${e.toLowerCase()}" /> ${e}` }))

dwgray: Merged your changes so that this example would work and added some explanation of the examples.

VividLemon (approved):

PR #2605, opened 2025-03-12
fix(BModal): fallback focus element always present (#2604)

### LINKED ISSUE #2604: BModal fallback focus element is always rendered
### Describe the bug

The fallback focus element in BModal is always rendered, even when focusable elements (such as the default header close button) are present. Haven't personally encountered any problems caused by it, but is a bug nonetheless.

Will submit a PR with unit tests shortly.

### Reproduction

https://bootstrap-vue-next.github.io/bootstrap-vue-next/docs/components/modal.html

### Used Package Manager

pnpm

IGx89:
# Describe the PR

Fixes #2604. It's checking the template ref element for focusable elements before the element is mounted instead of after, and since template refs are undefined until the component is mounted (noted in the Vue docs) it never finds any.

## Small replication

Open any modal in the docs (https://bootstrap-vue-next.github.io/bootstrap-vue-next/docs/components/modal.html) and you'll see the fallback focus element in the DOM.

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

### DIFF
--- a/packages/bootstrap-vue-next/src/composables/useActivatedFocusTrap.ts
+++ b/packages/bootstrap-vue-next/src/composables/useActivatedFocusTrap.ts
@@ -49,8 +49,9 @@ export const useActivatedFocusTrap = (
     )
     return !tabbableElements || tabbableElements.length === 0
   }
-  const needsFallback = ref(checkNeedsFocus())
+  const needsFallback = ref(false)
   onMounted(() => {
+    needsFallback.value = checkNeedsFocus()
     useMutationObserver(
       element,
       () => {
--- a/packages/bootstrap-vue-next/src/components/BModal/modal.spec.ts
+++ b/packages/bootstrap-vue-next/src/components/BModal/modal.spec.ts
@@ -382,6 +382,43 @@ describe('modal', () => {
     wrapper.unmount()
   })
 
+  it('fallback element is added when no focusable elements are present', async () => {
+    const wrapper = mount(BModal, {
+      attachTo: document.body,
+      global: {
+        stubs: {teleport: true, transition: false},
+      },
+      props: {
+        noHeader: true,
+        noFooter: true,
+      },
+    })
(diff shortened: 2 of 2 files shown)


packages/bootstrap-vue-next/src/composables/useActivatedFocusTrap.ts
```diff
@@ -49,8 +49,9 @@ export const useActivatedFocusTrap = (
     )
     return !tabbableElements || tabbableElements.length === 0
```
VividLemon, line 50: I think this could just be `tabbableElements?.length === 0`

packages/bootstrap-vue-next/src/composables/useActivatedFocusTrap.ts
```diff
@@ -49,8 +49,9 @@ export const useActivatedFocusTrap = (
...
   }
-  const needsFallback = ref(checkNeedsFocus())
+  const needsFallback = ref(false)
   onMounted(() => {
+    needsFallback.value = checkNeedsFocus()
```
VividLemon, line 54: checkNeedsFocus should likely be renamed to `checkNeedsFallback`

VividLemon (changes requested):

packages/bootstrap-vue-next/src/composables/useActivatedFocusTrap.ts
IGx89: Sure. I did `!tabbableElements?.length` to match the existing logic (otherwise if element was undefined it would evaluate to false).

packages/bootstrap-vue-next/src/composables/useActivatedFocusTrap.ts
IGx89: Renamed

VividLemon (approved):

Issue #2611, opened 2025-03-14
docs: "Error: Invalid value for <svg> attribute height="1.1rem"


aforcada:
### Describe the bug

I’ve encountered an issue in Safari 18.3.1 where the following error appears in the console:

`Error: Invalid value for <svg> attribute height="1.1rem"`

I will open a PR to fix this issue.

### Reproduction

https://stackblitz.com/edit/github-r4jgvv4n?file=src%2FApp.vue

### Used Package Manager

pnpm

xvaara:
If I'm not mistaken the correct way would be to give the svg `style="height: 1.1rem"` attribute.

PR #2612, opened 2025-03-14
fix(svg): use em instead of rem for height to avoid Safari error

aforcada:
# Describe the PR

Replace `rem` with `em` in svg height attribute for Safari compatibility.

Related to #2611

## Small replication

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

### DIFF
--- a/apps/docs/.vitepress/theme/Layout.vue
+++ b/apps/docs/.vitepress/theme/Layout.vue
@@ -55,7 +55,7 @@
             rel="noopener"
             link-classes="py-1 px-0"
           >
-            <component :is="link.icon()" height="1.1rem" aria-hidden />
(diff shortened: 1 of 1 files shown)


xvaara:
If I'm not mistaken the correct way would be to give the svg `style="height: 1.1rem"` attribute.

aforcada:
@xvaara  Actually, visually it remains the same, but if for any reason you want to keep the “rem” unit, I’ll modify it with a style as you mentioned.

xvaara:
Oh, what I'm saying is svg specs say:

height
The displayed height of the rectangular viewport.

it changes the viewport if viewPort attrs not defined. But I don't think this is a problem in icons.

xvaara (approved):

PR #2623, opened 2025-03-24
docs(otp): Track on this page with scrollspy

dwgray:
# Describe the PR

Use the `useScrollSpy` composable to track the position of the main document in the "On this page" contents.

## Small replication

See the docs

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

### DIFF
--- a/apps/docs/.vitepress/theme/Layout.vue
+++ b/apps/docs/.vitepress/theme/Layout.vue
@@ -196,8 +196,6 @@ import PageContents from '../../src/components/PageContents.vue'
 const {page} = useData()
 const route = useRoute()
 
-//debugger
-
 const content = useTemplateRef<ComponentPublicInstance<HTMLElement>>('_content')
 const target = useTemplateRef<ComponentPublicInstance<HTMLElement>>('_target')
 
--- a/apps/docs/.vitepress/theme/Layout.vue
+++ b/apps/docs/.vitepress/theme/Layout.vue
@@ -137,11 +137,12 @@
                 header-class="pb-0 d-flex offcanvas-hidden-width"
                 body-class="py-2"
               >
-                <PageContents />
+                <PageContents ref="_target" />
+                <div>Current: {{ current }}</div>
               </BOffcanvas>
             </ClientOnly>
           </aside>
-          <Content class="doc-content" />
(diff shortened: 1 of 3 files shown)


apps/docs/.vitepress/theme/Layout.vue
```diff
@@ -181,11 +191,19 @@ import {useData, useRoute, withBase} from 'vitepress'
...
+import PageContents from '../../src/components/PageContents.vue'
 
 // https://vitepress.dev/reference/runtime-api#usedata
 const {page} = useData()
 const route = useRoute()
 
+//debugger
+
+const content = useTemplateRef<ComponentPublicInstance<HTMLElement>>('_content')
+const target = useTemplateRef<ComponentPublicInstance<HTMLElement>>('_target')
```
dwgray, line 203: @VividLemon @xvaara , any idea why the target ref isn't resolving? In the vue devtools  it's showing up as  `target: Object (empty) (Ref)` while content is showing the expected `content: <VitePressContent(Ref)>`.

I thought it might be related to it being wrapped in a `ClientOnly` component, but I can get the ref to resolve if I put it directly on the `<Boffcanvas>` component or the current `<div>.` 

Then I thought it might have something to do with the conditional rendering of the `<BNav>`, but removing that condition didn't help.

Finally, I thought it might be something to do with being in `<BOverlay>` (although that was probably eliminated by the `<div>` test. Anyway, I tried an overlay example in the `App.vue` below, and both references in that example resolve.

No urgency on this, I'm hoping to eventually get scroll-spy working for the on-this-page sidebar,  but that not blocking anything.

xvaara: @dwgray I don't know. It's probably because there is also component beneath that.

You can use ids for the scrollSpy, and you need to customise the query. something like this:
```js
const {current, list, updateList} = useScrollspy('page-content', 'page-contents', {
  contentQuery: ':scope > div > h2[id], :scope > div > h3[id], h2#component-reference',
})
```

VividLemon: Any more issues?

dwgray: No, this should be good to go.

VividLemon (approved):

PR #2624, opened 2025-03-27
fix(BTab): error in recursion

### LINKED ISSUE #2620: BTab recursive attrs
### Describe the bug

We run into an issue where passing attrs to BTab (In our example style. Also happens when @click is passed in) causes recursive error:

 **Uncaught (in promise) Maximum recursive updates exceeded in component <BTabs>. This means you have a reactive effect that is mutating its own dependencies and thus recursively triggering itself. Possible sources include component template, render function, updated hook or watcher source function.**

The issue doesn't happen when we remove either of those:
- style / @click - attrs from BTab
- < slot :name="tab.identifier"></slot >
- not using v-for on BTab

Reproduction:
```
  <b-tabs>
      <b-tab
        v-for="(tab, index) in [
          { identifier: 'Tab1', label: 'Tab 1' },
          { identifier: 'Tab2', label: 'Tab 2' },
        ]"
        style="background-color: red"
        :key="index"
        :title="tab.label"
      >
        <slot :name="tab.identifier"></slot>
      </b-tab>
    </b-tabs>
```

BTab Snippet that seems to cause the issue:
```
const attrs = useAttrs()
const processedAttrs = computed(() => {
  const {onClick, ...tabAttrs} = attrs
  return {onClick, tabAttrs}
})
```

### Reproduction

https://stackblitz.com/edit/github-ewezexj6?file=src%2Fcomponents%2FComp.vue,src%2FApp.vue

### Used Package Manager

pnpm

xvaara:
fixes #2620

# Describe the PR

## Small replication

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

### DIFF
--- a/packages/bootstrap-vue-next/src/components/BTabs/BTab.vue
+++ b/packages/bootstrap-vue-next/src/components/BTabs/BTab.vue
@@ -7,7 +7,7 @@
     :class="computedClasses"
     role="tabpanel"
     :aria-labelledby="computedButtonId"
-    v-bind="processedAttrs.tabAttrs"
+    v-bind="processedAttrs"
   >
     <slot v-if="showSlot" />
   </component>
@@ -60,8 +60,9 @@ const el = useTemplateRef<HTMLElement>('_el')
 
 const attrs = useAttrs()
 const processedAttrs = computed(() => {
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const {onClick, ...tabAttrs} = attrs
-  return {onClick, tabAttrs}
+  return tabAttrs
 })
 
 const tab = computed(
@@ -75,7 +76,7 @@ const tab = computed(
       titleItemClass: () => props.titleItemClass,
       titleLinkAttrs: () => props.titleLinkAttrs,
       titleLinkClass: () => props.titleLinkClass,
-      onClick: processedAttrs.value.onClick,
+      onClick: attrs.onClick,
       el: el.value,
     }) as TabType
 )


VividLemon:
I don't get it

xvaara:
> I don't get it

I'm not 100% sure I get it either... I think it's so that we can't reference other computed or reactive directly. This could have been converted to a function like the others, but given that with this even this worked:
```vue
<template>
  <BContainer>
    <BRow>
      <BCol>
        <b-tabs>
          <b-tab
            v-for="(tab, index) in [
              {identifier: 'Tab1', label: 'Tab 1'},
              {identifier: 'Tab2', label: 'Tab 2'},
            ]"
            :key="index"
            style="background-color: red"
            :title="tab.label"
            @click="t"
          >
            {{ tab.identifier }}
          </b-tab>
        </b-tabs>
      </BCol>
    </BRow>
  </BContainer>
</template>

<script setup lang="ts">
import {computed, ref} from 'vue'

const d = ref(true)
const t = computed(() => (d.value ? console.warn : console.log))

setInterval(() => {
  d.value = !d.value
}, 1000)
</script>
```

xvaara:
The internally onClick doesn't have to be reactive since it doesn't effect the rendering.

packages/bootstrap-vue-next/src/components/BTabs/BTab.vue
```diff
@@ -60,8 +60,9 @@ const el = useTemplateRef<HTMLElement>('_el')
 
 const attrs = useAttrs()
 const processedAttrs = computed(() => {
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const {onClick, ...tabAttrs} = attrs
```
VividLemon, line 64:
```suggestion
  const {onClick: _, ...tabAttrs} = attrs
```

VividLemon (approved):

PR #2632, opened 2025-03-30
feat(BTable): Expose additional functions and document them

dwgray:
# Describe the PR

- Expose `getStringValue` and `displayedItems`
- Document the above + all the other exposed functions
- Create examples
- Update the migration guide

## Small replication

N/A

## PR checklist

- [x] Feature - `feat(...)`
- [ ] Other (please describe)

### DIFF
--- a/apps/docs/src/docs/components/demo/TableComplete.vue
+++ b/apps/docs/src/docs/components/demo/TableComplete.vue
@@ -111,6 +111,7 @@
     </BRow>
     <!-- Main table element for typed table-->
     <BTable
+      ref="complete-table"
       v-model:sort-by="sortBy"
       :sort-internal="true"
       :items="items"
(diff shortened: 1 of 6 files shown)


apps/docs/src/docs/components/demo/TableComplete.vue
```diff
@@ -177,6 +189,10 @@ interface Person {
   isActive: boolean
 }
 
+const table = useTemplateRef('complete-table')
+
+const displayItems = () => table.value?.displayItems ?? []
```
VividLemon, line 194: This seems like it should be a computed

apps/docs/src/docs/components/table.md
```diff
@@ -743,6 +754,14 @@ with a custom comparer:
 
 <<< FRAGMENT ./demo/TableSortCompareCustom.ts#snippet{ts}
 
+`getStringValue` is exposed to allow for customization of the simple string sorts by calling `localeCompare`
```
VividLemon, line 757:
```suggestion
The default sorting algorithm parses object syntax by using an internal function `getStringValue`. It's main purpose is to fetch the correct value from an object, say `{foo: {bar: '1'}}`, you can use `'foo.bar'` as an input as well as returning the final result from formatting. This function is exposed to allow for customization.
```

apps/docs/src/docs/components/demo/TableComplete.vue
dwgray: Yep, fixed. There was a bit of an issue where changing it to computed forced me to explicitly type the template reference, but I think that's all right.

VividLemon (approved):

Issue #2637, opened 2025-04-02
BButton doesn't support text-variant/bg-variant


hackel:
### Describe the bug

In #2200, info about the text-variant and bg-variant props was added to the BButton documentation, but it doesn't seem that the feature was actually implemented in BButton.

### Reproduction

*trivial*

### Used Package Manager

pnpm

Issue #2653, opened 2025-04-09
Button styling inconsistent due to "fix(BButton): Consume useColorVariantClasses #2640"


krasendimitrov:
### Describe the bug

The merge request introduced the class "text-bg-{variant}" to the button component which makes following changes to the previous behavior.:

- Hover effects via background hover no longer work since they are now overwritten
- customization of the button text color is more diffictult if someone wants a different color instead of the automaticaly assigned by bootstrap, since the color attribute has "!important"

I personally do not understand the reasoning of adding this class to the button, since it is not part of the standard bootstrap implementation and also not part of  "bootstrap-vue"  Button component implementation.

Relates to: #2640 

### Reproduction

https://stackblitz.com/edit/bootstra-vue-next-template-3ixpznbo?file=src%2FApp.vue,package.json,src%2Fmain.js,src%2Fmain.scss

### Used Package Manager

pnpm

ptdev:
Hi,

Also experiencing this.

This update messed up pretty much all of our buttons.

Removing the text-bg-* class from the buttons appear to solve the issue as hinted by @krasendimitrov

PR #2640, opened 2025-04-03
fix(BButton): Consume useColorVariantClasses

dwgray:
# Describe the PR

When I was documenting and fixing up `useColorVariantClasses`, I missed the fact that BButton didn't actually consume that composible. This fixes that oversight.

Fixes #2637

## Small replication

```
<template>
  <BRow
    ><BCol>Set 'variant': <BButton variant="success"> testing </BButton></BCol></BRow
  >
  <BRow
    ><BCol>Set 'bg-variant': <BButton bg-variant="info"> testing </BButton></BCol></BRow
  >
  <BRow
    ><BCol
      >Set 'text-variant' (background is 'secondary' since 'variant' defaults to 'secondary'):
      <BButton text-variant="info"> testing </BButton></BCol
    ></BRow
  >
  <BRow
    ><BCol
      >Set 'text-variant' and 'bg-variant':
      <BButton text-variant="primary" bg-variant="danger"> testing </BButton></BCol
    ></BRow
  >
  <BRow
    ><BCol
      >Set all three - 'variant' is overridden by 'bg-variant' and 'text-variant':
      <BButton text-variant="primary" bg-variant="danger" variant="success">
        testing
      </BButton></BCol
    ></BRow
  >
</template>
```

## PR checklist

**What kind of change does this PR introduce?** (check at least one)

- [x] Bugfix :bug: - `fix(...)`
- [ ] Feature - `feat(...)`
- [ ] ARIA accessibility - `fix(...)`
- [ ] Documentation update - `docs(...)`
- [ ] Other (please describe)

**The PR fulfills these requirements:**

- [x] Pull request title and all commits follow the [**Conventional Commits**](https://www.conventionalcommits.org/) convention or has an [**override**](https://github.com/googleapis/release-please#how-can-i-fix-release-notes) in this pull request body **This is very important, as the `CHANGELOG` is generated from these messages, and determines the next version type. Pull requests that do not follow conventional commits or do not have an override will be denied**

PR #2654, opened 2025-04-09
Revert "fix(BButton): Consume useColorVariantClasses (#2640)"

dwgray:
This reverts commit 379fd9acc185f87d806d62e4be4c8045a59da485.

# Describe the PR

As noted in #2653, the change in #2640 caused more issues than it fixed. I believe that it is because useColorVariant uses the variant property to set `text-bg-*` which overrides the `btn-variant-*` class in the .css.

I'd like to revert this change outright + remove the confusing docs.

Longer term we could consider implementing `bg-variant` and `text-variant` on `BButton` and possibly other components that support them, but for the short term a perfectly viable solution is to treat these as utility classes and have the user set them manually. @hackel - can you confirm that this would work-around the issue that you filed #2637 for?

## Small replication

See #2653

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

### DIFF
--- a/apps/docs/src/data/components/button.data.ts
+++ b/apps/docs/src/data/components/button.data.ts
@@ -55,7 +55,7 @@ export default {
                 default: 'secondary',
               },
             }),
-            ['bgVariant', 'size', 'tag', 'textVariant', 'variant']
+            ['size', 'tag', 'variant']
           ),
         } satisfies Record<
           Exclude<keyof BvnComponentProps['BButton'], keyof typeof linkProps>,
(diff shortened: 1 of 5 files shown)


VividLemon (approved):

hackel:
@dwgray Yes, that works for me, I think that's what I am already doing.  Thanks!

dwgray:
> @dwgray Yes, that works for me, I think that's what I am already doing. Thanks!

Okay, I've closed #2637 based on this, if you think we should keep it open to track potentially adding properties to support this please go ahead and reopen with a comment to that effect.

PR #2664, opened 2025-04-17
fix(BTabs): corrent classes on ssr

### LINKED ISSUE #2652: BTabs only becomes visible on mounted, resulting in CLS
### Describe the bug

This is maybe related to my other issue which was recently resolved #2601 

The `BTabs` component has no visible content rendered during SSR in a Nuxt environment. The `<ul>` element with the tab headers is still empty and all of the tab content `<div>` elements are still set to `display: none`. On client side during `mounted`, the tab headers are then rendered and the active tab content is made visible via the `active` class. I believe this is different behaviour from bootstrap-vue 2.x

This behaviour creates a cumulative layout shift, among other issues for me related to the tab content not being visible yet.

I provided a Stackblitz demo but it is difficult to demonstrate because the Stackblitz preview does not work with JS disabled. It is just a very basic `BTabs` example taken from the documentation.

### Reproduction

https://stackblitz.com/edit/nuxt-starter-mlet8bwm?file=app.vue

### Used Package Manager

yarn

xvaara:
fixes #2652

# Describe the PR

## Small replication

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved tab registration and activation timing for more reliable tab behavior.
  - Enhanced fallback logic to render tabs from slot content when internal registration is unavailable.
- **New Features**
  - Added a method to explicitly sort tabs, improving tab order consistency.
  - Introduced automatic flattening of nested tab fragments for better tab rendering.
  - Updated tab click handling for improved event processing.
  - Added support for custom navigation item classes on tabs.

### DIFF
--- a/packages/bootstrap-vue-next/src/components/BTabs/BTab.vue
+++ b/packages/bootstrap-vue-next/src/components/BTabs/BTab.vue
@@ -14,7 +14,7 @@
 </template>
 
 <script setup lang="ts">
-import {computed, inject, onMounted, onUnmounted, ref, useAttrs, useTemplateRef, watch} from 'vue'
+import {computed, inject, onUnmounted, ref, useAttrs, useTemplateRef, watch} from 'vue'
 import {useId} from '../../composables/useId'
 import {useDefaults} from '../../composables/useDefaults'
 import type {TabType} from '../../types/Tab'
@@ -81,13 +81,12 @@ const tab = computed(
     }) as TabType
 )
 
-onMounted(() => {
-  if (!parentData) return
+if (parentData) {
   parentData.registerTab(tab)
   if (activeModel.value) {
     parentData.activateTab(computedId.value)
   }
-})
+}
 
 onUnmounted(() => {
   if (!parentData) return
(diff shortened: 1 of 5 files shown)


xvaara:
@VividLemon the way we register the tabs, they arent there on the first pass so this is kinda a hack to get them working in the setup context for ssr.

xvaara:
@VividLemon do you see any other way how we could get the ssr context working? I'm not a fan of using the VNode structure, but I couldn't figure another way to fix this.

VividLemon:
> @VividLemon do you see any other way how we could get the ssr context working? I'm not a fan of using the VNode structure, but I couldn't figure another way to fix this.

Due to the immense overhead of this component, it would seem better to introduce another library to handle this, like Radix-vue.

xvaara:
> > @VividLemon do you see any other way how we could get the ssr context working? I'm not a fan of using the VNode structure, but I couldn't figure another way to fix this.
> 
> Due to the immense overhead of this component, it would seem better to introduce another library to handle this, like Radix-vue.

I don't really see how that would make any sense. They have a totally different way of doing tabs, totally different api. 
The basic philosophy of bootstrap-vue-next is to to non-verbose and do clever stuff. Reka-ui (new name for Radix-vue) is the opposite, it's very verbose, doesn't do stuff without you defining everything. 

And we'd have the same problems with it if we don't change the api (ie. use another component for the tab button, and another for tab content) and leave the problems to the user.

PR #2670, opened 2025-04-25
Fix patch for issue #2665

tbl0605:
Hi,
This patch fixes what appears to be a bug introduced by the fix for issue #2665.
My patch hasn't been tested at all. Consider it a reminder that there's some code in bootstrap-vue-next that needs to be fixed before it gets forgotten ;)

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved button visibility logic in pagination controls to ensure only visible buttons are included.
	- Enhanced keyboard navigation tests for pagination to better simulate button visibility and focus behavior.

### DIFF
--- a/packages/bootstrap-vue-next/src/components/BPagination/BPagination.vue
+++ b/packages/bootstrap-vue-next/src/components/BPagination/BPagination.vue
@@ -359,7 +359,7 @@ const getButtons = (): HTMLButtonElement[] =>
 
       const bcr = el.getBoundingClientRect()
 
-      return true || !!(bcr && bcr.height > 0 && bcr.width > 0)
+      return !!(bcr && bcr.height > 0 && bcr.width > 0)
     })
     ?.map((el) => el as HTMLButtonElement)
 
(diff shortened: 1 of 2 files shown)


VividLemon:
The unit test failed.

tbl0605:
> The unit test failed.

Hi @VividLemon,
I haven't fixed the unit tests (yet), as they seem to be "technically correct", but it's the `return !!(bcr && bcr.height > 0 && bcr.width > 0)` condition that isn't working as expected with the unit tests.

Could it be because of this [comment](https://github.com/bootstrap-vue-next/bootstrap-vue-next/blob/340edfdb1f5fffc57c8bbc4de735bc4b2f7f0c13/packages/bootstrap-vue-next/src/utils/dom.ts#L42) ?

```
  // All browsers support getBoundingClientRect(), except JSDOM as it returns all 0's for values :(
  // So any tests that need isVisible will fail in JSDOM
  // Except when we override the getBCR prototype in some tests
```

If this is the case, the unit tests will always fail, even if the `return !!(bcr && bcr.height > 0 && bcr.width > 0)` condition is correct.
I think the patch author used `return true` to make the unit tests work on his computer, but that's just a guess ;)

So what's the best solution? Remove the unit tests?

VividLemon (approved):

VividLemon:
For future reference please use https://www.conventionalcommits.org/en/v1.0.0/

PR #2679, opened 2025-05-01
docs: implement on this page expand/collapse with useScrollspy

dwgray:
# Describe the PR

This PR gets the on this page table of contents up to parity with the BSV version by driving the OTP off of the scrollspy data rather than the page data.
- This is a clean way to get the component reference sub-nodes, which can be important for larger component complexes like BTable
- I am using the active node to show sub-nodes in the tree only if they are part of the currently active branch, which makes it easier to see where you are in larger components and migration, where having everything expanded was unwieldy
- I can show a single node as active rather than the funky state where the parents of the active node were also being shown 

## Small replication

docs

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

## Summary by CodeRabbit

- **New Features**
  - Enhanced the "On this page" sidebar with a structured, hierarchical navigation of page headers, providing improved overview and navigation for documentation pages.
  - Sidebar now highlights the currently active section as you scroll.

- **Improvements**
  - Component references in documentation now display as anchorable headings, making it easier to link directly to specific components.
  - Sidebar navigation dynamically expands to show nested sections related to the active header.

- **Refactor**
  - Improved sidebar navigation logic for better accuracy and maintainability.


apps/docs/.vitepress/theme/Layout.vue
```diff
@@ -138,7 +138,7 @@
                 header-class="pb-0 d-flex offcanvas-hidden-width"
                 body-class="py-2"
               >
-                <PageContents />
+                <PageContents :contents="contents" :active-id="currentItem" />
               </BOffcanvas>
```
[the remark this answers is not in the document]
dwgray: currentItem is the variable returned from useScrollSpy as current. This is actually the id of the currently active item - a string | null. I'm renaming the variable to avoid this confusion, but not accepting the suggested change, since it wouldn't work - you can't take the id field of a string | null.

apps/docs/.vitepress/theme/Layout.vue
```diff
@@ -295,6 +298,47 @@ const set = (newValue: keyof typeof map) => {
...
+  })
+
+  if (root.length !== 1) {
+    // Something isn't right if we have no root items or more than one root item
+    // eslint-disable-next-line no-console
+    console.warn('Unexpected header structure:', headers, 'Root items:', root)
+  }
+
+  return root.length > 0 ? root[0] : undefined
+})
```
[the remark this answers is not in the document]
dwgray: This code rund in a controlled context where there will always be a single root and there will be other issues if there are multiple roots, emitting a warning if the content is malformed seems reasonable here.

apps/docs/src/components/PageContentsItem.vue
```diff
@@ -1,18 +1,33 @@
...
+    <template #default>{{ props.item.text }}</template>
     <template #after>
-      <BNav v-if="item.children?.length > 0" vertical small class="otp-nav">
-        <PageContentsItem v-for="child in item.children" :key="child.slug" :item="child" />
+      <BNav v-show="isVisible" vertical small class="otp-nav">
+        <PageContentsItem
+          v-for="child in item.children"
+          :key="child.id!"
+          :item="child"
+          :active-id="activeId"
```
VividLemon, line 10: Somewhere along the lines here you made a https://vuejs.org/guide/best-practices/performance#props-stability issue

VividLemon (changes requested):

apps/docs/src/components/PageContentsItem.vue
dwgray: Interesting - this is not quite as easily fixed as the vue example shows, since PageContentsItems is a recursive component - I can't just pass in a boolean. The solution I pushed was to check if the currently active id is in the subtree for the component and pass in null if it isn't. That maintains prop stability for everything except for the subtree that is activating and the one that is deactivating at the cost of extra complexity and computation in our code.

I'm not sure that's the `right` solution here, as I don't have a strong gut on the performance characteristics. In other contexts, I would run a profiler on the code before and after to get that sense. @VividLemon, do you have a gut sense of where you'd take the tradeoffs here?

VividLemon: I agree that prop drilling down twice is probably not the most ideal. It could just be larger benefit to revert the changes for simplicity. It's a minor perf change

dwgray: Thank - That's where my intuition goes as well, but it was interesting to go down the path of keeping the props stable. I've reverted the change.

PR #2701, opened 2025-05-17
docs(BPopover): Parity pass

### LINKED ISSUE #2702: disabled prop/programmatically disabling popovers not enabnled
### Missing feature is from

bootstrap-vue

### Docs

https://bootstrap-vue.org/docs/components/popover#programmatically-disabling-popover

### Describe the missing feature

I'm not entirely sure how useful this is, but it's implemented in BSV (see link).

If we implement this, we should also look at the `enabled` and `disabled` events that are tied to it.

I don't think this is required for a v1, since it should be additive and not a breaking change.

dwgray:
# Describe the PR

Parity pass for `BPopover` - note that is just for the component, not the directive or composable

- filled out documentation
- filled out component reference
- updated the migration guide
- filled out the parity spreadsheet
- added  missing types to `types.md`
- created issues for missing features
- fixed small issues with `variant` still being declared on the component and ordering of props
- fixes #2702 by documenting the props/events as deprecated in the migration guide along with a work-around

## Small replication

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

## Summary by CodeRabbit

- **New Features**
  - Added multiple demo components illustrating popover usage, positioning, triggers, styling, and programmatic control.
  - Introduced new prop `hideMargin` and support for `titleClass` to enhance popover customization.
  - Extended popover events and slot definitions for improved interaction handling.
  - Added comprehensive Popover documentation, including usage guides, migration notes, and type definitions.

- **Bug Fixes**
  - Corrected default values and enhanced descriptions for various popover properties.

- **Refactor**
  - Removed the deprecated `variant` prop from the Popover component and associated styling logic.
  - Reorganized prop declarations and event handling for clearer, more consistent behavior.

- **Documentation**
  - Fully expanded and restructured Popover docs with demos and detailed migration guidance.
  - Improved prop and event descriptions for better developer understanding.

- **Chores**
  - Exported new utility constants for standardized show/hide events and slot data.


apps/docs/src/docs/components/demo/PopoverVariant.vue
```diff
@@ -0,0 +1,14 @@
...
+  <!-- #region template -->
+  <div class="d-flex gap-2">
+    <BButton id="popover-variant">Button</BButton>
+    <BPopover
+      target="popover-variant"
+      title-class="text-bg-danger"
+      body-class="text-bg-danger"
+      title="Danger!"
+      body="Danger variant popover"
+    />
```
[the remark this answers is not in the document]
dwgray: This is wrong. In BSVN we've changed the `content` prop to `body` to make it consistent with other components. This was changed in an earlier PR and is documented in the migration guide changes in the PR.

apps/docs/src/docs/components/popover.md
```diff
@@ -1,85 +1,139 @@
...
-</HighlightCard>
+## Exposed functions
 
-<ComponentReference :data="data" />
+`BCollapse` exposes several functions allow manipulation of the state of the component.
+These are accessed through the [template ref](https://vuejs.org/guide/essentials/template-refs.html#template-refs)
 
-<script setup lang="ts">
-import { computed, ref, onMounted } from 'vue';
+<<< DEMO ./demo/PopoverExposed.vue
```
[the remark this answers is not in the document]
dwgray: Thanks, I've pushed a change to address this. I'm not taking the suggestion as an autocommit because it's unclear from the diff that the surrounding code is correctly preserved.

apps/docs/src/docs/components/popover.md
```diff
@@ -1,85 +1,139 @@
...
+</PageHeader>
+
+<<< DEMO ./demo/PopoverOverview.vue#template{vue-html}
+
+## Overview
+
+Things to know when using the popover component:
+
+- Popovers rely on the 3rd party library [floating-ui](https://floating-ui.com/docs) for positioning.
+- Popovers require BootstrapVueNext's custom SCSS/CSS to function correctly.
```
xvaara, line 18: We don't require custom css, it's just a small hack to get the arrow match the header when position is bottom

apps/docs/src/docs/components/popover.md
```diff
@@ -1,85 +1,139 @@
...
+- Specify `container` as `null` (default, appends to `<body>`) to avoid rendering problems in more
+  complex components (like input groups, button groups, etc.). You can use `container` to optionally
+  specify a different element to append the rendered popover to.
```
xvaara, line 21: There is no container anymore, it's teleportTo, and it's default undefined. Should use #teleport div for it if needed since that is ssr safe.

apps/docs/src/docs/components/popover.md
```diff
@@ -1,85 +1,139 @@
...
+- Triggering popovers on hidden elements will not work.
+- Popovers for `disabled` elements must be triggered on a wrapper element.
+- When triggered from hyperlinks that span multiple lines, popovers will be centered. Use
+  `white-space: nowrap;` on your `<a>`s, `<BLink>`s and `<router-link>`s to avoid this behavior.
```
xvaara, line 25: we speficaly have a prop inline for this. It's a floatingUi middleware.

apps/docs/src/docs/components/popover.md
```diff
@@ -1,85 +1,139 @@
...
+## Target
+
+The target is the _trigger_ element (or component) that will trigger the popover. The target is
+specified via the `target` prop, and can be any of the following:
+
+- A string identifying the ID of the trigger element (or ID of the root element of a component)
+- A reference (ref) to an `HTMLElement` or an `SVGElement` via a [Template Ref](https://vuejs.org/guide/essentials/template-refs.html)
+- A reference (ref) to a component that has either an `HTMLElement` or `SVGElement` as its root
+  element via [Template Ref](https://vuejs.org/guide/essentials/template-refs.html)
+- <NotYetImplemented/> A function (callback) that returns a reference to an `HTMLElement` or `SVGElement`
```
xvaara, line 36: I think we left the function callback out because we couldn't think of a reason for it's existing and I caused some problems with reactivity.

apps/docs/src/docs/components/popover.md
```diff
@@ -1,85 +1,139 @@
...
+- Specify `container` as `null` (default, appends to `<body>`) to avoid rendering problems in more
+  complex components (like input groups, button groups, etc.). You can use `container` to optionally
+  specify a different element to append the rendered popover to.
+- Triggering popovers on hidden elements will not work.
+- Popovers for `disabled` elements must be triggered on a wrapper element.
+- When triggered from hyperlinks that span multiple lines, popovers will be centered. Use
+  `white-space: nowrap;` on your `<a>`s, `<BLink>`s and `<router-link>`s to avoid this behavior.
+
+## Target
+
```
xvaara, line 28: I'd prefer we use the target slot.

I did want to use the default slot for the target and body/title slots for the popover (like some other frameworks do and seems clearer for me), but someone wanted it the other way, so now the body is in default slot and title slot for title, target slot for the target.

apps/docs/src/docs/components/popover.md
```diff
@@ -1,85 +1,139 @@
...
+- A reference (ref) to a component that has either an `HTMLElement` or `SVGElement` as its root
+  element via [Template Ref](https://vuejs.org/guide/essentials/template-refs.html)
+- <NotYetImplemented/> A function (callback) that returns a reference to an `HTMLElement` or `SVGElement`
+
+:::info NOTE
+
+The target element **must** exist in the document before `<BPopover>` is mounted. If the target
+element is not found during mount, the popover will never open. Always place your `<BPopover>`
+component lower in the DOM than your target element. This rule also applies if a callback function
+is used as the target element, since that callback is called only once on mount.
```
xvaara, line 43: Is this still a problem with vue3?
And we don't have callbacks

dwgray: Just tested and things work fine if the target is in the DOM below the popover, so I'm just deleting this whole section.

apps/docs/src/docs/components/popover.md
```diff
@@ -1,85 +1,139 @@
...
+
+By default, popovers are shown by `pointerenter` and `focus` events and closed by `pointerleave` and `blur` events
+on the `target` element by default. To override this behavior and make the popover show and hide based
+on `click` events, set the `click` prop to `true`.
+
+<<< DEMO ./demo/PopoverTriggers.vue#template{vue-html}
+
+To take finer control of popover visibility, set the `manual` prop and use the
+[`v-model`](#programmatic-control-via-v-model) or
+[exposed functions](#exposed-functions) to control visibility.
```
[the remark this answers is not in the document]
dwgray: Added pointers to useToggle and usePopoverController before the manual option.

apps/docs/src/docs/components/popover.md
dwgray: I put in a reference to the target slot before the details about the target prop. I think it's still worth keeping the details on the target prop here, though.

apps/docs/src/docs/components/popover.md
dwgray: Got it - I should have gone back through this after I went through the component data stuff...

apps/docs/src/docs/components/popover.md
dwgray: Got it. I changed this to note the teleport props and point to the vue.js teleport docs. I'm a little confused about what the behavior is here - it looks like `teleportTo` defaults to `undefined` and `teleportDisabled` defaults to false. The actual behavior seems to be to "teleport" the popover to the DOM directly after the placeholder. Is this intentional? Am I missing something?

```html
<div class="d-flex gap-2"><button class="btn btn-md btn-primary" type="button" id="popover-props">Using
        properties</button><span data-v-d01fe26b="" id="BootstrapVueNext__ID__v-169__popover____placeholder"
        style="display: none;"></span>
    <div data-v-d01fe26b="" id="BootstrapVueNext__ID__v-169__popover___"
        content="Embedding content using properties is easy" class="popover b-popover fade bs-popover-top"
        role="tooltip" tabindex="-1"
        style="position: absolute; left: 0px; top: 0px; transform: translate(688px, 4787.2px); display: none;">
        <div data-v-d01fe26b="" class="popover-arrow" data-popper-arrow="" style="position: absolute;"></div>
        <div data-v-d01fe26b="" class="overflow-auto">
            <div data-v-d01fe26b="" class="position-sticky top-0 popover-header">Prop Examples</div>
            <div data-v-d01fe26b="" class="popover-body">Embedding content using properties is easy</div>
        </div>
    </div><button class="btn btn-md btn-primary" type="button" id="popover-slots">Using slots</button><span
        data-v-d01fe26b="" id="BootstrapVueNext__ID__v-170__popover____placeholder" style="display: none;"></span>
    <div data-v-d01fe26b="" id="BootstrapVueNext__ID__v-170__popover___" class="popover b-popover fade bs-popover-top"
        role="tooltip" tabindex="-1"
        style="position: absolute; left: 0px; top: 0px; display: none; transform: translate(818.4px, 4787.2px);">
        <div data-v-d01fe26b="" class="popover-arrow" data-popper-arrow="" style="position: absolute;"></div>
        <div data-v-d01fe26b="" class="overflow-auto">
            <div data-v-d01fe26b="" class="position-sticky top-0 popover-header">Content via Slots</div>
            <div data-v-d01fe26b="" class="popover-body"> Embedding content <span class="text-danger">using slots</span>
                affords you <em>greater <strong>control.</strong></em> and basic HTML support. </div>
        </div>
    </div>
</div>
```

dwgray:
@xvaara Thanks for the thorough review - there were enough fundamental changes with how things work that I'm not surprised I missed some things. I believe I addressed all of your comments with my most recent commit. I did have one question about the default teleport behavior (above).

Quick process question - do you prefer to be the one that resolves your comments, or should I resolve the ones that I believe I've addressed (I've worked on teams that do it both ways...)

apps/docs/src/docs/migration-guide.md
```diff
@@ -594,13 +594,41 @@ See [Show and Hide](#show-and-hide) shared properties.
...
+The `container` prop has been deprecated, the `teleportTo` prop instead to specify where the popover
+should be mounted. See [Vue Teleport documentation](https://vuejs.org/guide/built-ins/teleport.html).
+
+`custom-class` has been changed to `body-class` and a `title-class` has been added for completeness.
+
+`fallback-placement` has been deprecated. Use the various options provided by [floating-ui](https://floating-ui.com/)
+to handle placement.
+
+The ability for the `target` prop to take a function has been deprecated.
+
```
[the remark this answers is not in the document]
dwgray: This line is stating that `target` is deprecated and is included in the migration guide to provide guidance for developers who may have depended on this behavior in previous versions.

xvaara:
@dwgray Awesome work! I can go and close the comments, don't really mind either way.

apps/docs/src/docs/migration-guide.md
```diff
@@ -594,13 +594,39 @@
...
+see [our documentation](/docs/components/popover#custom-classes-and-variants) for details.
+
+`delay` now defaults to 100ms for show and 300ms for hide rather than 50ms for both
+
+The default for `placement` is now `top` rather than `right`
+
+`$root` events are deprecated. See [usePopoverController](/docs/composables/usePopoverController) as an alternative.
+
+<NotYetImplemented/>The `disabled` prop and
+[Programmatically Disabling](https://bootstrap-vue.org/docs/components/popover#programmatically-disabling-popover)
```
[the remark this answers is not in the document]
dwgray: The <NotYetImplemented/> component includes the text to note that the following information is "Not Yet Implemented", the suggested change would be redundant.

apps/docs/src/docs/components/popover.md
dwgray: Got it - clarified the docs in this PR.

apps/docs/src/docs/components/demo/PopoverOverview.vue
```diff
@@ -0,0 +1,15 @@
...
+  <div class="d-flex gap-2">
+    <BButton v-b-popover.focus.top="'I am popover directive content!'" title="Popover Title">
+      Hover Me
+    </BButton>
+
+    <BButton id="popover-target-1"> Hover Me </BButton>
+    <BPopover target="popover-target-1" placement="top">
+      <template #title>Popover Title</template>
+      I am popover <b>component</b> content!
+    </BPopover>
```
xvaara, line 12:
```suggestion
    <BPopover>
      <template #target> <BButton> Hover Me </BButton></template>
      <template #title>Popover Title</template>
      I am popover <b>component</b> content!
    </BPopover>
```

I'd prefer we suggest the slot usage over having to put id:s and targets on elements, less errorprone.

apps/docs/src/docs/components/popover.md
```diff
@@ -1,85 +1,132 @@
...
+- When triggered from hyperlinks that span multiple lines, popovers will be centered. Set the `inline` prop to improve the positioning see the [Floating UI docs](https://floating-ui.com/docs/inline) for details.
+
+## Target
+
+The target is the _trigger_ element (or component) that will trigger the popover. The target is
+specified via the `target` slot or prop, and can be any of the following:
+
+The `target` prop may be any of the following:
+
+- A string identifying the ID of the trigger element (or ID of the root element of a component)
```
xvaara, line 30:
```suggestion
- A string identifying the ID of the trigger element (or ID of the root element of a component)
- A string with querySelector. (ie. '#toolbar > div:first-child')
```

apps/docs/src/docs/components/demo/PopoverTriggers.vue
```diff
@@ -0,0 +1,10 @@
+<template>
+  <!-- #region template -->
+  <div class="d-flex gap-2">
+    <BButton id="popover-target-default-trigger"> Default Trigger </BButton>
+    <BPopover target="popover-target-default-trigger" placement="top" body="Default Trigger" />
+    <BButton id="popover-target-click-trigger"> Click Trigger </BButton>
+    <BPopover target="popover-target-click-trigger" click placement="top" body="Click Trigger" />
```
xvaara, line 7:
```suggestion
    <BPopover body="Default Trigger">
      <template #target>
        <BButton> Default Trigger </BButton>
      </template>
    </BPopover>
    <BPopover click body="Click Trigger">
      <template #target>
        <BButton> Click Trigger </BButton>
      </template>
    </BPopover>
```

I'd prefer the examples syntax to be "narrow" so that it's more readable on mobile

apps/docs/src/docs/components/demo/PopoverContent.vue
```diff
@@ -0,0 +1,18 @@
...
+    <BPopover
+      target="popover-props"
+      title="Prop Examples"
+      content="Embedding content using properties is easy"
+    />
+    <BButton id="popover-slots" variant="primary">Using slots</BButton>
+    <BPopover target="popover-slots">
+      <template #title>Content via Slots</template>
+      Embedding content <span class="text-danger">using slots</span> affords you
+      <em>greater <strong>control.</strong></em> and basic HTML support.
```
xvaara, line 14:
```suggestion
      <em>greater <strong>control.</strong></em> and HTML support.
```
All the html is usable, since we don't need to jump through hoops to support it (like it was in bv2 and directives)

apps/docs/src/docs/components/demo/PopoverVariant.vue
```diff
@@ -0,0 +1,14 @@
+<template>
+  <!-- #region template -->
+  <div class="d-flex gap-2">
+    <BButton id="popover-variant">Button</BButton>
+    <BPopover
+      target="popover-variant"
+      title-class="text-bg-danger"
+      body-class="text-bg-danger"
```
xvaara, line 8:
```suggestion
      class="text-bg-danger"
```

the body-class isn't that great with backgrounds, since bootstrap css doesn't contain the rounded borders for it, like it does for header. So using it makes the bottom non-rounded.

apps/docs/src/docs/components/demo/PopoverStyles.vue
```diff
@@ -0,0 +1,20 @@
+<template>
+  <!-- #region template -->
+  <div class="d-flex gap-2">
+    <BButton id="popover-style">Button</BButton>
+
+    <BPopover
+      :model-value="true"
```
xvaara, line 7:
```suggestion
      show
```
true doesn't work here. It need to be variable.

apps/docs/src/docs/components/demo/PopoverModel.vue
```diff
@@ -0,0 +1,21 @@
+<template>
+  <div class="d-flex flex-column text-md-center">
+    <div class="p-2">
+      <BButton id="popover-button-sync" variant="primary">I have a popover</BButton>
```
xvaara, line 4:
```suggestion
      <p>Popover is {{ model ? 'visible' : 'hidden' }}</p>
```

apps/docs/src/docs/components/demo/PopoverStartOpen.vue
```diff
@@ -0,0 +1,9 @@
+<template>
+  <div class="text-center">
+    <BButton id="popover-button-open" variant="primary">Button</BButton>
+
+    <BPopover :model-value="true" target="popover-button-open" title="Popover">
```
xvaara, line 5:
```suggestion
    <BPopover show target="popover-button-open" title="Popover">
```

apps/docs/src/docs/components/popover.md
```diff
@@ -1,85 +1,132 @@
...
+<<< DEMO ./demo/PopoverStyles.vue#template{vue-html}
+
+## Programmatic control via v-model
+
+You can manually control the visibility of a popover via the v-model. Setting it to true will show the popover,
+while setting it to false will hide the popover.
+
+<<< DEMO ./demo/PopoverModel.vue
+
+To make the popover shown on initial render, simply set `model-value` on `<BPopover>` to true:
```
xvaara, line 101:
```suggestion
To make the popover shown on initial render, simply add prop `show` to `<BPopover>`:
```
As commented above native `true` and `false` don't work anymore with vue3 defineModel (it worked with vueuse useVModel that we previously used)

apps/docs/src/docs/components/demo/PopoverCloseOnHide.vue
```diff
@@ -0,0 +1,31 @@
...
+        <BButton>Click me. Popover closes when clipped</BButton>
+      </template>
+      Scroll me out of view
+    </BPopover>
+    <BPopover
+      :click="true"
+      :close-on-hide="true"
+      :delay="{show: 0, hide: 0}"
+      :boundary-padding="{top: navHeight}"
+    >
```
xvaara, line 14:
```suggestion
    <BPopover :click="true" :delay="{show: 0, hide: 0}" :boundary-padding="{top: navHeight}">
```

I think this is supposed not to close

apps/docs/src/utils/showhide-emits.ts
```diff
@@ -0,0 +1,69 @@
...
+    event: 'shown',
+    args: [
+      {
+        arg: 'value',
+        type: 'BvTriggerableEvent',
+      },
+    ],
+    description: 'Always emits just after component is shown.',
+  },
+]
```
xvaara, line 69: We also emit `toggle` but that can only be used from the exposed functions / composables. I don't know if that needs documentation.

apps/docs/src/utils/showhide-slotsdata.ts
```diff
@@ -0,0 +1,35 @@
...
+  },
+  {
+    prop: 'toggle',
+    type: '() => void',
+    description: 'Function to toggle the component visibility',
+  },
+  {
+    prop: 'active',
+    type: 'boolean',
+    description: 'Indicates if the component is active (shown)',
```
xvaara, line 28:
```suggestion
    description: 'Indicates if the component is active (starting show, before/after animations)',
```

apps/docs/src/data/components/popover.data.ts
```diff
@@ -97,129 +117,114 @@ export default {
...
+          description:
+            'Emitted when the pointer leaves the target element. This event is not emitted when the pointer leaves the popover element.',
+          args: [
+            {
+              arg: 'value',
+              type: 'PointerEvent',
+              description: 'Native pointer event',
+            },
+          ],
         },
```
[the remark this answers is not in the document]
dwgray: > Not sure we need to document the native events?

I'm not generally documenting native events if they haven't been defined in the component (via defineEmits) - in [this case it is](https://github.com/dwgray/bootstrap-vue-next/blob/b65c1e4496e636e889b97e42c27e28dfb4231e45/packages/bootstrap-vue-next/src/types/ComponentEmits.ts#L36)

I'm unclear as to when declaring DOM events that bubble up through the component (which is what I think is happening here) is recommended. According to the [vue.js docs](https://github.com/dwgray/bootstrap-vue-next/blob/b65c1e4496e636e889b97e42c27e28dfb4231e45/packages/bootstrap-vue-next/src/types/ComponentEmits.ts#L36):

> Although optional, it is recommended to define all emitted events in order to better document how a component should work. It also allows Vue to exclude known listeners from [fallthrough attributes](https://vuejs.org/guide/components/attrs#v-on-listener-inheritance), avoiding edge cases caused by DOM events manually dispatched by 3rd party code.

Other examples I see of explicitly declaring DOM events in defineEmits are cases where we're capturing the event and re-emitting them - BButton click, BFormInput blur. In the first case, we include it in the component documentation; in the latter, we don't.

I don't have a preference either way, I'm guessing it would be easier for a naive user to see everything that a component can emit in one place. If we're declaring emits for native events to get full type checking, then it should be reasonably easy to also add them to the documentation (although I'm not sure I would do a full consistency pass on that until I'm done with my first parity pass). Alternately, we could link to the MDN page for the native control, also something for later.

@VividLemon, any thoughts on this?

apps/docs/src/docs/components/demo/PopoverTriggers.vue
dwgray: I'm using prettier to format - I think I'm using the settings suggested by the project. I can look at overriding the line length in prettier for the demo directories (I think that's probably possible) - Do you have a suggestion on line length for the samples?

apps/docs/src/utils/showhide-emits.ts
dwgray: Omitting `toggle` and `toggle-prevented` was an oversight on my part. Do you mean these can only be triggered by exposed functions /composables? Presumably, the event can be seen by the popover's container in the normal ways...

apps/docs/src/data/components/popover.data.ts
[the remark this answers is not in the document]
xvaara: Popover doesn't use them, but tooltip emits:

https://github.com/bootstrap-vue-next/bootstrap-vue-next/blob/790679d21ba648664e114ddd93907ffb5c7ba961/packages/bootstrap-vue-next/src/components/BTooltip/BTooltip.vue#L15-L16

But I don't see a reason for that, you can remove them. Since the event is passed through to the BPopover anyway. And I don't even know what it means in this context. We should avoid it as much as we can, and instead bind the attrs to the right element.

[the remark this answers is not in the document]
dwgray: > Remove them and they are native.

That doesn't work. If I remove them, I get this warning `hook.js:608  [Vue warn]: Component emitted event "blur" but it is neither declared in the emits option nor as an "onBlur" prop.`

Looking at the code more closely, I believe what is going on is that BPopover is adding event listeners for these events in the bind function. In the case of `pointerleave` that binds the event to `tryHide` which (conditionally) calls `hide` from `useShowHide` which builds the `BvTriggerableEvent` object and emits the `pointerleave` event with that obect as its argument.

Furthermore, the `blur` event is against the target while the `pointerleave` event is against the popover. So these are initiated by native events, but they aren't the native events, and I don't think they'll bubble up.

All that said, you've got a lot more experience in this code base, and you wrote this code, so I wouldn't be surprised if I'm missing something. But in that case, I'm out of my depth, so I don't know how to do what you're asking.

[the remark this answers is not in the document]
xvaara: Ah, wrote a long reply and pressed resolve conversation and it got lost.

But you are correct. I didn't remember we were emitting a event with the native events type. Just looking at the tryHide gives me a headache since it was so hard to try to figure the correct way of checking when to hide.

Both of those events trigger when leaving the trigger element, but if the focus/pointer goes to the popover element they are looped until another hide event comes along, or the focus/pointer leaves the popover/trigger combo. This only happens if the popover is interactive (by default BPopover is and BTooltip isn't)

dwgray:
> I generally would prefer that we'd use the target slot so we don't need to have id/target combo in most examples.

Okay, did that. Let me know if there is anything else that you feel is blocking this (I am having a little difficulty understanding the github UI around this).

apps/docs/src/docs/migration-guide.md
```diff
@@ -594,13 +595,41 @@ See [Show and Hide](#show-and-hide) shared properties.
...
+
+The `variant` prop has been deprecated. Use Bootstrap’s color and background utility classes to style
+popovers instead. See [Popover custom classes and variants](/docs/components/popover#custom-classes-and-variants)
+for details.
+
+The `disabled` prop and
+[Programmatically Disabling](https://bootstrap-vue.org/docs/components/popover#programmatically-disabling-popover) have
+been deprecated along with the `disabled` and `enabled` events. Use `manual=true` to disable BootstrapVueNext’s automatic
+trigger handling and if your own code shows the popover disable those mechanisms as well. If you believe that implementing
+full parity with the BootstrapVue feature is useful, please open an issue or propose a pull request.
```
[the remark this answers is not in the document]
xvaara: we are referring to the old version in the migration guide.

apps/docs/src/docs/components/demo/PopoverStyles.vue
dwgray: > true doesn't work here. It need to be variable.

@xvaara I don't think the model-value passed in for something that uses vue's defineModel has to be a variable. I couldn't tell one way or the other from the docs, but this seems conclusive: https://play.vuejs.org/#eNqNUk1PwzAM/StRTiCN9QCnqqsEaAeQ+NA2brmU1i0ZaVLlo0yq+t9xUla6MU2cWtvv2c/O6+ht08xbBzSmick1bywxYF2TMsnrRmlLOg1lT0qtasIoIhkdS2v8CnhSBYg9YB5Nkr6xhzOZK2ksqQNyQbDjhdUOLplMomEqzsPAAnIzCxgRkhS8Td9kASWXUAzkmCST/lGaRB40ojfY9BSQxCF51WbCwYJRP5zRP/QVlP9hh2hKT6KJcDqj1uC+Ja/mW6MkXrbzExjNVd1wAfqlsRzvwWhMQsXXMiHU12PIeXGzfT7/gPzzRH5rdj7H6KsGA7rFdcaazXQFdigv18+ww/+xiOKdQPSZ4gqMEs5rHGB3ThYoe4ILah+CCbisNma5syDNfqlwXUT2AR9Mc39m9V+51/ObwGOyxyseGenYoURksvJvid3w7IcOG0wTuMm7UgIymV50mM2csDEpM2GgP2+/rvvp1qOewxfuvwFBphmT

I think the naked `show` prop provides a better dev experience, but there are a bunch of places in the docs where I've been using `:model-value="true"` pattern (`BAlert`s scattered through the docs for one) - I think I'll eventually convert them to `show` but it's a higher priority if they're actually broken. Also, I may update the migration guide if both techniques work, there may be cases where the `:model-value="true"` makes more sense.

PR #2713, opened 2025-05-27
fix(BTableLite): field attribute isRowHeader now being respected

rgeerts:
# Describe the PR

According to documentation at https://bootstrap-vue-next.github.io/bootstrap-vue-next/docs/components/table.html#field-definition-reference
and to keep inline with the original "bootstrap-vue" implementation. The "isRowHeader" should a allow a td to be a th for if the field definition indicates a true value.

## Small replication
See below that the "first_name" field should be a th instead of a td for all the rows.
```
<template>
  <div>
    <BTable striped hover :items="items" :fields="fields" />
  </div>
</template>

<script setup lang="ts">
const fields = [{key:'first_name', isRowHeader: true}, 'last_name', 'age']
const items = [
  {isActive: true, age: 40, first_name: 'Dickerson', last_name: 'Macdonald'},
  {isActive: false, age: 21, first_name: 'Larsen', last_name: 'Shaw'},
  {isActive: false, age: 89, first_name: 'Geneva', last_name: 'Wilson'},
  {isActive: true, age: 38, first_name: 'Jami', last_name: 'Carney'},
]
</script>
```

## PR checklist

- [x ] Bugfix :bug: - `fix(...)`
- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

- [ x] Pull request title and all commits follow the [**Conventional Commits**](https://www.conventionalcommits.org/) convention or has an [**override**](https://github.com/googleapis/release-please#how-can-i-fix-release-notes) in this pull request body **This is very important, as the `CHANGELOG` is generated from these messages, and determines the next version type. Pull requests that do not follow conventional commits or do not have an override will be denied**

## Summary by CodeRabbit

- **New Features**
  - Table cells now dynamically render as header or data cells based on field properties, enhancing table accessibility and customization.
- **Tests**
  - Added tests to ensure correct rendering of table cells as header or data cells according to field settings.


packages/bootstrap-vue-next/src/components/BTable/table-lite.spec.ts
```diff
@@ -528,4 +528,29 @@ describe('btablelite', () => {
...
+      })
+      const $tbody = wrapper.get('tbody')
+      const $tr = $tbody.findAll('tr')
+      const $th = $tr[0].findAll('th')
+      expect($th.length).toBe(1)
+      expect($th[0].text()).toContain('Dickerson')
+      const $td = $tr[0].findAll('td')
+      expect($td.length).toBe(2)
+      expect($td[0].text()).toBe('40')
+      expect($td[1].text()).toBe('true')
```
VividLemon: This test doesn't really assert that isRowHeader changes the value we want. It would be more useful to check and ensure that the th becomes a td, likely by applying an id to the element, and checking the element that matches that id exists/not exists

packages/bootstrap-vue-next/src/components/BTable/BTableLite.vue
```diff
@@ -85,7 +85,8 @@
             @mouseleave="!filterEvent($event) && emit('row-unhovered', item, itemIndex, $event)"
             @mousedown="handleMiddleClick(item, itemIndex, $event)"
           >
-            <BTd
+            <Component
```
VividLemon: This should be lowercase, `<component`

packages/bootstrap-vue-next/src/components/BTable/table-lite.spec.ts
rgeerts: updated. used a class instead of an id, but I think it accomplishes what your asking for here.

packages/bootstrap-vue-next/src/components/BTable/BTableLite.vue
rgeerts: done

packages/bootstrap-vue-next/src/components/BTable/table-lite.spec.ts
```diff
@@ -528,4 +528,25 @@ describe('btablelite', () => {
...
+      ]
+      const wrapper = mount(BTableLite, {
+        props: {
+          items,
+          fields,
+        },
+      })
+      const $tbody = wrapper.get('tbody')
+      const $tr = $tbody.find('tr')
+      expect($tr.find('th.first-name').exists()).toBe(true)
```
VividLemon, line 547:
```suggestion
      expect($tr.find('th.first-name').exists()).toBe(true)
      await wrapper.setProps({ fields: [
        {key: 'name.first', label: 'Actions', isRowHeader: false, class: 'first-name'},
        {key: 'age', label: 'Age', isRowHeader: false, class: 'age'},
        {key: 'isActive', label: 'Active', isRowHeader: false, class: 'active'},
      ]})
      expect($tr.find('th.first-name').exists()).toBe(false)
```

Something like that. We are wanting to assert that it modifies the expected component

VividLemon (changes requested):

packages/bootstrap-vue-next/src/components/BTable/table-lite.spec.ts
rgeerts: thanks for the suggestion, updated!

VividLemon (approved):

PR #2732, opened 2025-06-06
draft of BApp

xvaara:
BEGIN_COMMIT_OVERRIDE
feat!: new `BApp` component as the recommended way to configure bootstrap-vue-next applications, replacing the plugin-based approach. 
feat!: Merged the Orchestrator components into one.
feat!: renamed the controller composables to `useToast`, `useModal`, and `usePopover`
feat!: combined the old plugins to match the new BApp and composables.
feat!: the old `useModal` was removed since `useToggle` does all what it did.

Release-As: 0.40.0
END_COMMIT_OVERRIDE

## Summary by CodeRabbit

* New Features
  * Introduced BApp as the app root with global defaults, RTL, and built‑in orchestration.
  * Added BOrchestrator to centralize modals, toasts, and popovers.
  * New composables: useModal, useToast, usePopover with promise-based create() APIs.

* Refactor
  * Replaced controller-based APIs and legacy orchestrators with unified registries and keys.
  * Simplified setup (BApp recommended; registry/orchestrator plugins available for legacy).

* Documentation
  * New/updated pages for App, Orchestrator, and composables.
  * Revised setup, migration guide, and examples; navigation updated.

* Tests
  * Added comprehensive BApp test suite.

* Chores
  * Updated demos (playground/Nuxt) and formatted changelogs.


dwgray:
@xvaara, I really like this for the core setup. It's what I would use in my app, given a choice between this and the current plugin.

The main question I have is about tree shaking. As @VividLemon mentioned previously, there is some segment of our user base that is quite concerned about the deployed size of BSVN. If we eventually deprecate the plugins and only allow this method of installation, do we effectively disable tree shaking?

VividLemon:
> @xvaara, I really like this for the core setup. It's what I would use in my app, given a choice between this and the current plugin.
> 
> The main question I have is about tree shaking. As @VividLemon mentioned previously, there is some segment of our user base that is quite concerned about the deployed size of BSVN. If we eventually deprecate the plugins and only allow this method of installation, do we effectively disable tree shaking?

I think the best solution to this is to effectively give people the tools to build their own "bapp" and document how to do it well

packages/bootstrap-vue-next/src/components/BApp/BApp.vue
```diff
@@ -0,0 +1,256 @@
...
+const props = withDefaults(defineProps<BAppProps>(), {
+  noModals: false,
+  noToasts: false,
+  noPopovers: false,
+  inherit: false,
+  appendToast: false,
+  teleportTo: undefined,
+  defaults: undefined,
+  mergeDefaults: false,
+  deepMerge: false,
```
VividLemon: Maybe mergeDefaults is `boolean | function` to where the function is the algorithm to use for merging?

VividLemon: I generally dislike props that just effect the behavior of other props. Begs the question why it's not just an object

xvaara:
> The main question I have is about tree shaking. As @VividLemon mentioned previously, there is some segment of our user base that is quite concerned about the deployed size of BSVN. If we eventually deprecate the plugins and only allow this method of installation, do we effectively disable tree shaking?

@dwgray BApp or the orchestrators imports nothing complex, just keys and conditionalTeleport, so if the user doesn't import them or the composables there shouldn't be any extra components included.

dwgray:
> > The main question I have is about tree shaking. As @VividLemon mentioned previously, there is some segment of our user base that is quite concerned about the deployed size of BSVN. If we eventually deprecate the plugins and only allow this method of installation, do we effectively disable tree shaking?
> 
> @dwgray BApp or the orchestrators imports nothing complex, just keys and conditionalTeleport, so if the user doesn't import them or the composables there shouldn't be any extra components included.

@xvaara I've been thinking about this some more. If this doesn't significantly affect tree shaking, I am still a bit torn.

I'd very much prefer it if we land on one solution or the other, rather than maintaining both. @VividLemon 's suggestion of "give people the tools to build their own 'bapp' and document how to do it well" sounds appealing as a solution to only maintaining one implementation, but that really ends up being what I consider a `scaffolding` solution in one way or the other and moves the burden of managing changes over time on the user of the library (I wrote a [Medium article](https://medium.com/@dwgray/maintaining-scaffolded-code-case-study-with-asp-net-core-identity-ui-1ce8b6af5ebb) on this a while back - just to show how frustrating I find it).

Advantages of BApp:
- Easier initial use
- More flexibility for us to modify how things work without forcing users to migrate/react to breaking changes (see above)
- This looks like what Vuetify.js does, and I've found their decisions have generally been solid when I've dug into them

Advantages of Plugin:
- Less churn for existing users
- More granular control/Less abstraction
- Feels less notional - enforcing the use of a top-level app moves us from a set of components that one can use to build a vue.js app, to a framework where we provide the top-level app and the developer fills in their details - this may just be a historical quirk from someone who's dealt with way too many variations on this issue over the years but I lean towards being less notional where we can.

Overall, that does land me on seeing if we can make BApp work, again assuming it doesn't affect tree shaking. I started digging into the mechanics of tree-shaking to convince myself that we would be safe, and quickly realized that this is one of those problems where I wouldn't be able to truly convince myself of the results without experimental evidence. I think it's probably good for us to be able to regression test tree shaking overall, since some part of our user base really cares about this (and others may care more if they dug into it).

My basic idea around this is to have a small project that we can build and establish baseline sizes to confirm that we've not changed the pay-to-play nature of the tree shaking. We should then be able to just change the target bsvn installation in the package.json file and rebuild to see how sizes are affected.

Here's the stats for a recent build without the BApp changes:

## Simple Vue.js/Vite app using create vue@latest
```
vite v6.2.4 building for production...
✓ 26 modules transformed.
dist/index.html                  0.43 kB │ gzip:  0.28 kB
dist/assets/index-TXLObeWa.css   3.71 kB │ gzip:  1.19 kB
dist/assets/index-usgfJ6g8.js   69.64 kB │ gzip: 27.43 kB
✓ built in 865ms
```

## Add bootstrap and BSVN
```
✓ 136 modules transformed.
dist/index.html                   0.43 kB │ gzip:  0.29 kB
dist/assets/index-BtuzhIq3.css  254.56 kB │ gzip: 34.92 kB
dist/assets/index-91SiXKPq.js    85.35 kB │ gzip: 31.81 kB
✓ built in 1.82s
```

## Add a reference to BLink
```
vite v6.2.4 building for production...
✓ 136 modules transformed.
dist/index.html                   0.43 kB │ gzip:  0.29 kB
dist/assets/index-BRDhJoMO.css  254.56 kB │ gzip: 34.92 kB
dist/assets/index-CMzBWzqN.js    92.56 kB │ gzip: 34.24 kB
✓ built in 1.50s
```

## Add a reference to BAvatar
```
vite v6.2.4 building for production...
✓ 136 modules transformed.
dist/index.html                   0.43 kB │ gzip:  0.28 kB
dist/assets/index-vRN9tOtg.css  254.56 kB │ gzip: 34.92 kB
dist/assets/index-r0dbNEhh.js   101.61 kB │ gzip: 36.74 kB
✓ built in 1.58s
```

Overall, these seem like reasonable results and show that adding references to new components increases the size of the packet (presumably more than the reference). I also did a bit of digging into the generated files to prove to myself that there weren't references to `BLink` or `BAvatar` in the generated results until I added them in my app.

The place where I stalled out is that I wasn't able to change the reference to BSVN to @xvaara's BApp branch. I think that's because it doesn't build? But it may be because I don't have the syntax right. Here's what I tried:

```
    "bootstrap-vue-next": "bootstrap-vue-next/bootstrap-vue-next#bapp-composables-makeover",
```

I've pushed the project up to https://github.com/[dwgray/bsvn-tree-shake](https://github.com/dwgray/bsvn-tree-shake) - let me know if you think this is a reasonable way to approach confirming our tree-shaking (and that we don't degrade it with features like this) @xvaara  and @VividLemon and if I should add other features to test (I'm thinking `BModal` or something that requires an orchestrator would be a good next step).

P.S. Looks like I accidentally wrote over your comment rather than replying to it - I think I managed to restore the original comment, but that's why it's tagged as edited by me...

xvaara:
@dwgray you can use the pr.new packages now. I've never gotten the direct git+npm to work on this project.

> I think the best solution to this is to effectively give people the tools to build their own "bapp" and document how to do it well

@VividLemon 
So document the default provide?:
```vue
import { defaultsKey, type BvnComponentProps} from 'bootstrap-vue-next'
provide(defaultsKey, ref({BPopover: {bodyClass: 'primary'}}) as Ref<Partial<BvnComponentProps>>)
```

Or what do you mean by tools? Composable that creates the registries?

This was just a quick hack to see how it goes, of course we would create one place to keep the implementation. I even started to merge the registries to one, but didn't finish that yet.

dwgray:
> @dwgray you can use the pr.new packages now. I've never gotten the direct git+npm to work on this project.

Thanks @xvaara! This looks great. I just rebuilt the "final" version of my test app which references just `BLink` and `BAvatar` and the `BApp` version is actually _smaller_.

## BLink & BAvatar BSVN 0.30.3
```
✓ 137 modules transformed.
dist/index.html                   0.43 kB │ gzip:  0.28 kB
dist/assets/index-BatG3cbY.css  255.44 kB │ gzip: 35.16 kB
dist/assets/index-CQTmLpcJ.js   101.59 kB │ gzip: 36.74 kB
✓ built in 1.83s
```
## BLink & BAvatar BSVN 0.30.3 BApp Branch
```
✓ 142 modules transformed.
dist/index.html                   0.43 kB │ gzip:  0.28 kB
dist/assets/index-BatG3cbY.css  255.44 kB │ gzip: 35.16 kB
dist/assets/index-afQ1bg1x.js    94.82 kB │ gzip: 34.96 kB
✓ built in 1.97s
```

I also checked that the code actually ran (I've been bitten more times than I care to mention by being excited about a performance improvement just to find that the code didn't actually work).

So I don't think this is contributes to the BApp vs. Plugin discussion.   I do think it's probably worth keeping an eye on this as we move forward and I'm happpy to take point on that. For that project, I think it's worth tracking something more complext that currently uses the plugin/orchestrator model. Do you or @VividLemon have an opion on that? And if so if there is a best candidate? Maybe `BPopover` or `BModal`?

I may also look into tooling to be able to visualize tree shaking results - I think I saw a reference there, but if either of you has any experience with that, I'd be interested in a link.

I don't think I'm going to take the time to try to automate this (I have a significant amount of experience with automated performance testing, and it's generally been a nightmare), so I'm thinking in terms of periodic sniff testing to check that we haven't done something to totally break tree-shaking.

dwgray:
> Or what do you mean by tools? Composable that creates the registries?

I don't know if this is what @VividLemon was thinking, but I read that as doing something like making it easy to create a project from scratch ala `pnpm create bootstrap-vue-next` or providing templates. That's what lead me down the ramble about scaffolding code. Based on the work that I recently did with the group of students, I think it might be worthwhile to do something like that in any case - I, at least, have been focused on supporting migrating users, but it is probably worth smoothing the path for new adopters as we get closer to a v1...

VividLemon:
Or what do you mean by tools? Composable that creates the registries?

Yes, simply this and a small demo on how to do it.

VividLemon:
I think there are better things to test than performance. Just keep the bundle size low and we should be ok

packages/bootstrap-vue-next/src/components/BApp/BApp.vue
```diff
@@ -0,0 +1,256 @@
...
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    registry: readonly(registry) as Readonly<Ref<Map<number, ComponentInternalInstance>>>,
+    stack: valuesStack,
+    pushStack,
+    removeStack,
+    pushRegistry,
+    removeRegistry,
+  })
+}
```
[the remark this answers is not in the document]
xvaara: @VividLemon is this true or not? if it's true it's also a bug in the current implementation

packages/bootstrap-vue-next/src/composables/shared.ts
```diff
@@ -0,0 +1,107 @@
...
+        })
+      }
+      store.value.splice(
+        store.value.findIndex((el) => el._self === _id),
+        1
+      )
+    },
+    async [Symbol.asyncDispose]() {
+      await this.destroy()
+    },
```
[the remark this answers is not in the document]
xvaara: If someting mutates the store.value the index will change, doesn't that mean that we should not save the index before the await?
it already checks the index with _self after the await, or am I mistaken?

packages/bootstrap-vue-next/src/components/BApp/BApp.vue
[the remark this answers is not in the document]
xvaara: @coderabbitai on what do you base this comment that vue ref doesn't track interals of new Map()?

the documentation says:
Refs can hold any value type, including deeply nested objects, arrays, or JavaScript built-in data structures like Map.

A ref will make its value deeply reactive.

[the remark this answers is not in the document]
xvaara: @coderabbitai where in vue documentation it says ref with a Map isn't reactive? And what search results are you basing this on? I can't find any.

xvaara:
@coderabbitai full review

xvaara:
stackblitz for testing:

old plugin and BOrchestrator:
https://stackblitz.com/edit/github-nkwgqa5r?file=src%2Fcomponents%2FComp.vue,src%2FApp.vue,src%2Fmain.ts

BApp with plugin:
https://stackblitz.com/edit/github-nkwgqa5r-m13ubiby?file=src%2Fcomponents%2FComp.vue,src%2FApp.vue,src%2Fmain.ts

BApp: 
https://stackblitz.com/edit/github-nkwgqa5r-ogfk1jwn?file=src%2Fcomponents%2FComp.vue,src%2FApp.vue,src%2Fmain.ts

BOrchestrator and composables in App.vue:
https://stackblitz.com/edit/github-nkwgqa5r-bsp2uwre?file=src%2Fcomponents%2FComp.vue,src%2FApp.vue,src%2Fmain.ts

packages/bootstrap-vue-next/src/components/BApp/BApp.vue
```diff
@@ -0,0 +1,35 @@
...
+  mergeDefaults: false,
+  noOrchestrator: false,
+  rtl: false,
+})
+
+useProvideDefaults(props.defaults, props.mergeDefaults)
+useRoot(props.rtl)
+if (!props.noOrchestrator) {
+  useOrchestratorRegistry(props.inherit)
+}
```
xvaara: @VividLemon @dwgray I'm not sure how much we want to separate the registries. 

My reasoning for this three ways is:
- defaults, you might want to override them in child components
- Orchestator, you might want to have seperate registries on some branch (not sure about this)
- useRoot (showHidestorage, modalmanager, breadcrumRegistry and rtlRegistry) are all global functionalities

VividLemon:
I apologize to make this point now, kind of changing my mind, but simply having a registry point that contains a few refs is nothing to be concerned about in terms of performance.

The major issue with the old size of the plugins was the scale of the business logic in them. If this logic is moved to composables, then by simply not using the composable, it is not included in the bundle (unlike the plugins).

So if the "registries" are just some .provide() of refs, then we could just be best off forcing it and making no concessions for their removal (beyond not using bapp). 

Does anyone have any thoughts on this? I feel pretty bad upon thinking about this so late.

packages/bootstrap-vue-next/src/components/BApp/BApp.vue
VividLemon: Does my latest comment, if we force them to use the registries always, fix this issue at all? 

Otherwise it's likely fine to keep those three items always bound together

`useOrchestratorRegistry(props.inherit)` `props.inherit` etc isn't reactive. I haven't looked at the source code or anything, but considering it's `use_` (the naming for a composable) then it seems like it should have a reactive input?

xvaara:
> I apologize to make this point now, kind of changing my mind, but simply having a registry point that contains a few refs is nothing to be concerned about in terms of performance.
> 
> The major issue with the old size of the plugins was the scale of the business logic in them. If this logic is moved to composables, then by simply not using the composable, it is not included in the bundle (unlike the plugins).
> 
> So if the "registries" are just some .provide() of refs, then we could just be best off forcing it and making no concessions for their removal (beyond not using bapp).
> 
> Does anyone have any thoughts on this? I feel pretty bad upon thinking about this so late.

They are only a few provides, yes. All the logic is in the composables now anyways. So I also see no reason for the user the be able to "disable" them. 

Do we want to continue supporting the plugin "way" of createBootstrap()? I don't know a way to use composables directly in them since you have to use the passed app prop for provide. In this PR i'm pulling in a function from the composable that I can just use in app.provide.

packages/bootstrap-vue-next/src/components/BApp/BApp.vue
xvaara: @VividLemon I'm not sure anymore how the inherit stuff should go, we need to think about it a little more. I'm not really sure we need different registries with "inheritance" maybe we need a way to initilalise a new registry and pass that to a new orchestrator, which can be used in a branch of the app?

@dwgray what are your thoughts on the registries?

xvaara: BOrchestrator has a test with a filter function, so you could show some in a different orchestrator... like 
```js
const filter = (i) => i.id.match(/^profile/) && i.type === 'toast'
```
to match profile toasts do a different orchestrator. (and you need to filter them out of the primary orchestrator).

Do we need more than this actually?

dwgray: @xvaara I haven't used regestries before so I don't haveany strong opinions. I spent some time today trying to get my head wrapped around what you're doing here and wasn't fully successful.

I wonder if you could expand a little on the three points above from the perspective of use cases? Since you're using vue's provide/inject, for the defaults managed by useProvideDefaults, if some child component of BApp calls useProvideDefaults with different values, all _its_ children will use the new defaults (or merged defaults) - that seems useful, but fairly corner case - like maybe the settings page of an app has different defaults for some of its components than the main App?

Is the same thing true of what you're doing with useOrchestratorRegistry? From my reading, the registry sets up the props for the different components that can be ochestrated (Popover, Modal, etc). I'm assuming these defaults override the defaults set using useProviderDefaults? But I get lost in where you would want to use one mechanism rather than the other.

I guess the other thing that you're doing with the filter is allowing different Orchestrators for the same type of orchestrated compnent based on some attribute of the component (in your example, that would be name & type match, but presumably you could have some arbitrary other field (like 'area') to filter on. But I'm struggly to see what that buys us.

xvaara: I removed the inherit all together from orchestratorRegistry. Didn't wrap my head around how it would be very useful, only edge cases came to mind.

@dwgray they get the new defaults or the merged ones, depending on the options for that. useOrchestratorRegistry doesn't set any defaults, just initialises the registries that the composables use.

My idea for the filter was that you can have multible ochestrators wrapped in different defaults, so maybe filter out error toasts and center them and make them red. Not 100% sure on the idea.

dwgray:
> I think there are better things to test than performance. Just keep the bundle size low and we should be ok

Yep - I just want to be able to notice if we do something that pulls in all of BSVN (or a large part of it) for doing something like consuming `BButton`.

apps/docs/.vitepress/theme/Layout.vue
```diff
@@ -1,161 +1,166 @@
...
-        <BNav>
-          <BNavItem
-            v-for="link in headerLinks"
-            :key="link.route"
-            :to="link.route"
-            :active="route.path === `${link.route}.html`"
-            class="py-2"
-            active-class="active fw-bold"
-            >{{ link.label }}</BNavItem
+  <BApp>
```
dwgray, line 3: I'll merge #2802 once this goes in.

apps/docs/src/data/components/app.data.ts
```diff
@@ -0,0 +1,57 @@
...
+          },
+          rtl: {
+            type: 'boolean | { rtlInitial?: boolean; localeInitial?: string }',
+            default: false,
+            description:
+              'If an object, it can specify initial RTL state and locale. If `false`, disables RTL mode.',
+          },
+        } satisfies Record<keyof BvnComponentProps['BApp'], PropertyReference>,
+      },
+      emits: [],
```
dwgray: This will conflict with #2806

apps/docs/src/docs.md
```diff
@@ -96,6 +139,34 @@ app.mount('#app')
...
+- **Easier Configuration**: Set defaults and options through component props
+- **Automatic Orchestrators**: Built-in support for modals, toasts, and popovers
+- **Modern Vue 3 Patterns**: Uses composition API and provide/inject
+- **Better Tree Shaking**: Only includes what you actually use
+
+See the [BApp component documentation](/docs/components/bapp) for complete configuration options and the [migration guide](/docs/migration-guide#bapp-vs-plugin-setup-recommended) for migration instructions.
+
 Now, you can begin importing and using components
 
 #### Automatic Registering of Components
```
dwgray, line 133: Maybe make this ### instead of #### (it's weird that it's hidden under `Why BApp is Recommended` - ither that or move this section up before the previous two section - which might be a good idea anyway - for someone that is approaching BSVN for the first time the legacy and comparison sections aren't all that important...

apps/docs/src/docs.md
```diff
@@ -96,6 +139,34 @@ app.mount('#app')
...
+
+The `BApp` component offers several advantages:
+
+- **Better Type Safety**: Full TypeScript support with proper type inference
+- **Easier Configuration**: Set defaults and options through component props
+- **Automatic Orchestrators**: Built-in support for modals, toasts, and popovers
+- **Modern Vue 3 Patterns**: Uses composition API and provide/inject
+- **Better Tree Shaking**: Only includes what you actually use
+
+See the [BApp component documentation](/docs/components/bapp) for complete configuration options and the [migration guide](/docs/migration-guide#bapp-vs-plugin-setup-recommended) for migration instructions.
```
dwgray: I don't think there is such a section in the migration guide. I would assert that the migration guide shouldn't have an app vs. plugin section, but it should have some discussion of creating an app and setting defaults, since that is different than BSV.

dwgray (commented):
The only substantive comments I have are on how to structure the getting started guide and the migration guide. I did pull your code and play with it a bit and it appears to be working as advertised (and is indeed easier to use than the plugin approach) - if I have time I'll try consuming this in my app, but don't block on that. I did not fully get my head wrapped around the implementation, so I think it would be worthwhile for @VividLemon to comment.

dwgray:
@xvaara I only had a few minutes to play with this today, but I did make a little progress.

My initial intention was to run this in "compatibility mode" - e.g. just pull in this PR via setting `"bootstrap-vue-next": "https://pkg.pr.new/bootstrap-vue-next/bootstrap-vue-next@2732",` and see what happens.

The good news is that my app runs fine - it's pretty straightforward I don't explicitly use plugins, but I do use a few toasts, modals, etc. and they all seem to work.

The bad news:
I have a test that explicitly uses plugins (for historical reasons, I don't really need this, but it was nice to have in this case because it failed, and if we're assuming backwards compatibility with plugins, I don't think it should fail):

```ts
import { mount } from "@vue/test-utils";
import { describe, expect, test } from "vitest";
import { MenuContext } from "@/models/MenuContext";
import MainMenu from "../MainMenu.vue";
import { mockResizObserver } from "@/helpers/TestHelpers";
import { modalControllerPlugin, modalManagerPlugin, toastPlugin } from "bootstrap-vue-next";

describe("MainMenu.vue", () => {
  test("Renders MainMenu for an anonymous user", () => {
    mockResizObserver();
    const context = new MenuContext();
    const wrapper = mount(MainMenu, {
      props: { context },
      global: { plugins: [modalControllerPlugin, modalManagerPlugin, toastPlugin] },
    });
    expect(wrapper.html()).toMatchSnapshot();
  });
});
```

Generated errors that BSVN doesn't export these plugins. So is this is an issue with the code, or am I misunderstanding your intent> (which may mean that we need to clarify the documentation).

I'm not blocked on this, if I have a few cycles tomorrow, I'll try doing the actual conversion to `BApp` and see what happens. This isn't a particularly great test case since I'm not a deep user of some of these features, but it may turn over some additional low hanging fruit.

xvaara:
> The bad news: I have a test that explicitly uses plugins (for historical reasons, I don't really need this, but it was nice to have in this case because it failed, and if we're assuming backwards compatibility with plugins, I don't think it should fail):
> 
> ```ts
> import { mount } from "@vue/test-utils";
> import { describe, expect, test } from "vitest";
> import { MenuContext } from "@/models/MenuContext";
> import MainMenu from "../MainMenu.vue";
> import { mockResizObserver } from "@/helpers/TestHelpers";
> import { modalControllerPlugin, modalManagerPlugin, toastPlugin } from "bootstrap-vue-next";
> 
> describe("MainMenu.vue", () => {
>   test("Renders MainMenu for an anonymous user", () => {
>     mockResizObserver();
>     const context = new MenuContext();
>     const wrapper = mount(MainMenu, {
>       props: { context },
>       global: { plugins: [modalControllerPlugin, modalManagerPlugin, toastPlugin] },
>     });
>     expect(wrapper.html()).toMatchSnapshot();
>   });
> });
> ```
> 
> Generated errors that BSVN doesn't export these plugins. So is this is an issue with the code, or am I misunderstanding your intent> (which may mean that we need to clarify the documentation).
> 
> I'm not blocked on this, if I have a few cycles tomorrow, I'll try doing the actual conversion to `BApp` and see what happens. This isn't a particularly great test case since I'm not a deep user of some of these features, but it may turn over some additional low hanging fruit.

There is no plugin for toast or modals anymore, it's true we'd need to document that. 

```ts
export {createBootstrap} from './createBootstrap'
export {registryPlugin} from './registry'
export {rtlPlugin} from './rtl'
export {orchestratorPlugin} from './orchestrator'
```

apps/docs/src/docs.md
xvaara: I think copilot added that in, and I removed the migration guide but forgot to remove the links. Good catch

apps/docs/.vitepress/theme/Layout.vue
xvaara: Well it seams @VividLemon already merged that in.

packages/bootstrap-vue-next/CHANGELOG.md
```diff
@@ -2,112 +2,100 @@
...
+- **BModal:** fix ssr. ([df4a90d](https://github.com/bootstrap-vue-next/bootstrap-vue-next/commit/df4a90dacbcc99ee86b03499eb53501e240eb0ba))
+- **BTableLite:** field attribute isRowHeader now being respected ([#2713](https://github.com/bootstrap-vue-next/bootstrap-vue-next/issues/2713)) ([9a4b648](https://github.com/bootstrap-vue-next/bootstrap-vue-next/commit/9a4b648a31678ab1ea99f8c81b556ed166ae3fb3))
+- **BTabs:** Add type=button to buttons to prevent form submission ([#2740](https://github.com/bootstrap-vue-next/bootstrap-vue-next/issues/2740)) ([1a72f1b](https://github.com/bootstrap-vue-next/bootstrap-vue-next/commit/1a72f1b50494512b692b53f968a40222e66edd40))
+- **BToast:** close BToast correctly if modelValue is changed from number to false ([df4a90d](https://github.com/bootstrap-vue-next/bootstrap-vue-next/commit/df4a90dacbcc99ee86b03499eb53501e240eb0ba))
+- **BToast:** close BToast correctly if modelValue is changed from number to false ([#2745](https://github.com/bootstrap-vue-next/bootstrap-vue-next/issues/2745)) ([df4a90d](https://github.com/bootstrap-vue-next/bootstrap-vue-next/commit/df4a90dacbcc99ee86b03499eb53501e240eb0ba))
```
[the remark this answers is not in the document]
xvaara: Ignore CHANGELOG.md since it's autogenerated.

packages/bootstrap-vue-next/src/components/BApp/BOrchestrator.vue
```diff
@@ -0,0 +1,161 @@
...
+              options,
+              _component,
+              ...val
+            } in items?.filter((el) => el.position === key) || []"
+            :key="_self"
+          >
+            <component
+              :is="_component"
+              v-bind="val"
+              :ref="(ref: ComponentPublicInstance) => (promise.value.ref = ref)"
```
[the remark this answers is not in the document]
xvaara: Where did you get this idea?
Vue documentation says: The <template> also supports TypeScript in binding expressions when <script lang="ts"> or <script setup lang="ts"> is used.

apps/docs/src/docs/composables/usePopover.md
```diff
@@ -118,39 +123,41 @@
...
 
 ```js
-await using popover = popover({title: 'Hello World!'})
+await using pop = popover({title: 'Hello World!'})
 ```
 
 <script setup lang="ts">
-import {usePopoverController} from 'bootstrap-vue-next/composables/usePopoverController'
+import {usePopover} from 'bootstrap-vue-next/composables/usePopover'
 import HighlightCard from '../../components/HighlightCard.vue'
```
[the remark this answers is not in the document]
xvaara: the package.json has an export for ./composables/* isn't that enough? It works and builds.

VividLemon (approved):

packages/bootstrap-vue-next/src/composables/useRegistry/index.ts
```diff
@@ -0,0 +1,139 @@
...
+
+    const rtlInitial = typeof rtl === 'boolean' ? rtlDefault : (rtl?.rtlInitial ?? rtlDefault)
+
+    const localeInitial =
+      typeof rtl === 'boolean' ? localeDefault : (rtl?.localeInitial ?? localeDefault)
+
+    const isRtl = ref(rtlInitial)
+    const locale = ref(localeInitial)
+
+    provide(rtlRegistryKey, {isRtl, locale})
```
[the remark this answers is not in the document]
VividLemon: This does seem like an issue tho

PR #2741, opened 2025-06-09
docs: WIP create a more strongly typed schema for emits

VividLemon:
@xvaara I'm curious on what you think about the way I handled passthrough of emits to btablesimple here. Do you think it should be applied for more components that inherit emits from others? It's a bit bulky to convert from it's emit, to include 'on-' then to camelize it (all in types), then pass through as a v-bind. But it provides additional type safety in adding new emits in sub components. 

@dwgray I'm also curious what you think in the benefit shown to the documentation. It allows you to get the name of the emit, but provides no additional benefit for the arguments in that emit (as they are a tuple and can't get the named values in a tuple). It's also pretty unlike the already made 'BvnComponentEmits', as it just does `export type * as BvnComponentEmits`. Ideally the schema would be like `{ [componentName]: { props: {}, emits: {} }` to which you could do `BvnComponent['BAccordion']['emit']`, to keep everything all tidy together, but this is also quite bulky to do

Does this seem worthwhile? Should slots follow in addition for their inheritance? Originally the thought process was to stay away from inheritance, but it seems like these three could benefit from it.

### DIFF
--- a/apps/docs/src/data/components/popover.data.ts
+++ b/apps/docs/src/data/components/popover.data.ts
@@ -29,9 +29,18 @@ export default {
           },
           click: {
             type: 'boolean',
-            default: false,
-            description:
-              'show/hide the popover on click of the target element (default is hover/focus)',
+            default: undefined,
+            description: 'Enable/disable click trigger. See [Triggers](#triggers) for details.',
+          },
+          focus: {
+            type: 'boolean',
+            default: undefined,
+            description: 'Enable/disable focus trigger. See [Triggers](#triggers) for details.',
+          },
+          hover: {
+            type: 'boolean',
+            default: undefined,
+            description: 'Enable/disable hover trigger. See [Triggers](#triggers) for details.',
           },
           delay: {
             type: 'number | { show: number; hide: number }',
--- a/apps/docs/src/docs/components/demo/PopoverTriggers.vue
+++ b/apps/docs/src/docs/components/demo/PopoverTriggers.vue
@@ -1,23 +1,71 @@
 <template>
   <!-- #region template -->
-  <div class="d-flex gap-2">
-    <BPopover
-      placement="top"
-      body="Default Trigger"
-    >
+  <div class="d-flex flex-wrap gap-2">
+    <!-- Default: Hover + Focus -->
+    <BPopover>
       <template #target>
-        <BButton> Default Trigger </BButton>
+        <BButton variant="secondary">Default (hover/focus)</BButton>
       </template>
+      Default triggers: hover and focus
     </BPopover>
+
+    <!-- Click only -->
+    <BPopover click>
+      <template #target>
+        <BButton variant="primary">Click Only</BButton>
+      </template>
+      Click to toggle
+    </BPopover>
+
+    <!-- Hover only -->
+    <BPopover hover>
+      <template #target>
+        <BButton variant="success">Hover Only</BButton>
+      </template>
+      Hover to show (no focus)
+    </BPopover>
+
+    <!-- Focus only -->
+    <BPopover focus>
+      <template #target>
+        <BButton variant="info">Focus Only</BButton>
+      </template>
+      Focus to show (no hover)
+    </BPopover>
+
+    <!-- Multiple triggers -->
     <BPopover
       click
-      placement="top"
-      body="Click Trigger"
+      hover
+      focus
     >
       <template #target>
-        <BButton> Click Trigger </BButton>
+        <BButton variant="warning">All Triggers</BButton>
       </template>
+      Click, hover, or focus
+    </BPopover>
+
+    <!-- Manual control -->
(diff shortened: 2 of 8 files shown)


dwgray:
> @dwgray I'm also curious what you think in the benefit shown to the documentation. It allows you to get the name of the emit, but provides no additional benefit for the arguments in that emit (as they are a tuple and can't get the named values in a tuple). It's also pretty unlike the already made 'BvnComponentEmits', as it just does `export type * as BvnComponentEmits`. Ideally the schema would be like `{ [componentName]: { props: {}, emits: {} }` to which you could do `BvnComponent['BAccordion']['emit']`, to keep everything all tidy together, but this is also quite bulky to do
> 
> Does this seem worthwhile? Should slots follow in addition for their inheritance? Originally the thought process was to stay away from inheritance, but it seems like these three could benefit from it.

Let me try to paraphrase to make sure I'm understanding. Should we do the same thing with emits and slots as you did with props, where the component data in the docs is an object that is type-checked against the actual component's type rather than an array, even if we can't type check the args? I'd say yes.

I'm curious by what you mean by bulky? Adding the typing to each emeit section in the docs? I'd definitely take that for the benefit of keeping the docs more in sync with the source.

VividLemon:
> I'm curious by what you mean by bulky? Adding the typing to each emeit section in the docs? I'd definitely take that for the benefit of keeping the docs more in sync with the source.

This:

```ts
const boundBTableLiteEmits = {
  onHeadClicked: (fieldKey, field, event, isFooter = false) => {
    emit('head-clicked', fieldKey, field, event, isFooter)
    handleFieldSorting(field)
  },
  onRowClicked: (row, index, e) => {
    if (props.noSelectOnClick === false) {
      handleRowSelection(row, index, e.shiftKey, e.ctrlKey, e.metaKey)
    }
    emit('row-clicked', row, index, e)
  },
  onRowDblclicked: (...args) => emit('row-dblclicked', ...args),
  onRowContextmenu: (...args) => emit('row-contextmenu', ...args),
  onRowHovered: (...args) => emit('row-hovered', ...args),
  onRowUnhovered: (...args) => emit('row-unhovered', ...args),
  onRowMiddleClicked: (...args) => emit('row-middle-clicked', ...args),
} as const satisfies {
  [K in keyof BTableLiteEmits<Items, FieldsType> as CamelCase<`on-${K & string}`>]: (
    ...args: BTableLiteEmits<Items, FieldsType>[K]
  ) => void
}
```

dwgray:
> > I'm curious by what you mean by bulky? Adding the typing to each emeit section in the docs? I'd definitely take that for the benefit of keeping the docs more in sync with the source.
> 
> This:
> 
> ```ts
> const boundBTableLiteEmits = {
>   onHeadClicked: (fieldKey, field, event, isFooter = false) => {
>     emit('head-clicked', fieldKey, field, event, isFooter)
>     handleFieldSorting(field)
>   },
>   onRowClicked: (row, index, e) => {
>     if (props.noSelectOnClick === false) {
>       handleRowSelection(row, index, e.shiftKey, e.ctrlKey, e.metaKey)
>     }
>     emit('row-clicked', row, index, e)
>   },
>   onRowDblclicked: (...args) => emit('row-dblclicked', ...args),
>   onRowContextmenu: (...args) => emit('row-contextmenu', ...args),
>   onRowHovered: (...args) => emit('row-hovered', ...args),
>   onRowUnhovered: (...args) => emit('row-unhovered', ...args),
>   onRowMiddleClicked: (...args) => emit('row-middle-clicked', ...args),
> } as const satisfies {
>   [K in keyof BTableLiteEmits<Items, FieldsType> as CamelCase<`on-${K & string}`>]: (
>     ...args: BTableLiteEmits<Items, FieldsType>[K]
>   ) => void
> }
> ```

I see it does feel a bit bulky @ 20ish lines of code. But doesn't that code replace lines 5-32 (28 lines)+ the definitions of onRowClick (7 lines) and onFieldHeadClick (10 lines)  = 45 lines? Seems like an improvement to me.

xvaara:
@VividLemon This seems fine. Passing the emits is annoying anyway, so I'm for more type safety.

apps/docs/src/components/NoteAlert.vue
```diff
@@ -5,7 +5,7 @@
 </template>
 
 <script setup lang="ts">
-import {BAlert} from 'bootstrap-vue-next'
+import {BAlert} from 'bootstrap-vue-next/components/BAlert'
```
dwgray: Since Unplugin is enabled for docs, can't you just delete the explicit reference to components in cases like this - everywhere except for where we need the explicit reference to access exposed methods?

VividLemon:
List of things to do:

- sync main for changes
- fill in for items that were identified as missing
- Grammar check descriptions in data.ts files

(Use an AI model to grammar check?)

PR #2771, opened 2025-07-08
docs(migration): add component aliases guidelines

homersimpsons:
# Describe the PR

Add section in migration guide about removal of component aliases.

Also fix "the the" typos (I can extract this in another pull request if necessary).

Note that I built (manually) the list of aliases from BootstrapVue's documentation, but maybe some of them do not exist on BootstrapVueNext.

## Small replication

NA

## PR checklist

- [-] Bugfix :bug: - `fix(...)`
- [-] Feature - `feat(...)`
- [-] ARIA accessibility - `fix(...)`
- [-] Other (please describe)

## Summary by CodeRabbit

* **Documentation**
  * Fixed typographical errors and improved clarity in component descriptions and documentation.
  * Updated the migration guide with a new section on component aliases, including a comprehensive comparison table between BootstrapVue and BootstrapVueNext.
  * Enhanced overall readability and accuracy of documentation content.

### DIFF
--- a/apps/docs/src/docs/migration-guide.md
+++ b/apps/docs/src/docs/migration-guide.md
@@ -49,9 +49,9 @@ For any deprecated feature, especially the last case listed above, please feel f
 ## Sync modifier
 
 A number of components in `bootstrap-vue` use `v-bind`'s `.sync` modifier. This modifier has been replaced by properties
-on the the model (generally named models).
+on the model (generally named models).
 
-For instance, in order to two-way bind to the `indeterminate` property in `BFormCheckBox` you `v-bind` to the the model
+For instance, in order to two-way bind to the `indeterminate` property in `BFormCheckBox` you `v-bind` to the model
 named `indeterminate` rather than adding the sync modifier to the `indeterminate` property:
 
 <<< FRAGMENT ./demo/SyncBefore.vue#template{vue-html}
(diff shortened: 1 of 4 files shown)


VividLemon:
Perhaps we should just export them instead of discontinuing them? What do you think

homersimpsons:
> Perhaps we should just export them instead of discontinuing them? What do you think

In my opinion we should not, not having them means:
- reduced api surface
- one way to use the component
- avoid requests to add an alias for this or that component

But having them also has advantages:
- shorter component names
- easier upgrade from BootstrapVue

We could also argue that anyone can declare those component aliases themselves if they need to.

VividLemon:
But we do support binput as an alias. And I think one other. Thats the only one though

VividLemon:
But I do agree. One unified api is my prefered

homersimpsons:
> But we do support binput as an alias. And I think one other. Thats the only one though

Should this alias be removed ? Or at least not documented (if it is currently) ?

apps/docs/src/docs/migration-guide.md
```diff
@@ -147,6 +147,43 @@ user supplied string, but the BootstrapVueNext library isn't adding an extra lay
 
 <<< DEMO ./demo/RadioGroupMigration.vue
 
+## Component aliases
+
+BootstrapVue had a number of component aliases — for instance, `<b-btn>` was an alias for `<b-button>`.
```
dwgray, line 152: Thanks for addressing this @homersimpsons! A couple of thoughts:

Shouldn't we point folks to the option of adding their own aliases to the [unplugin-vue resolver](http://localhost:8000/bootstrap-vue-next/docs.html#aliasing) if they so choose? And assuming we intend to continue to support that option.

Also, I'm not quite sure how I would do this, but we're trying to keep the names in the docs and samples to the preferred Vue 3 naming conventions (e.g. `BButton` rather than `b-button`). Maybe make the BSVN column use that convention and note the kebab syntax is still supported (since a newbie might get confused between the idea of aliasing and Vue's support for different naming conventions).

homersimpsons: I addressed all those comments and amended my commit:
- Use `BButton` (pascal case) convention in BootstrapVueNext column
- Add a paragraph about the possibility to set aliasing option
- Add a note about `BButton` vs `b-button` usage

apps/docs/src/docs/migration-guide.md
```diff
@@ -147,6 +147,47 @@ user supplied string, but the BootstrapVueNext library isn't adding an extra lay
...
+| `b-rating`                                                 | `BFormRating`         |
+| `b-select`                                                 | `BFormSelect`         |
+| `b-select-option`                                          | `BFormSelectOption`   |
+| `b-option-group`                                           | `BFormOptionGroup`    |
+| `b-tags`                                                   | `BFormTags`           |
+| `b-tag`                                                    | `BFormTag`            |
+| `b-textarea`                                               | `BFormTextarea`       |
+| `b-timepicker`                                             | `BFormTimepicker`     |
+| `b-nav-item-dd`, `b-nav-dropdown`, `b-nav-dd`              | `BNavItemDropdown`    |
+| `b-nav-toggle`                                             | `BNavbarToggle`       |
```
VividLemon, line 187: This isn't a necessary addition. The point I believe @dwgray was making was that when you refer to these 

```txt
BootstrapVue had a number of component aliases — for instance, `<b-btn>` was an alias for `<b-button>`.
```

They should be in pascal-case

This isn't really the most useful, one could simply add a line that says aliases are generally not supported and one should add them themselves

apps/docs/src/docs/migration-guide.md
```diff
@@ -147,6 +147,47 @@ user supplied string, but the BootstrapVueNext library isn't adding an extra lay
...
+
+Note: While BootstrapVueNext recommends using Vue 3 naming convention `BButton` it is still possible to use `b-button`.
```
VividLemon, line 189: This isn't a useful addition, it's not really up to us to relay this, as it's just a Vue suggestion. We should somewhat assume that a perrson migration to vue 3 should have read the vue documentation. Therefore, in my opinion, this isn't our burden of responsibility to say. Nor is it really a promotion, users can use whatever style they prefer. We simply just in our documentation and codebase are consistent to use pascal on our end.

VividLemon:
@xvaara @dwgray Should we just remove those aliases we already have?

apps/docs/src/docs/migration-guide.md
homersimpsons: > This isn't a necessary addition.

This remove the needs from a user to manually skim through BootstrapVue documentation to find aliases such as https://bootstrap-vue.org/docs/components/button#comp-ref-b-button-aliases

> The point I believe @dwgray was making was that when you refer to these [...] They should be in pascal-case

I'm not sure, the second paragraph (in https://github.com/bootstrap-vue-next/bootstrap-vue-next/pull/2771#discussion_r2214379391) was clearly mentioning the table:

>> Maybe make the BSVN column use that convention 

Furthermore, this sentence specifically refers to BootstrapVue hence to me it make sense to use their conventions.

> This isn't really the most useful, one could simply add a line that says aliases are generally not supported and one should add them themselves

Clearly a user could figure the aliases themselves but I feel like having this list they can rely on here helps in multiple scenario:
1. When evaluating upgrading to BootstrapVueNext, they can see that the list of aliases is not that big
2. They can easily find some aliases in their project to check how that would affect them
3. They can easily build the list of aliases in their project when upgrading

Should I update the BootstrapVue sentence about aliases?
Should I keep the table?

apps/docs/src/docs/migration-guide.md
homersimpsons: This was suggested in https://github.com/bootstrap-vue-next/bootstrap-vue-next/pull/2771#discussion_r2214379391:
> and note the kebab syntax is still supported (since a newbie might get confused between the idea of aliasing and Vue's support for different naming conventions)

I do not really have opinion whether we should include it there or no. But generally I wouldn't expect to see it there as we already are deep in the documentation and there is a lot of chances the user already figured this out if he wasn't familiar.

Should I remove it?

dwgray: I _do_ think this table is useful. One of my goals with the migration guide is that when someone encounters an error/warning because something has been deprecated, they can search this guide for an answer. Having this table enumerate the aliases that we deprecated will make that happen for those cases.

On the naming convention. I didn't mean to cloudy the waters here. I am fine with the table staying with the way it is (slight preference for consistency on the Vue3 end of things, but in this context, I don't care that much). I think the biggest counter-argument to my statement above is that the person doing a conversion would have enough experience to understand the kebab/Pascal translations - I had a bit of a random firing synapse on this because I'm thinking a bit more about how to make the documentation _in general_ more friendly to someone who is writing a Vue app for the first time, which generally wouldn't apply to the migration guide.

dwgray:
> @xvaara @dwgray Should we just remove those aliases we already have?

I think yes, especially if we keep the table that @homersimpsons is proposing as well as pointing to how to add their own, which he's doing as well.

homersimpsons:
Do we agree that alias removal should be part of a new pull request? Or should I tackle it there too?

dwgray:
> LGTM

I am fine with removal being a separate PR, but that's really @VividLemon's call

Issue #2778, opened 2025-07-16
BNavbar auto-close


amokrunnr:
### Describe the bug

The auto-close prop does not appear to be functional for the BNavbar component.
According to the [docs](https://bootstrap-vue-next.github.io/bootstrap-vue-next/docs/components/navbar.html), it should accept boolean, 'inside', or 'outside'. However it's typed for only boolean.
It also does not work. It is specified as true by default, however if you view the first example in the [docs](https://bootstrap-vue-next.github.io/bootstrap-vue-next/docs/components/navbar.html) on mobile and expand it, then click outside, it does not auto-close.

https://github.com/user-attachments/assets/3169e379-f2ad-40e6-8141-157e858df3c0

### Reproduction

https://bootstrap-vue-next.github.io/bootstrap-vue-next/docs/components/navbar.html

### Used Package Manager

npm

xvaara:
@VividLemon @dwgray is this a documentation error? Since this functionality was never there or it's not in BV2 either. Maybe when moving autoClose to common-props.ts from dropdown which has these functionality? I'm not sure the inside/outside makes sense for BNavbar? Also user could add this easily with vueuse to their own code: https://vueuse.org/core/onClickOutside/#onclickoutside

dwgray:
> [@VividLemon](https://github.com/VividLemon) [@dwgray](https://github.com/dwgray) is this a documentation error?

In the sense that the documentation is out of sync with the code, yes. Looking at the history, it seems that you (@xvaara) introduced the inside/outside concept in #1417 when you filled out `BDropdown` and then when @VividLemon filled out `BNavItemDropdown`, he implemented it as a thin wrapper around `BDropdown`.

So `BNavItemDropdown` already supports inside/outside values and it seems like for consistency we should enable that on the `BNav` version of the `auto-close` prop.

I've got  a draft PR #2789 in progress. @xvaara Let me know if this makes sense, and I'll do some cleanup and additional testing and get it in.

For consistency, I think the alternative path would be to pull inside/outside out of BDropdown - but if it's useful there, I'm assuming it would be helpful here.

xvaara:
@dwgray I think the problem is that in common-props.ts the description is `Controls the automatic closing of the component when clicking. See above for details.` and BNavbar is toggle-able, so the end user (like in this case the OP) is expecting the auto-close to mean the navbar, not the potential dropdowns inside it. So maybe we should rename it auto-close-dropdown?

[...]

PR #2789, opened 2025-07-27
fix(BNavBar): update auto-close behavior

dwgray:
# Describe the PR

We've implemented `inside` and `outside` values for `auto-close` on `BDropdown`. Since `BNavItemDropdown` is a thin wrapper on `BDropdown` we should be consistent and implement those values on `BNaviItemDropdown` as well.

- Change the prop on `BNavBar` to `auto-close-dropdowns`
- Add a  `auto-close-dropdowns` to `BNav`
- Use the existing inject mechanism to fall back to the prop set on the Nav[Bar]
- Remove the `tag` field form the injection key, as it wasn't used
- Fix up some types
- Update the docs

Addresses issues raised by #2778 (but doesn't fix it as the direct ask in that bug is unplanned)

## Small replication

See the examples in docx

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

## Summary by CodeRabbit

* **New Features**
  * Added support for customizing dropdown auto-close behavior in navigation and navbar components using the new `autoCloseDropdowns` property and `noAutoClose` flag.
  * Introduced new interactive demo components demonstrating various dropdown auto-close configurations.
  * Standardized dropdown auto-close options with a new type supporting `true`, `false`, `"inside"`, and `"outside"` values.

* **Documentation**
  * Expanded and clarified documentation for dropdown auto-close behavior in navigation and navbar components, including new sections and demo snippets.
  * Added type definitions and usage examples for the new auto-close options.

* **Refactor**
  * Unified and renamed internal injection keys and properties related to dropdown auto-close behavior for consistency across components.
  * Updated component props and internal logic to support the new auto-close configuration model.

### DIFF
--- a/apps/docs/src/docs/components/demo/NavbarItemDropdown.vue
+++ b/apps/docs/src/docs/components/demo/NavbarItemDropdown.vue
@@ -5,22 +5,22 @@
     variant="dark"
   >
     <BNavbarNav>
-      <BNavItem href="#">Home</BNavItem>
+      <BNavItem href="#bnavitemdropdown">Home</BNavItem>
       <BNavItemDropdown
         text="Lang"
         right
       >
-        <BDropdownItem href="#">EN</BDropdownItem>
-        <BDropdownItem href="#">ES</BDropdownItem>
-        <BDropdownItem href="#">RU</BDropdownItem>
-        <BDropdownItem href="#">FA</BDropdownItem>
+        <BDropdownItem href="#bnavitemdropdown">EN</BDropdownItem>
+        <BDropdownItem href="#bnavitemdropdown">ES</BDropdownItem>
+        <BDropdownItem href="#bnavitemdropdown">RU</BDropdownItem>
+        <BDropdownItem href="#bnavitemdropdown">FA</BDropdownItem>
       </BNavItemDropdown>
       <BNavItemDropdown
         text="User"
         right
       >
-        <BDropdownItem href="#">Account</BDropdownItem>
-        <BDropdownItem href="#">Settings</BDropdownItem>
+        <BDropdownItem href="#bnavitemdropdown">Account</BDropdownItem>
+        <BDropdownItem href="#bnavitemdropdown">Settings</BDropdownItem>
       </BNavItemDropdown>
     </BNavbarNav>
   </BNavbar>
--- a/apps/docs/src/data/components/navbar.data.ts
+++ b/apps/docs/src/data/components/navbar.data.ts
@@ -22,6 +22,12 @@ export default {
             description:
               "Set to 'top' for fixed to the top of the viewport, or 'bottom' for fixed to the bottom of the viewport",
           },
+          noAutoClose: {
+            type: 'boolean',
+            default: 'false',
+            description:
+              'Set to true to prevent the dropdown from closing automatically when an item is clicked',
+          },
           print: {
             type: 'boolean',
             default: false,
--- a/apps/docs/src/docs/components/demo/NavbarNoAutoClose.vue
+++ b/apps/docs/src/docs/components/demo/NavbarNoAutoClose.vue
@@ -1,0 +1,27 @@
+<template>
+  <!-- #region template -->
+  <BNavbar
+    v-b-color-mode="'dark'"
+    toggleable
+    no-auto-close
+    variant="dark"
+  >
+    <BNavbarBrand href="#">NavBar</BNavbarBrand>
+    <BNavbarToggle target="navbar-toggle-collapse" />
+    <BCollapse
+      id="navbar-toggle-collapse"
+      is-nav
+    >
+      <BNavbarNav class="ms-auto">
+        <BNavItem href="#collapse-no-auto-close">Link 1</BNavItem>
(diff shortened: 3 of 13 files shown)


xvaara:
I'm just copy-pasting this here:
I think the problem is that in common-props.ts the description is `Controls the automatic closing of the component when clicking. See above for details.` and BNavbar is toggle-able, so the end user (like in this case the OP) is expecting the auto-close to mean the navbar, not the potential dropdowns inside it. So maybe we should rename it auto-close-dropdown?

dwgray:
From CodeRabbitAI

> packages/bootstrap-vue-next/src/components/BDropdown/BDropdownItem.vue (1)
> 100-106: Potentially incomplete handling of 'inside' | 'outside' options
> Like in BLink.vue, the click handler collapses the navbar only when autoClose === true. If 'inside' should also collapse on an item > click (which occurs inside), extend the predicate accordingly to avoid a behavioural regression.

@xvaara do you remember the intent behind the click handler in `BDropdownItem.vue`? For dropdowns, I think we want this to hide on autoClose == true || autoClose == 'inside' - but the examples all work, so I'm not sure what's going on.

The other question is why we're doing something with collapseData here. I don't see any examples of BDropDownItems being contained in a BCollapse. Is that a supported scenario?

xvaara:
> From CodeRabbitAI
> 
> > packages/bootstrap-vue-next/src/components/BDropdown/BDropdownItem.vue (1)
> > 100-106: Potentially incomplete handling of 'inside' | 'outside' options
> > Like in BLink.vue, the click handler collapses the navbar only when autoClose === true. If 'inside' should also collapse on an item > click (which occurs inside), extend the predicate accordingly to avoid a behavioural regression.
> 
> @xvaara do you remember the intent behind the click handler in `BDropdownItem.vue`? For dropdowns, I think we want this to hide on autoClose == true || autoClose == 'inside' - but the examples all work, so I'm not sure what's going on.

I think that covers that case... but -->

> The other question is why we're doing something with collapseData here. I don't see any examples of BDropDownItems being contained in a BCollapse. Is that a supported scenario?

navbar is collapsed (when small screen) using BCollapse. And auto-close controls collapsing when link/dropdown is clicked. I don't know what 'outside' should do in this case... now it does nothing I think?

xvaara:
@dwgray I think we are confusing two auto-close on this one. auto-close on navbar collapses the nav when navigating, auto-close on dropdown closes the dropdown. So basically just the type on the navbar documentation was wrong.

Sorry, didn't remember what it actually did in navbar/collapse situation. 
Maybe it should be auto-collapse in navbar?

xvaara:
in popover there is no-auto-close, that would probably be better in line with our other components. auto-collapse would be bad because it collapses on small screens automatically. Just that the openned collapse hides automatically when navigating.

dwgray:
> in popover there is no-auto-close, that would probably be better in line with our other components. auto-collapse would be bad because it collapses on small screens automatically. Just that the openned collapse hides automatically when navigating.

Wow. Okay, yes this is kind of a mess. I like the idea of doing a simple boolean `no-auto-close` for the Navbar itself. I'll pass that down to use in the `BDropdownItem` to cause the `BCollapse` to collapse, and then if dropdownData is defined, I'll only hide for `navbarData?.autoClose === true || navbarData?.autoClose === 'inside'`. If someone wants to implement autoclose outside for the collapse on the navabar, I think that should be straightforward (and I might try this in the docs just to confirm - we've got a bit of a mess in how things work with the menu there right now).

dwgray:
> > in popover there is no-auto-close, that would probably be better in line with our other components. auto-collapse would be bad because it collapses on small screens automatically. Just that the openned collapse hides automatically when navigating.
> 
> Wow. Okay, yes this is kind of a mess. I like the idea of doing a simple boolean `no-auto-close` for the Navbar itself. I'll pass that down to use in the `BDropdownItem` to cause the `BCollapse` to collapse, and then if dropdownData is defined, I'll only hide for `navbarData?.autoClose === true || navbarData?.autoClose === 'inside'`. If someone wants to implement autoclose outside for the collapse on the navabar, I think that should be straightforward (and I might try this in the docs just to confirm - we've got a bit of a mess in how things work with the menu there right now).

Okay, I think I've got this working @xvaara - does my last commit handle the cases you were considering?

apps/docs/src/docs/components/demo/NavbarItemDropdown.vue
```diff
@@ -5,22 +5,22 @@
     variant="dark"
   >
     <BNavbarNav>
-      <BNavItem href="#">Home</BNavItem>
+      <BNavItem href="#bnavitemdropdown">Home</BNavItem>
```
dwgray, line 8: In case you're wondering, I've added these links so that the click doesn't scroll to the top and obscure the behavior we're attempting to demo. I'm considering a more general fix for the docs for href="#", at which point I can probably pull these out.

xvaara:
I was very unclear it seem. Only thing this pr should change is autoClose -> noAutoClose and document that correctly. All the other changes seems to be wrong in my view. I'll close this PR, and create a new one since you've used enough time because of my bad communication. the pr is #2799

Issue #2699, opened 2025-05-15
Docs: sidebar menu displays bug


xvaara:
### Discussed in https://github.com/bootstrap-vue-next/bootstrap-vue-next/discussions/2407

<div type='discussions-op-text'>

<sup>Originally posted by **enix90s** December  2, 2024</sup>
![BSV-docs-2](https://github.com/user-attachments/assets/956ccbfc-e5ba-4be3-86d5-b109d520071c)
</div>

Benalex8797:
I will first replicate the sidebar menu display bug in the Docs to identify the root cause whether it's a CSS layout issue, JavaScript interaction problem, or a responsive design conflict. I will then inspect the sidebar component structure and its state logic to ensure it behaves correctly on all screen sizes and interactions. My fix will ensure consistent, smooth rendering across devices and user actions, with a focus on accessibility and performance.
I have strong experience in frontend development, especially with responsive UI design, JavaScript, and CSS. I've worked on similar issues in previous projects, such as building dynamic navigation menus and resolving display inconsistencies across devices. I’m confident in debugging layout and interaction issues.
I enjoy solving UI bugs because they have an immediate impact on user experience. This issue is an opportunity to contribute meaningfully to the Docs interface while showcasing my frontend skills and attention to detail.

VividLemon:
It definitely doesn't pertain but when accessing the site through duckduckgo it led me to /docs and not /docs.HTML which ended up breaking the layout in weird ways. The on this page section didnt have anything, and various content didnt exist while some content did.

Was an annoying issue since one could assume accessing the site through a search engine would have their first interaction with the lib to be a mess.

a-kriya:
Been noticing the same issue. Here's a clip:

https://github.com/user-attachments/assets/e986c387-5245-4df6-8c09-f1c8d6d436b2

I built the docs and looks like HTML files are output for every route but only with a VitePress schema and an empty `#app` element, intended to be rendered on the client. This is a bit strange: either it should be a SPA with a single `index.html`, or a fully rendered site. If it's configured like this for SEO purposes, I don't think it'll help much to have multiple HTML pages but with no rendered content.

PR #2798, opened 2025-08-04
Resolve warnings when building docs

a-kriya:
# Describe the PR

While looking into #2699, I ended up resolving an error and a warning that would flood the screen when building docs:

- `ReferenceError: document is not defined` (from `useScrollspy` call in Layout.vue)
- "resolveDirective" is imported from external module "vue" but never used in.... (from auto-importing via `unplugin-vue-components`)

## PR checklist

- [ ] Feature - `feat(...)`
- [X] Other (please describe)


apps/docs/.vitepress/config.mts
```diff
@@ -30,11 +30,14 @@ export default defineConfig({
...
         resolvers: [BootstrapVueNextResolver()],
-        // eslint-disable-next-line @typescript-eslint/no-explicit-any
-      }) as any,
+        include: [/\.vue$/, /\.vue\?vue/],
+      }),
+      // separating this prevents unused "resolveDirectives" import warning on build
+      Components({
+        resolvers: [BootstrapVueNextResolver({directives: false})],
+        include: [/\.md$/],
+      }),
```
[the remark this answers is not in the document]
a-kriya: The TS error should be addressed in the next commit on top. Are you evaluating the commits individually?

a-kriya:
@dwgray @VividLemon Both `type-check` and `build` works for me locally, but failing in the CI for some reason. Any ideas?

VividLemon:
Different package versions perhaps

a-kriya:
I regenerated the lockfile, so CI's cache should be invalidated and the deps reinstalled.

xvaara:
The scrollspy should be fixed to be ssr safe, not to go around the error. I've fixed the warning in #2800

a-kriya:
@xvaara If the other warning fix is worth it, I'll rebase, if not I'll drop this PR.

apps/docs/.vitepress/config.mts
```diff
@@ -30,11 +30,14 @@ export default defineConfig({
       Icons(),
       Components({
         globs: ['components/*.vue', 'docs/**/demo/*.vue'],
-        dts: true,
-        include: [/\.vue$/, /\.vue\?vue/, /\.md$/, /\.ts$/],
         resolvers: [BootstrapVueNextResolver()],
-        // eslint-disable-next-line @typescript-eslint/no-explicit-any
-      }) as any,
+        include: [/\.vue$/, /\.vue\?vue/],
```
dwgray, line 34: Thanks @a-kriya  for taking a look at the doc warnings! I think we still need to have .ts included in one or the other of these resolvers - take a look at my notes in #2724 - Basically, if you get rid of this, my expectation is that the samples will silently stop working in the built version of the docs (but not in the dev version)...

apps/docs/.vitepress/theme/Layout.vue
```diff
@@ -286,6 +267,29 @@ const colorMode = useColorMode({
 })
 
 onMounted(() => {
+  const {current, list} = useScrollspy(content, target, {
```
dwgray: I'm assuming this improves SSR in some way? If the contents aren't rendered at the point that useScrollspy is called, I'd expect the query to completely fail and the whole thing to not work. I have basically no experience with SSR, so I'll take your word for this, but if you've got a moment to let me know why and/or point me to a resource that would help me understand this, I'd appreciate it.

dwgray (changes requested):

apps/docs/.vitepress/theme/Layout.vue
a-kriya: This change will be gone as it was addressed in #2800 in the hook's source instead of here.

Anything that accesses browser-only APIs (like `document`) will lead to errors when run by Node, so care must be taken to only have such code run on client-side. `mounted` only triggers when a component actually mounts in the browser DOM, so that's one way to handle this. Another is to wrap components in `<ClientOnly>` tags to prevent rendering them on the server. #2800 addresses it by explicitly checking for presence of `document`.

apps/docs/.vitepress/config.mts
a-kriya: Ah, okay. Thanks for pointing. I'll update this PR.

a-kriya: Adding back `.ts` brought back the warning regarding `resolveDirective`, and I couldn't find a way to suppress it.

Well, at least the more egregious one has now been fixed in #2800. I'll continue looking into #2699, and close this one.

apps/docs/.vitepress/theme/Layout.vue
dwgray: Thanks, @a-kriya.

PR #2806, opened 2025-08-09
docs: fix missing data and use a stronger schema to catch missing errors

VividLemon:
# Describe the PR

## Small replication

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)


apps/docs/src/components/ComponentReference.vue
```diff
@@ -78,39 +87,44 @@
...
                               {{ table.name }}
                             </BLink>
                             <template v-else>
                               {{ table.name }}
                             </template>
                           </template>
                           <BTable
-                            :items="table.ref"
-                            :fields="fields.props"
+                            :items="table._data.map((el) => ({prop: el[0], ...el[1]}))"
```
dwgray, line 126: I was having some difficulty following this logic, so I grabbed your code to run it locally. Something is not working in the case where there are extensions, I'm not seeing the props:
image
I'm guessing something is going wrong in `sortData`?

apps/docs/src/components/NoteAlert.vue
```diff
@@ -5,7 +5,7 @@
 </template>
 
 <script setup lang="ts">
-import {BAlert} from 'bootstrap-vue-next'
+import {BAlert} from 'bootstrap-vue-next/components/BAlert'
```
dwgray: In the cases where we don't explicitly need the type of a component, can't we just remove the import and let unplugin do the work for us? I think I asked this before, but since I have the code on my machine I tried deleting the whole script section here and it seems to work fine (I tried both dev and build mode). Even if I'm correct here, I wouldn't block this PR on that, I just want to make sure I'm understanding the implications correctly.

dwgray (changes requested):
Overall this is a big improvement - I kept running into issues where event/slot docs weren't being kept in sync with the code, so this should help that. 

On a quick run through the code, I only saw the two things that I comment on separately, and of those, I think only the extensions issue needs to be fixed with this PR.

apps/docs/src/components/NoteAlert.vue
VividLemon: Correct

apps/docs/src/components/ComponentReference.vue
VividLemon: There was actually a `// This is wrong` note :rofl:

VividLemon: https://github.com/bootstrap-vue-next/bootstrap-vue-next/pull/2806/commits/8b0094763737913bc6f6b01a3a5deeead5ff788b#diff-f15ea091ca11917a21ef96530b8fce06000e61f07ca211dfdb98772b5154813dL289

PR #2869, opened 2025-10-02
docs: Enable directly loading examples into StackBlitz

dwgray:
# Describe the PR

Add a button to the examples in our documentation to open them in StackBlitz.

I implemented this by using the StackBlitz  sdk and pulling in our existing vite template from source and substituting the demo code for the Comp.vue file in our template.

Tested in vite and build modes and all works without errors or warnings.

This only works for examples that are factored into their own files, so once this goes through I will do a pass to finish that refecotring as soon as I’ve got some cycles.

## Small replication

Most of the examples in the component docs.

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

## Summary by CodeRabbit

- New Features
  - Docs examples now show a language label and an “Open in StackBlitz” button to launch a live playground when full source is available.

- Bug Fixes
  - More reliable build-time reading and handling of example files with graceful fallbacks and reduced duplicate path resolution.

- Chores
  - Added StackBlitz SDK as a dev dependency, new raw-template and Vite env type declarations, and updated TypeScript config to include them.


apps/docs/src/components/HighlightCard.vue
```diff
@@ -4,10 +4,102 @@
...
+      </div>
       <BCardBody class="bg-body-tertiary">
         <slot name="html" />
       </BCardBody>
     </template>
   </BCard>
 </template>
+
+<script setup lang="ts">
+import {computed} from 'vue'
```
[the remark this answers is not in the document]
dwgray: This is incorrect for our project configuration. If I make this change it gets autocorrected back to the no-space version on save.

apps/docs/src/components/HighlightCard.vue
```diff
@@ -4,10 +4,102 @@
...
+  'tsconfig.json': tsConfig,
+  'tsconfig.node.json': tsConfigNode,
+})
+
+const openInStackBlitz = async () => {
+  if (!fullFileContent.value) return
+
+  try {
+    // Dynamically import StackBlitz SDK
+    const {default: sdk} = await import('@stackblitz/sdk')
```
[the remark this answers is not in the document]
dwgray: This is incorrect for our project configuration. If I make this change it gets autocorrected back to the no-space version on save.

apps/docs/src/components/HighlightCard.vue
```diff
@@ -4,10 +4,102 @@
...
+
+const createProjectFiles = () => ({
+  'index.html': indexHtml,
+  'src/main.ts': mainTs,
+  'src/App.vue': appVue,
+  'src/components/Comp.vue': fullFileContent.value || '',
+  'package.json': packageJson,
+  'vite.config.ts': viteConfig,
+  'tsconfig.json': tsConfig,
+  'tsconfig.node.json': tsConfigNode,
```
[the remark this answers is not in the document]
dwgray: Updated based on this suggestion

apps/docs/src/components/HighlightCard.vue
```diff
@@ -4,10 +4,121 @@
...
+})
+
+const openInStackBlitz = async () => {
+  if (!props.fullFile) return
+
+  try {
+    // Dynamically import StackBlitz SDK
+    const {default: sdk} = await import('@stackblitz/sdk')
```
VividLemon, line 109: This seems like it could be a top level import. Not sure why it's dynamically imported.

apps/docs/.vitepress/plugins/demo-container.ts
```diff
@@ -41,18 +42,34 @@ export const demoContainer = (md: MarkdownRenderer, srcDir: string) => {
...
     const component = isDemo ? `<${title.substring(0, title.indexOf('.'))}/>` : ''
 
+    // Resolve path early so it can be used in the prefix token
+    const {realPath, path: _path} = state.env as MarkdownEnv
+    const resolvedPath = path.resolve(path.dirname(realPath ?? _path), filepath)
+
+    // Read the source code file during build time
+    let fullFileContent = ''
+    try {
+      if (fs.existsSync(resolvedPath)) {
```
VividLemon: This is actually a race condition. If memory serves, it's mostly recommended to let readFile throw an error, rather than checking for access/exist

VividLemon: Although practicality it does nothing here

VividLemon (changes requested):

apps/docs/src/components/HighlightCard.vue
dwgray: Yeah, this PR is an attempt at vibe coding. I asked the AI why it dynamically imported the SDK and it gave me a bunch of reasons. Reading back, I think they’re all bogus, because in the case we care about (which is the build case), this code will run for every example unconditionally, so no reason not to load the SDK statically. I’ll make the change. 

**UPDATE:** I misread - this is in HighlightCard.vue, I was thinking it was in democontainer.ts - in this case I think the the SDK code doesn’t get run until someone hits the stackblitz button, so the reasons below make sense to me. Let me know if I’m missing something, please.

—

If you’re interested, here is what Claude had to say:

1. Bundle Size Optimization
The StackBlitz SDK is a relatively large library that most users will never use. By dynamically importing it:

It’s not included in the main bundle - only loaded when someone actually clicks the StackBlitz button
Reduces the initial page load size for all documentation visitors
Follows the principle of “load only what you need, when you need it”
2. Code Splitting
Dynamic imports create automatic code splitting:

Vite/Webpack automatically splits this into a separate bundle file that’s loaded asynchronously.

3. Performance
Faster initial page loads - the docs load faster since StackBlitz SDK isn’t in the main bundle
Better perceived performance - users see the page content immediately
Lazy loading - only users who actually want to use StackBlitz pay the cost of downloading it
4. Error Handling
Dynamic imports allow us to gracefully handle cases where:

The SDK fails to load (network issues)
The SDK has breaking changes
StackBlitz service is unavailable
5. Optional Dependency
Since StackBlitz integration is a “nice-to-have” feature rather than core functionality:

The docs work perfectly fine without it
If the SDK has issues, it doesn’t break the documentation
We can easily remove or replace it later without major refactoring

apps/docs/.vitepress/plugins/demo-container.ts
```diff
@@ -41,18 +42,32 @@ export const demoContainer = (md: MarkdownRenderer, srcDir: string) => {
...
 
+    // Resolve path early so it can be used in the prefix token
+    const {realPath, path: _path} = state.env as MarkdownEnv
+    const resolvedPath = path.resolve(path.dirname(realPath ?? _path), filepath)
+
+    // Read the source code file during build time
+    let fullFileContent = ''
+    try {
+      fullFileContent = fs.readFileSync(resolvedPath, 'utf-8')
+    } catch (error) {
```
VividLemon, line 53: It's likely better to let this throw during compilation

apps/docs/src/components/HighlightCard.vue
VividLemon: Sure. Add a loading indicator to the button though

VividLemon (changes requested):
Last was wrong

VividLemon (approved):

PR #2880, opened 2025-10-15
fix(BToggle): console error when vBToggle directive unmounted

### LINKED ISSUE #2856: BToggle directive producing console error when unmounted quickly
### Describe the bug

While updating to BSV v0.40 I've noticed the following console error popping up occasionally in my logs, both during unit tests and in the browser:

> Could not find original vnode, will not inherit provides

It's being thrown from the call to findProvides from the handleUnmount method of the BToggle directive. It appears that findProvides throws that error if the element passed in has been removed from the DOM. This error can be reproduced by adding a v-if to the element that the directive is on and then changing it to false after it initially being true, forcing an unmount.

The logic inside findProvides is a bit too complex for me to attempt a PR myself, though I did notice that the project that that code was borrowed from (vuetify) is not using it in an unmount handler, just in mount.

### Reproduction

https://stackblitz.com/edit/github-tkn26tgr?file=src%2Fcomponents%2FComp.vue

### Used Package Manager

npm

IGx89:
# Describe the PR

Fixes #2856 by changing injection logic used in directive hooks. I ended up greatly simplifying the logic by looking at what Vue's inject method does behind the scenes and matching it. I don't know the background of the original logic here and why it was so complex, but as far as I can tell the complexity wasn't needed. All unit tests are passing, though coverage in this area appears somewhat low.

Main downside of this approach is the use of an internal API, however the previous solution was using even more internal APIs so that seems to be considered acceptable.

Ideally we would simply use `inject`, however until https://github.com/vuejs/core/issues/5002 is resolved we won't be able to.

## Small replication

See repro on #2856 as well as added unit test.

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

## Summary by CodeRabbit

* **Refactor**
  * Simplified directive internals for Popover, Tooltip, and Toggle to use a direct injection approach and remove vnode dependence, reducing complexity and improving reliability.

* **Bug Fixes**
  * Safer handling when defaults are absent to prevent potential runtime errors in Popover and Tooltip.

* **Tests**
  * Added comprehensive tests validating directive-based injections and override behavior across nested providers, ensuring correct lifecycle handling and stability.

### DIFF
--- a/packages/bootstrap-vue-next/src/directives/BPopover/index.ts
+++ b/packages/bootstrap-vue-next/src/directives/BPopover/index.ts
@@ -1,4 +1,4 @@
-import {type Directive, type Ref} from 'vue'
+import {type Directive} from 'vue'
 import {
   bind,
   type ElementWithPopper,
@@ -8,11 +8,11 @@ import {
   unbind,
 } from '../../utils/floatingUi'
 import {defaultsKey} from '../../utils/keys'
-import {findProvides} from '../utils'
+import {injectWithinDirective} from '../utils'
 
 export const vBPopover: Directive<ElementWithPopper> = {
-  mounted(el, binding, vnode) {
-    const defaults = (findProvides(binding, vnode) as Record<symbol, Ref>)[defaultsKey]?.value
+  mounted(el, binding) {
+    const defaults = injectWithinDirective(binding, defaultsKey)?.value
     const isActive = resolveActiveStatus(binding.value)
     if (!isActive) return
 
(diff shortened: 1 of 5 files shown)


IGx89:
Converted to draft while I perform some manual testing -- found one scenario that might not be handled by this.

IGx89:
A different approach is needed unfortunately. This works great if the directive is on a component but fails if it's on a plain element and BApp is used (showHide registry not an app-level provide).

PR #2889, opened 2025-10-24
fix(BOffcanvas): Add a footer class

m-mohr:
# Describe the PR

title and body have default classes set, but footer has not. Why? It seems inconsistent, so adding one.

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

## Summary by CodeRabbit

* **Style**
  * Improved offcanvas footer styling by adding a dedicated CSS class to the footer container when a footer slot is present, preserving any existing footer class bindings for compatibility and consistent appearance.

### DIFF
--- a/packages/bootstrap-vue-next/src/components/BOffcanvas/BOffcanvas.vue
+++ b/packages/bootstrap-vue-next/src/components/BOffcanvas/BOffcanvas.vue
@@ -59,7 +59,7 @@
           <div class="offcanvas-body" :class="props.bodyClass" v-bind="props.bodyAttrs">
             <slot v-bind="sharedSlots" />
           </div>
-          <div v-if="hasFooterSlot" :class="props.footerClass">
+          <div v-if="hasFooterSlot" class="offcanvas-footer" :class="props.footerClass">
             <slot name="footer" v-bind="sharedSlots" />
           </div>
         </template>


dwgray:
@m-mohr - The reason that we have `offcanvas-body` and `offcanvas` header classes set on the body and header is that they are part of what is required by bootstrap to make the offcanvas component work - https://getbootstrap.com/docs/5.3/components/offcanvas/#offcanvas-components

As far as I can tell, the bootstrap offcanvas component _does not include_ a footer. In general, we’ve taken a minimalist approach and not included placeholder classes in cases like this. So I think the real question is why we have a footer slot at all, if BS5 doesn’t. It does allow one to put content at the bottom and outside of the body div, so that might be useful. @VividLemon / @xvaara thoughts?

@m-mohr the other way to solve this problem (assuming we keep a footer slot and don’t add an explicit class) - is to add your own class via the `footer-class` prop.

VividLemon:
Yes, the reason why it was never intentionally included is because it's not referenced in bootstrap. Having a class that doesn't do anything isn't quite useful, especially when you can add your own classes. It doesn't feel like it has much purpose. Unlike offcanvas header, which does have css associated with it

m-mohr:
Okay, fair enough. Just felt like a bug when you get \<div class=""> for the footer, while all other sections actually have a class assigned by default. So the conclusion is to close this PR and use a custom class instead?

VividLemon:
Yes

PR #2899, opened 2025-11-07
fix: BFormRadioGroup - prevent forwarding option attrs that override input id

### LINKED ISSUE #2898: BFormRadioGroup: click not registering when using props text-field / value-field
### Describe the bug

With BFormRadioGroup, when using props text-field & value-field, the mouse click is not registering at all when using button style, and is only registering on the html input element when using classical style.

### Reproduction

https://stackblitz.com/edit/6m5rdsm1?file=src%2Fcomponents%2FComp.vue

### Used Package Manager

npm

Saayanel:
Fixes #2898

Disclaimer: this fix was coded by Copilot using "Code with agent mode" button.

### Summary

This PR fixes a bug where object option keys (for example `Id`) were forwarded via $attrs onto the native <input> and could override the computed id (HTML attribute names are case-insensitive). That broke the label for association and prevented clicks on labels/buttons from selecting the radio when using textField/valueField and buttons styles.

What I changed:
- Do not spread original option object into normalized option. Build a clean object with only: id, value, text, disabled.
- Generate deterministic id per option (<group-id>__<index>) when not provided.
- Add unit test that reproduces the case where options use capitalized Id as valueField and buttons: true.

Validation:
- Built package and ran tests locally: all unit tests pass and lint has no errors (only minor warnings).

Files changed:
- packages/bootstrap-vue-next/src/components/BFormRadio/BFormRadioGroup.vue
- packages/bootstrap-vue-next/src/components/BFormRadio/form-radio-group.spec.ts

Please review and let me know if you prefer forcing generated ids even when id is provided by options.

### DIFF
--- a/packages/bootstrap-vue-next/src/components/BFormRadio/BFormRadioGroup.vue
+++ b/packages/bootstrap-vue-next/src/components/BFormRadio/BFormRadioGroup.vue
@@ -90,18 +90,29 @@ const normalizeOptions = computed<
     [key: string]: unknown
   }[]
 >(() =>
-  props.options.map((el) =>
+  props.options.map((el, index) =>
     typeof el === 'string' || typeof el === 'number'
       ? {
+          id: `${computedId.value}__${index}`,
           value: el,
           disabled: props.disabled,
           text: el.toString(),
         }
       : {
-          ...el,
-          value: el[props.valueField],
-          disabled: el[props.disabledField] as boolean | undefined,
-          text: el[props.textField] as string | undefined,
+          // Only include a small set of well-known properties on the
+          // normalized option object. We deliberately DO NOT spread the
+          // original `el` object because arbitrary keys (for example a key
+          // named `Id`) would be forwarded as attributes ($attrs) to the
+          // child `BFormRadio` and then bound onto the native <input>, which
+          // can inadvertently override the input `id` (case-insensitive in
+          // HTML) and break the label `for` association. By building a
+          // clean object we avoid such accidental attribute collisions.
+          id: (el as Record<string, unknown>).id ?? `${computedId.value}__${index}`,
+          value: (el as Record<string, unknown>)[props.valueField] as unknown,
+          disabled: (el as Record<string, unknown>)[props.disabledField] as
+            | boolean
+            | undefined,
+          text: (el as Record<string, unknown>)[props.textField] as string | undefined,
         }
   )
 )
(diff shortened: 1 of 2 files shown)


VividLemon:
What I've noticed is that because Id is uppercased in your example, it ends up getting processed as an attribute, rather than a prop. The handler in BFormCheckbox runs correctly, it appears to be correct in that when you spread the attributes, Id becomes an attribute id on the element. It appears correct, as that would be the expected output of the `computedId`, however, the truth of that is that it is due to it being processed as an attribute. 

Then that also explains why computedId is wrong for the label. It doesn't see the uppercased Id as a prop, as that is being processed as an attribute. So it's building what is the computedId

This is what's happening, the attribute is taking over:
image

image


It appears like the reproduction is more an issue with using Vue than it is with the library. As if you were using the component variation, one would see (with TS) that you'd have a type issue

Saayanel:
Interesting, but why is there a spread in the first place ? The docs says only 'value', 'text' and 'disabled' are supported. Is the purpose of the spread to support every props of BFormRadio/BFormCheckbox ?

image

VividLemon:
> Interesting, but why is there a spread in the first place ? The docs says only 'value', 'text' and 'disabled' are supported. Is the purpose of the spread to support every props of BFormRadio/BFormCheckbox ?
> 
> image

Perhaps its best to add in the documentation. The purpose is that anything could be forwarded, attributes and props alike. It's just that those specific values are special

Saayanel:
Then, maybe we should prevent the forwarding of the attributes declared in textField / valueField ? I don't see a case where one would expect/need the properties contained in textField / valueField to be forwarded.

Something like:

```
const normalizeOptions = computed<
  {
    text: string | undefined;
    value: any;
    disabled?: boolean | undefined;
    [key: string]: unknown;
  }[]
>(() =>
  props.options.map((el) =>
    typeof el === 'string' || typeof el === 'number'
      ? {
          value: el,
          disabled: props.disabled,
          text: el.toString(),
        }
      : (() => {
          const clone: Record<string, unknown> = { ...el };

          delete clone[props.valueField as string];
          delete clone[props.textField as string];
          delete clone[props.disabledField as string];

          return {
            ...clone,
            value: (el as any)[props.valueField],
            disabled: el[props.disabledField] as boolean | undefined,
            text: el[props.textField] as string | undefined,
          };
        })()
  )
);
```

VividLemon:
https://github.com/bootstrap-vue-next/bootstrap-vue-next/pull/2899#issuecomment-3533685940

PR #2929, opened 2025-11-30
feat(BFormSelect): add properties to add styling for option element

anrolmar:
# Describe the PR

- Added optionAttrs on BFormSelect and BFormSelectOption to style the option elements

## Small replication

If we want to set some styling for the option element, we could do:

`
<BFormSelect v-model="selected" :options="ex1Options" :options-attrs="{ style:'background: red; color: white'}" />
`

so the result would be

Captura de pantalla 2025-11-30 164249

## PR checklist

- [x] Feature - `feat(...)`
- [ ] Other (please describe)


VividLemon:
added in different pr

PR #2973, opened 2026-01-06
Refactor btable

VividLemon:
# Describe the PR

## Small replication

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

## Summary by CodeRabbit

* **Breaking Changes**
  * "Row details" renamed to "row expansion"; several table and tabs events now emit a single object payload; field.key no longer supports nested paths (use accessor).

* **New Features**
  * v-model for expanded rows, controller-style expansion/selection APIs, richer table event objects, new lightweight table helpers/controllers, and additional exported table types.

* **Documentation**
  * Migration guide and docs updated for event shapes, expansion model, field/accessor changes, and tab activation payloads.

* **Chores**
  * Multiple dependency and build/export updates.

* **Tests**
  * Tests updated for new payload shapes, slot names, and APIs.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


apps/docs/src/docs/migration-guide.md
```diff
@@ -1054,15 +1054,104 @@ that is emitted for each row that is unselected. There is also a named model `se
...
+<script setup>
+import {ref, onMounted} from 'vue'
+
+const tableRef = ref()
+const expandedItems = ref([])
+
+// Set default expanded items after mount
+onMounted(() => {
+  expandedItems.value.push(tableRef.value.get(items[1]))
+})
```
tbl0605, line 1149: > v-model:expanded-items for controlling row expansion with primary-key persistence.

Hi @VividLemon,
I'm not entirely sure how the `v-model:expanded-items="expandedItems"` feature will work here, sorry to bother you with some questions ^^ 

What happens when the `items` list is reset to an empty array? Will the `expandedItems` list also be emptied automatically? Or do we need to add watchers to "manually" handle `expandedItems` 's content depending on the `items` list?
What happens when an "expanded" item is removed from the `items` list using `pop()` or `shift()` ? Will it also be removed from the `expandedItems` list?

My concern is that after some times the `expandedItems` list will contain elements that are no longer in the `items` list.
I don't know if it's already been done, but I would prefer that this synchronization process be implemented internally by the BTable component rather than by the user, if that is feasible of course.

VividLemon: Nothing internal was implemented, but may be best to add. I'm not sure the best way for this though. 

Is the intended behavior that if you had 3 pages, expanded (or use the selected item feature) an item on page 1, then moved to page 3, then back to page 1, should it suddenly disappear? (Lets assume that if the item is an object, that it is not a **new** object and is the same reference for equality)

VividLemon: The previous implementation internally kept track of things with a WeakMap. That way if the user disposed of the object, that it would automatically leave the `ref`. But this was difficult to manage a synced array prop and weakmap ref, had edge case issues. It could be possible to make the prop some sort of iterable, that the user could just pass in their own weakmap I guess.

tbl0605: > Nothing internal was implemented, but may be best to add. I'm not sure the best way for this though.

Ideally, I'd like to be able to use the "item expansion feature" without having to explicitly provide the `expandedItems` list via `v-model:expanded-items`. This means the BTable should be able to internally store its own expansion list, capable of emptying or cleaning itself when:
a) a new list of "items" (as a new array) is provided
b) the `pop()` or `shift()` functions are used on the item list (I don't use these, but it might be cleaner to implement).

Why? Because I simply need to be able to expand the table rows, nothing more, nothing less. Until now, I added `_showDetails` to my "items", and everything was handled automatically. So why should I declare and manage a secondary list just to enable the "item expansion feature"?

The user can always explicitly provide `v-model:expanded-items="expandedItems"` if he wants to perform more complex operations, such as the initial expansion of a row, etc...

I hope you understand my point of view ^^

VividLemon: You don't. Suppyling a value to v-model is optional

tbl0605: I'll see once it's merged, thank you ^^

apps/docs/src/docs/migration-guide.md
```diff
@@ -1054,15 +1054,104 @@ that is emitted for each row that is unselected. There is also a named model `se
...
 
+#### Row Expansion (formerly Row Details)
+
+**Terminology changes:** BootstrapVue used "details" terminology for expanding rows, which has been changed to "expansion" for clarity and linguistic correctness. The following changes have been made:
+
+- Scoped slot variable `detailsShowing` is now `expansionShowing`
+- Scoped slot function `toggleDetails` is now `toggleExpansion`
+- The concept of "detailed items" is now "expanded items"
+
+**v-model instead of object property:** The expansion state is no longer tracked using a property on item objects. Instead, use the `v-model:expanded-items` binding (or its alias `v-model:item-details` for compatibility) to manage which rows are expanded.
```
VividLemon:
```suggestion
**v-model instead of object property:** The expansion state is no longer tracked using a property on item objects. Instead, use the `v-model:expanded-items` binding to manage which rows are expanded.
```

packages/bootstrap-vue-next/src/components/BTable/BTableLite.vue
```diff
@@ -425,53 +356,24 @@ const calculatedFooterSlot = (key: LiteralUnion<keyof Items>): keyof typeof slot
...
-const toggleRowDetails = (tr: Items) => {
-  if (isTableItem(tr)) {
-    const prevValue = getDetailsValue(tr)
-    setDetailsValue(tr, !prevValue)
-    tr._showDetails = !prevValue
-  }
-}
-
+/**
+ * @deprecated I am unsure if the best path forward is to separate the accessor and the key
```
VividLemon: @dwgray @xvaara Thoughts on this one? It's best to do in this PR if we'd like to

dwgray: I would come down on the side of making the breaking change now. Have a separate (optional) accessor prop that needs to be used whenever the key is not a simple field on the item object. Then we can strongly type key.  Something like:

```ts
export interface TableField<T = Record<string, unknown>> {
  key: LiteralUnion<keyof T>
  label?: string
  accessor?: string | ((item: T) => unknown) // String path or getter function
  // ...existing properties
}
```
Sample use:

```ts
// Simple property (no accessor needed)
const fields = [
  { key: 'name' }, // Accesses item.name directly
  { key: 'email' }
]

// Nested property
const fields = [
  { key: 'userCity', label: 'City', accessor: 'user.address.city' }
]

// Custom function
const fields = [
  { 
    key: 'fullName', 
    accessor: (item) => `${item.firstName} ${item.lastName}`
  },
  {
    key: 'status',
    accessor: (item) => item.isActive ? 'Active' : 'Inactive'
  }
]

// Key can have special characters when using accessor
const fields = [
  { key: 'user-email', accessor: 'user.email' } // Key doesn't have to match accessor
]
```
Or we could simplify further and require the user to pass in a function if they are doning anything more complex that simple field lookup - I feel like we made that choice for something else recently, but can’t remember what…

VividLemon: Part of the issue with that is do we fall back to the implicit previous behavior, or break away so each has a single responsibility? Keys for stability, get for item fetching, label for what to display? A simple "key" that does it all certainly is quickest and simplest

packages/bootstrap-vue-next/CHANGELOG.md
dwgray, line 1: Looks like this is just whitespace changes? I've noticed that loading this file in VSCode sometimes changes the markup, but hadn't noticed this particular behavior before - nothing blocking here, just want to understand so we do bounce back and forth between formats...

packages/nuxt/CHANGELOG.md
dwgray, line 1: Same questions as with the main BSVN changelog

packages/bootstrap-vue-next/src/components/BTable/BTable.vue
```diff
@@ -230,169 +222,141 @@ const dynamicFootSlots = computed(
...
-    })
+const providerController = useTableProvider({
+  events: {
+    onFiltered: () => {
+      emit('filtered', computedItems.value)
+    },
+  },
+  busy: busyModel,
+  provider: toRef(() => props.provider),
+  debounce: {
```
dwgray, line 270: I'm seeing type errors on `debouce.maxWait`, `debounce.wait`, `currentPage` and `perPage`

```
Type 'ComputedRef<number>' is not assignable to type 'MaybeRef<number>'.
  Type 'ComputedRef<number>' is missing the following properties from type 'WritableComputedRef<number, number>': [WritableComputedRefSymbol], [ComputedRefSymbol], [RefSymbol]
```

It looks like the fix may be just to tighten the types for these to `MaybeRef<number>`?

packages/bootstrap-vue-next/src/components/BTable/BTable.vue
```diff
@@ -421,494 +385,67 @@ const getRowClasses = (item: Items | null, type: TableRowType): TableStrictClass
...
+}))
+
 defineExpose({
+  expansion: {
+    ...expandedItemsController,
+    expandedItems: readonly(expandedItems),
+  },
   // The row selection methods are really for compat. Users should probably use the v-model though
-  ...exposedSelectableUtilities,
+  selection: {
```
dwgray, line 441: This is a breaking change - you've fixed up the samples and the docs in `table.md`, but I'm not seeing a comment on this in `migration-guide.md`

dwgray (commented):
Looks great! I think he expansion model is a significant improvement. See my inline comments, but I they're basically just making sure we keep the migration guide up-to-date and don't thrash with whitespace changes in the chagen logs.

packages/bootstrap-vue-next/CHANGELOG.md
VividLemon: Sometimes the linter would automatically run it against the changelog. I think it's a vscode extension I have to lint md files. It's not really an issue.

packages/bootstrap-vue-next/src/components/BTable/BTable.vue
VividLemon: I think this may be your environment. Perhaps you need to update your node_modules. All of those types are `MaybeRefOrGetter` not MaybeRef

packages/bootstrap-vue-next/src/components/BTable/BTable.vue
VividLemon: Done

PR #3142, opened 2026-04-01
fix(modal): don't steal focus after shown handler focuses input

saschabuehrle:
Fixes #3141

When no explicit `focus` prop is provided, focus trap activation could still force focus back to the modal root even if a consumer already focused an input in the `shown` handler.

This change keeps existing behavior for explicit `focus` values, but skips forcing initial focus when an element inside the modal is already active. Added a regression test that focuses an input from `onShown` and verifies focus remains there.

Greetings, saschabuehrle

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed modal focus trapping behavior to avoid redirecting focus when the user's focus is already positioned on an element contained within the modal, resulting in a smoother interaction experience with form inputs and other interactive components.

### DIFF
--- a/packages/bootstrap-vue-next/src/components/BModal/BModal.vue
+++ b/packages/bootstrap-vue-next/src/components/BModal/BModal.vue
@@ -319,7 +319,20 @@ const {needsFallback} = useActivatedFocusTrap({
     ref: fallbackFocusElement,
     classSelector: fallbackClassSelector,
   },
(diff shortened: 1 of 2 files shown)


VividLemon:
@saschabuehrle This implementation and https://github.com/bootstrap-vue-next/bootstrap-vue-next/pull/3143 both appear to have merits. And both seem to fix the issue. I am not sure the best implementation, or perhaps both are warranted. Could you look over the other PR here and see which seems better (or both) for the issue you raised?

PR #3154, opened 2026-04-14
feat(BAspect): Create BAspect component

NilLlisterri:
# Describe the PR

This pull request migrates the BAspect component from BootstrapVue keeping the properties (aspect, tag) and slot (default).
In Bootstrap 5 ratios are specified using "16x9" instead of "16:9", so I've added this possiblity keeping the old.

## Small replication

A demo has been added to the documentation.

## PR checklist

- [x] Feature - `feat(...)`
- [ ] Other (please describe)

## Summary by CodeRabbit

* **New Features**
  * Added BAspect component for responsive aspect-ratio containers (accepts numeric or string ratios like 16:9, 4:3, 1.5); renders at 100% width.

* **Documentation**
  * Added component docs with overview, behavior details, usage examples, and an interactive demo showing default 1:1 and 16:9.

* **Tests**
  * Added tests covering tag rendering, reactivity, structure, and multiple aspect formats.

### DIFF
--- a/packages/bootstrap-vue-next/src/components/BAspect/BAspect.vue
+++ b/packages/bootstrap-vue-next/src/components/BAspect/BAspect.vue
@@ -1,0 +1,32 @@
+<template>
+  <component :is="props.tag" class="b-aspect d-flex">
+    <div :style="{paddingBottom: computedPadding, height: 0}" class="b-aspect-sizer flex-grow-1" />
+    <div class="b-aspect-content flex-grow-1 w-100 mw-100" style="margin-left: -100%">
+      <slot />
+    </div>
+  </component>
+</template>
+
+<script setup lang="ts">
+import {computed} from 'vue'
+import type {BAspectProps} from '../../types/ComponentProps'
+import {useDefaults} from '../../composables/useDefaults'
+
+const _props = withDefaults(defineProps<BAspectProps>(), {
+  tag: 'div',
+  aspect: '1:1',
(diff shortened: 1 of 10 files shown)


packages/bootstrap-vue-next/src/components/BAspect/BAspect.vue
```diff
@@ -0,0 +1,32 @@
+<template>
+  <component :is="props.tag" class="b-aspect d-flex">
```
VividLemon, line 2: class `b-aspect` & `b-aspect-content` don't exist in this PR and would not have an effect. Not sure if they are needed or just placeholders.

VividLemon (changes requested):

packages/bootstrap-vue-next/src/components/BAspect/BAspect.vue
NilLlisterri: I've kept them for BootstrapVue compatibility. They weren’t styled there either, but may help avoid breaking changes during migration. Happy to remove if we prefer to drop them in this new version.

VividLemon: No, it's fine. There are some other components that have similar class targets. They come in handy when targeting via querySelector. But it was just a mystery if you meant to include specific styles.

VividLemon (approved):

PR #3216, opened 2026-05-25
feat: bootstrap mcp monorepo package

### LINKED ISSUE #2937: useModal create with function type in custom props
Using the useModal composable with a custom component and custom props is working fine since last update 0.40.9. However, when passing a function in a custom prop, the type is not correct and causes a runtime error. Instead of a function type itself, the prop is typed as the function's return type.

On custom prop with function type, i.e.:
```
custom_getter_prop: () => getItems(),
```
It seems to be passed as the function return type instead of a function type.

_Originally posted by @memen45 in https://github.com/bootstrap-vue-next/bootstrap-vue-next/issues/2919#issuecomment-3568368474_

### LINKED ISSUE #3233: Upgrade to @floating-ui/vue  v2
### Describe the bug

v2 was released. This is relevant for consumers of bootstrap-vue-next that also have @floating-ui/vue installed for type resolution. Best to align on latest.

https://github.com/floating-ui/floating-ui/blob/master/packages/vue/CHANGELOG.md

### Reproduction

n/a

### Used Package Manager

pnpm

VividLemon:
# Describe the PR

## Small replication

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)


VividLemon:
vue-tsc bug and vueuse bug (result from vite 8 upgrade)

apps/docs/src/data/migration.data.ts
VividLemon, line 1: I'm unsure what this purpose is just yet.

packages/mcp/eslint.config.ts
VividLemon, line 1: This file should be a ts file

VividLemon:
Requires https://github.com/vueuse/vueuse/pull/5388

VividLemon:
TODO: I need an Ai agent to run through the docs to clarify the usage of the orchestrator methods.

"create" does nothing on its own, it just adds to the entries. You need to use ".show". 

You aren't required to await .show(), but for methods like in BModal, awaiting the value returns something useful, like in using a dialog popup to ask if someone wants to delete, you likely want the return value, as it has meaning.

You can manually take control and use .show and .hide, to modify the visibility of the component. This is used to create long lived entries, like if you want too reuse the same dialog, maybe for perf, maybe to keep inputs in a dialog persistent even though you want to hide the modal. 

"destroy" or "Symbol.asyncDispose" (await using) is used to clear out the entry from the list. ".hide" method will make the component disappear, but that entry still exists in the list, which could exist for the lifetime of your entire application and wont be garbage collected, even on other pages. SO it's important to either use this syntax or manually, try/finally the destroy method, or destroy at a later time. Whatever works for ones application.

Any v-modelable value, from said component's component reference, is a candidate for writing to when passed into the create() method. So if you pass in a ref, it should at least have writable v-model "modelValues" or other two-way bound values. If you need reactive inputs into this method, reactive() is usually the most straightforward path, to maintain two-way reactivity. But if you don't have a reactive input, if it's a "static" object, then you could just pass in an object. The object is wrapped in ref() anyways. So both work. Anything as long as you pass in a value we can write modelValue. So naturally no "readonly()" or non-WritableCOmputed()'s (a regular computed would fail upon a re-computation of computed, naturally, despite the fact that it would "technically" work)

PR #3220, opened 2026-06-11
Fix various bugs around BFormGroup's auto label-for handling.

mmulqueen:
# Describe the PR

Issues before our changes
- Auto label-for picked the last ID it saw if there were one or more inputs.
- Despite the docs saying about leaving the label-for prop empty for multiple inputs, because of the auto label-for behaviour, that was impossible.
- The last wins behaviour of auto label-for could lead to various undesirable situations: Wrong field focused (if any field was to be focused by clicking the label, only the first field makes sense). Where a checkbox/radio was the last field, this is even worse because a click on the form group's label would lead to a selection being made that seems arbitrary to the user.
- Differentiating between fieldset and div[role=group] is pointless. A fieldset implies ARIA role=group. Fieldsets are legal with only 1 input and no legend. So there's no benefit to div[role=group] that I can come up with.
- Auto for-label on the form group was applied even for BFormCheckbox/BFormRadio, which had their own labels already.
- BFormFloatingLabel could be used to label an input already, but that input would still be eligible for auto label-for if contained within a form group.
- BootstrapVue didn't have this auto label-for behaviour, it's a BootstrapVueNext improvement (ce016480358f50653354d27b3d4ad59db58a221f) that happens to have also caused a regression. So in BootstrapVue, if you didn't set label-for you'd unconditionally get fieldset/legend.

Implementation details regarding the proposed fix:
- Implements a register/unregister pattern using provide/inject, similar to what's already used in BTabs and BAccordion. Triggered during initial setup, by ID changes and when unmounted.
- A new `useProvideFormGroupData` composable encapsulates the provide-side machinery (registry, watch, onUnmounted) and is used by both BFormGroup and BFormFloatingLabel - boundaries are just re-providing with passed-through state/disabled and a fresh isolated registry that nothing reads.
- The injection contract is a callback. The arg is an optional `Readonly<Ref<string | null>>`: pass an id-ref to claim label-for; pass nothing to consume state/disabled without claiming. That's how checkbox/radio opt out when they render their own label - the ref resolves to null, the watch sees null, no claim.
- BFormFloatingLabel properly stops the propagation of the inputs for auto label-for purposes.
- Checkboxes/radios with their own internal label do not attempt to register.
- Eliminating the switching between fieldset and div[role=group] is necessary because the :is="..." causes unmount/mount of the children in the default slot - which creates a feedback loop of the children being unregistered/registered again that would never settle.
- The documented behaviour of the disabled prop on BFormGroup was presumably a consequence of toggling between fieldset/div - I saw no reason to preserve this detail, the disabled prop now works everywhere.
- Tests expanded: multi-input fallback to legend, BFormCheckbox/BFormRadio claim-vs-skip cases (one positive, one negative each).
- Docs updated, including multi-input example in the BFormGroup docs.

This is a larger first PR than I'd ideally contribute for the first time, but it all needed to be tackled together.

## Small replication

Before this PR, with no `label-for` set:

```vue
<BFormGroup label="Range">
  <BInputGroup>
    <BFormInput placeholder="Start" />
    <BFormInput placeholder="End" />
  </BInputGroup>
</BFormGroup>
```

BFormGroup rendered `<label for="end-input-id">Range</label>` - pointing at whichever input registered last. Clicking the label focused the wrong field. The docs claimed leaving `label-for` empty for multi-input groups would render `<legend>`, but auto label-for made that impossible. After this PR the same markup renders `<fieldset><legend>Range</legend>...</fieldset>`.

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

- [x] Pull request title and all commits follow the [**Conventional Commits**](https://www.conventionalcommits.org/) convention or has an [**override**](https://github.com/googleapis/release-please#how-can-i-fix-release-notes) in this pull request body

## Summary by CodeRabbit

* **New Features**
  * Added a demo showing a form group with multiple input fields.

* **Bug Fixes**
  * Improved label association for grouped/multi-input controls; falls back to a legend when association is ambiguous.
  * Ensured form groups consistently render as fieldsets and correctly propagate disabled state, including when floating labels or controls provide their own labels.

* **Documentation**
  * Clarified form-group labeling, accessibility guidance, and recommendations for multi-input scenarios.

* **Tests**
  * Updated tests to match new DOM/labeling and disabled behavior.

### DIFF
--- a/apps/docs/src/docs/components/demo/FormGroupMultipleInputs.vue
+++ b/apps/docs/src/docs/components/demo/FormGroupMultipleInputs.vue
@@ -1,0 +1,10 @@
+<template>
+  <!-- #region template -->
+  <BFormGroup label="Enter your name">
+    <BInputGroup>
+      <BFormInput aria-label="Forename" placeholder="Forename" />
+      <BFormInput aria-label="Surname" placeholder="Surname" />
+    </BInputGroup>
+  </BFormGroup>
+  <!-- #endregion template -->
+</template>
--- a/packages/bootstrap-vue-next/src/components/BForm/BFormFloatingLabel.vue
+++ b/packages/bootstrap-vue-next/src/components/BForm/BFormFloatingLabel.vue
@@ -11,7 +11,10 @@
 
 <script setup lang="ts">
 import type {BFormFloatingLabelProps, BFormFloatingLabelSlots} from '../../types'
+import {inject} from 'vue'
 import {useDefaults} from '../../composables/useDefaults'
+import {useProvideFormGroupData} from '../../composables/useProvideFormGroupData'
+import {formGroupKey} from '../../utils/keys'
 
 const _props = withDefaults(defineProps<BFormFloatingLabelProps>(), {
   label: undefined,
@@ -20,4 +23,10 @@ const _props = withDefaults(defineProps<BFormFloatingLabelProps>(), {
 })
 const props = useDefaults(_props, 'BFormFloatingLabel')
 defineSlots<BFormFloatingLabelSlots>()
+
+// Boundary: BFormFloatingLabel renders its own <label for>. State/disabled forward through
+// from the outer BFormGroup; the isolated label-target registry isn't read, so descendant
+// claims never reach the outer group.
+const parentFormGroupData = inject(formGroupKey, null)?.()
+if (parentFormGroupData) useProvideFormGroupData(parentFormGroupData)
 </script>
--- a/packages/bootstrap-vue-next/src/components/BForm/BFormFloatingLabel.vue
+++ b/packages/bootstrap-vue-next/src/components/BForm/BFormFloatingLabel.vue
@@ -27,6 +27,6 @@ defineSlots<BFormFloatingLabelSlots>()
 // Boundary: BFormFloatingLabel renders its own <label for>. State/disabled forward through
 // from the outer BFormGroup; the isolated label-target registry isn't read, so descendant
 // claims never reach the outer group.
-const parentFormGroupData = inject(formGroupKey, null)?.()
+const parentFormGroupData = inject(formGroupKey, null)?.(null)
 if (parentFormGroupData) useProvideFormGroupData(parentFormGroupData)
 </script>
--- a/packages/bootstrap-vue-next/src/components/BForm/BFormFloatingLabel.vue
+++ b/packages/bootstrap-vue-next/src/components/BForm/BFormFloatingLabel.vue
@@ -27,6 +27,6 @@ defineSlots<BFormFloatingLabelSlots>()
 // Boundary: BFormFloatingLabel renders its own <label for>. State/disabled forward through
 // from the outer BFormGroup; the isolated label-target registry isn't read, so descendant
 // claims never reach the outer group.
-const parentFormGroupData = inject(formGroupKey, null)?.(null)
+const parentFormGroupData = inject(formGroupKey, null)?.()
 if (parentFormGroupData) useProvideFormGroupData(parentFormGroupData)
 </script>
--- a/packages/bootstrap-vue-next/src/composables/useProvideFormGroupData.ts
+++ b/packages/bootstrap-vue-next/src/composables/useProvideFormGroupData.ts
@@ -1,4 +1,4 @@
-import {computed, onUnmounted, provide, type Ref, ref, watch} from 'vue'
+import {computed, onUnmounted, provide, type Ref, shallowRef} from 'vue'
 import type {ValidationState} from '../types/CommonTypes'
 import {formGroupKey} from '../utils/keys'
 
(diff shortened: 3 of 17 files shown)


packages/bootstrap-vue-next/src/composables/useProvideFormGroupData.ts
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  // The callback runs in the descendant's setup, so watch/onUnmounted bind to the descendant.
+  provide(formGroupKey, (idRef) => {
+    if (idRef) {
+      watch(
+        idRef,
+        (curr, prev) => {
+          if (prev) unregisterLabelTarget(prev)
+          if (curr) registerLabelTarget(curr)
```
VividLemon: If you made labelTargetIds `shallowRef<Ref<string>[]>` you wouldn't need to watch for child element changes

packages/bootstrap-vue-next/src/composables/useProvideFormGroupData.ts
```diff
@@ -0,0 +1,53 @@
...
+        },
+        {immediate: true}
+      )
+      onUnmounted(() => {
```
VividLemon: coupled with the above comment about removing the watcher, the beset thing to do here is 

`return {state, disabled, unregister}` then have the consumer handle cleanup

(I prefer to not automatically perform cleanups)

packages/bootstrap-vue-next/src/composables/useProvideFormGroupData.ts
```diff
@@ -0,0 +1,53 @@
...
+ * label-target registry: BFormGroup uses the resulting `singleLabelTargetId` to drive its
+ * label/legend rendering. Intermediate components (e.g. BFormFloatingLabel) call this with
+ * passed-through state/disabled, acting as a boundary — their registry exists but is unused,
+ * so descendant claims never reach an outer BFormGroup.
+ */
+export const useProvideFormGroupData = (opts: {
+  state: Readonly<Ref<ValidationState | undefined>>
+  disabled: Readonly<Ref<boolean>>
+}) => {
+  const {state, disabled} = opts
```
VividLemon:
```suggestion
export const useProvideFormGroupData = ({state, disabled}: {
  state: Readonly<Ref<ValidationState | undefined>>
  disabled: Readonly<Ref<boolean>>
}) => {
```

packages/bootstrap-vue-next/src/composables/useProvideFormGroupData.ts
```diff
@@ -0,0 +1,53 @@
...
+  const labelTargetIds = ref<string[]>([])
```
VividLemon: Lets use a shallowRef here.

packages/bootstrap-vue-next/src/composables/useProvideFormGroupData.ts
```diff
@@ -0,0 +1,53 @@
...
+  const registerLabelTarget = (id: string) => {
+    labelTargetIds.value.push(id)
```
VividLemon:
```suggestion
    labelTargetIds.value = [...labelTargetIds.value, id]
```

packages/bootstrap-vue-next/src/composables/useProvideFormGroupData.ts
```diff
@@ -0,0 +1,53 @@
...
+  }
+  const unregisterLabelTarget = (id: string) => {
+    const idx = labelTargetIds.value.indexOf(id)
+    if (idx !== -1) labelTargetIds.value.splice(idx, 1)
```
VividLemon:
```suggestion
    if (idx !== -1) {
        labelTargetIds.value = labelTargetIds.value.toSpliced(idx, 1)
    }
```

Can we use toSpliced? idk. Whatever the functional equivalent code is, prefer the immutable versions please (as would be required when using shallowRef)

apps/docs/src/docs/components/demo/FormGroupMultipleInputs.vue
```diff
@@ -0,0 +1,10 @@
+<template>
+  <!-- #region template -->
+  <BFormGroup label="Enter your name">
+    <BInputGroup>
+      <BFormInput aria-label="Forename" placeholder="Forename" />
+      <BFormInput aria-label="Surname" placeholder="Surname" />
```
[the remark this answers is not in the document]
mmulqueen: This concern seems unfounded. I've manually inspected the generated DOM and there are no ID collisions nor any risk of collisions for this component. I note that some examples on the page have an explicit ID, but there are loads of generated IDs in other examples. We deliberately generate IDs which are guaranteed to be unique, so it seems bizarre to not take advantage of that and may give people reading the docs the impression that they need to specify an ID when they don't.

packages/bootstrap-vue-next/src/components/BForm/BFormFloatingLabel.vue
```diff
@@ -20,4 +23,10 @@ const _props = withDefaults(defineProps<BFormFloatingLabelProps>(), {
 })
 const props = useDefaults(_props, 'BFormFloatingLabel')
 defineSlots<BFormFloatingLabelSlots>()
+
+// Boundary: BFormFloatingLabel renders its own <label for>. State/disabled forward through
+// from the outer BFormGroup; the isolated label-target registry isn't read, so descendant
+// claims never reach the outer group.
+const parentFormGroupData = inject(formGroupKey, null)?.()
```
VividLemon: The logic of the injected function taking an optional parameter feels odd. It might be that it is an edge case and there isn't really a way around it, but perhaps instead of optional, it is `Ref<> | undefined` (and add a jsdoc explaining this. 

It might be worth exploring different options, if there are any that don't cause code duplication.

VividLemon (changes requested):

packages/bootstrap-vue-next/src/composables/useProvideFormGroupData.ts
mmulqueen: This is the only thing I've not addressed in the latest commit. I'm torn. It seems a bit asymmetric to mix - initial registration would be automatic, changes to the value would be automatic (mechanism differs in my first commit to the latest commit, but still automatic), so we'd be singling out onUnmounted - and there's no situation where you'd want to provide a ref and have that omitted. That callback is like a pseudo-composable and it's normal to put lifecycle hooks in a composable.

But take a look at the simpified version, if you still want unregister to be explicitly called by onUnmounted at all 5 call sites I can change that.

Btw, I also used filter instead of toSliced because of the ES version the project uses for TS.

VividLemon: I agree. I suppose the only other comment I'd make about the subject is that the unmounted call shouldn't be in the scope of `provide`. Move it outside of provide() please.

VividLemon (changes requested):

mmulqueen:
I've refactored the provided callback, so it has an explicit track method (got rid of the conditional idRef on the callback itself) and it's more obviously a composable now. The mechanism is the same, but the code structure is more obvious. I discovered gaps in test coverage relating to this PR around floating labels and fixed that too. There's a broader diff because of the new call signature at the inject sites. This whole change feels cleaner.

VividLemon:
@coderabbitai review

VividLemon (approved):

PR #3248, opened 2026-08-11
fix: ensure BFormGroup label is rendered as <label> in floating mode for SSR compatibility

jonty-comp:
## Describe the PR

Fixes an SSR/hydration mismatch in `BFormGroup` when the `floating` prop is used. The label was rendered as a `<legend>` element on the initial render pass, only switching to a `<label>` once the descendant input registered its id — which happens after the parent's render pass, during hydration. Because SSR serializes only the first render pass, the server HTML contained a `<legend>`, which Bootstrap's `.form-floating > label` selector does not style. The label therefore rendered below the input's bounding box (increasing the element's height), and after hydration the element flipped to a `<label>`, causing a full repaint and layout shift — visible as a flash in Nuxt SSR apps.

The fix ensures the floating label is always a `<label>` element from the first render by deriving the label tag from `isFieldset` (now floating-aware) instead of the child input's id. The `for` attribute still wires up once the child registers, but as an attribute-only change it has no layout impact. Non-floating behavior is unchanged — verified byte-identical against the old code via a before/after test matrix (input, textarea, select, checkbox, radio, horizontal, `labelFor`, and auto-generated id combinations). As a bonus, this also removes the incorrect `tabindex="-1"` from the floating label and gives it a proper programmatic association.

## Small replication

SSR output for `<BFormGroup floating label="Email"><BFormInput /></BFormGroup>` before the fix:

```html
<div class="form-floating">
  <input id="BootstrapVueNext__ID__v-5__input___" class="form-control" type="text">
  <legend tabindex="-1" class="form-label bv-no-focus-ring col-form-label pt-0">Email</legend>
</div>
```

After the fix (identical to the hydrated DOM):

```html
<div class="form-floating">
  <input id="BootstrapVueNext__ID__v-5__input___" class="form-control" type="text">
  <label class="form-label d-block">Email</label>
</div>
```

The label now floats inside the input from first paint; hydration only adds the `for` attribute.

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved floating form groups to render the correct label structure.
  - Ensured floating groups use appropriate label styling, focus behavior, spacing, and display rules.
  - Prevented horizontal form groups from incorrectly using floating behavior.
  - Preserved correct input ID association after child fields register.

- **Tests**
  - Added coverage for floating labels and server-side rendering.


packages/bootstrap-vue-next/src/components/BFormGroup/BFormGroup.vue
```diff
@@ -216,6 +216,7 @@ const labelColProps = computed(() => getColProps(props, 'label'))
 const isHorizontal = computed(
   () => Object.keys(contentColProps.value).length > 0 || Object.keys(labelColProps.value).length > 0
 )
+const floating = computed(() => props.floating && !isHorizontal.value)
```
VividLemon, line 219: This variable needs to be renamed. It is a shadow of a prop.

VividLemon (changes requested):

PR #3330, opened 2026-09-11
feat(useToast): add `hide` and `hideAll`, matching `useModal`

### LINKED ISSUE #3327: useToast: add hide/hideAll, matching useModal

As a developer using BootstrapVueNext I want to dismiss a toast by id — and dismiss all toasts —
without tracking controllers myself, so that "show a notification" and "take that notification back"
are equally easy.

`useModal` already has exactly this:

```ts
const hide = (trigger, id) => { ... }
const hideAll = (trigger) => { ... }
```

`useToast` has neither. It returns only `{ _isToastAppend, _isOrchestratorInstalled, store, create }`,
so the only handle on a toast is the `ComponentController` returned by `create()`. That leaves callers
who dismiss toasts in response to something other than user interaction — a background refresh, a
route change, a websocket message — keeping a parallel registry of live controllers just to find the
one to hide later.

Two things make that registry harder than it sounds:

1. **`create()` throws before the orchestrator is mounted** (`"The BApp component must be mounted to
   use the toast composable"`). So a notification emitted early has to be queued by the caller until
   `_isOrchestratorInstalled` flips. While it sits in that queue it has no controller, so it is
   invisible to any dismissal logic and pops onto the screen after it was already cleared.
2. **`create()` with an id that is already in the store silently replaces the entry.** The previous
   controller now resolves to the replacement's store entry, so the caller has to detect that and
   stop touching it, or it will act on a toast it no longer owns.

The result is that "dismiss this toast" needs a set of pending entries, a cancellation flag, and a
superseded flag — all of it bookkeeping the library is better positioned to do, and all of it
unnecessary for modals.

### Suggested solution

Give `useToast` the same two functions `useModal` has, implemented against the same orchestrator store:

```ts
const { hide, hideAll } = useToast()

hide('clear', 'session-expiring')  // dismiss one toast by id
hideAll('route-change')            // dismiss all toasts
```

`useModal`'s implementation looks like it ports over nearly as-is — for the orchestrator-store case
it sets `modelValue: false` on the entry and lets the normal hide cycle run.

Separately (happy to split this into its own issue if you'd prefer): allow `create()` to be called
before the orchestrator mounts, queuing the toast until it is. `create()` only writes to a reactive
store that `BOrchestrator` renders from whenever it mounts, so the throw reads like a DX guard rather
than a technical requirement. Lifting it would remove the need for callers to build a queue in front
of `create()` at all — and with it the window where a toast exists conceptually but has no controller.

### Alternative

Callers can track controllers themselves, which is what we do today. It works, but every consumer
that dismisses toasts programmatically re-implements the same three pieces of state, and the
queued-before-mount window is easy to miss entirely — we shipped it as a bug and only found it in
review.

Reaching into `store.value.toast` and setting `modelValue: false` on the entry directly also works,
but that is internal shape, not public API, and it is a straight duplicate of what `useModal.hide`
already does.

### Additional context

Related: `ComponentController.destroy()` cannot be used for this, because it deadlocks on a visible
item. Filing a separate bug report for that.

IGx89:
## Describe the PR

Adds `hide` and `hideAll` to `useToast`, covering the first half of #3327. A toast can now be dismissed by id from anywhere in the app, without the caller keeping the `ComponentController` that `create` returned.

```ts
const {hide, hideAll} = useToast()

hide('clear', 'session-expiring') // dismiss one toast by id
hideAll('route-change') // dismiss every toast
```

The `create()`-before-mount half of the issue is deliberately left out of this PR.

### How it resolves a toast

`hide(trigger, id)` tries, in order:

1. the mounted `BToast` registered under that id in the show/hide registry — going through the component is what lets `trigger` flow through the regular hide cycle, and it also picks up a plain `<BToast id="...">` declared in a template. That registry is shared by every show/hide component, so the lookup walks the id's instances newest-first (the way the registry's own `getActive` does) and takes the first `BToast`, leaving a `BCollapse` or `BModal` that carries the same id alone
2. the orchestrator store entry, setting `modelValue: false` the way `useModal.hide` does — the only path for a toast the orchestrator has yet to render (so a toast can be dismissed before it ever becomes visible), for toasts created without an id (symbols never reach the registry), and for a toast rendered through a custom `component`, which is not a `BToast` instance
3. nothing, if neither matches

Calling `hide(trigger)` with no id hides every toast, matching `$bvToast.hide()` in BootstrapVue. `hideAll(trigger)` walks the store and hides each entry through the same path.

Two known edges, both documented on the composable page:

- a toast hidden through its store entry reports `modelValue` as its `trigger`, since there is no component to run the hide cycle through. `useModal` has the same limitation on its store branch. Fixing it properly would mean exposing `hide` on the store entry's `fns`, which is a type shared with modal/popover/tooltip — happy to do that here instead if you would prefer it
- `hideAll` covers the store and not template-declared toasts. An orchestrated toast created without an id cannot be correlated with its registry entry, so including the registry in `hideAll` would hide some toasts twice

## PR checklist

- [x] Tests — `useToast.spec.ts` gains hide by id with trigger propagation, hide by symbol id through the store, hide of a template-declared toast, unknown-id no-op, two id-collision cases (a `BCollapse` sharing the id is left alone, and the toast is the one hidden when both exist), `hideAll` over two toasts, `hide()` with no id, and hiding a toast the orchestrator has not rendered yet
- [x] Docs — new "Globally Hiding Toasts" section plus a `UseToastGlobalHide.vue` demo
- [x] `pnpm --filter bootstrap-vue-next run test:lint`, `test:unit:ci` (5556 pass), `build`, `type-check`, and the docs build/lint all pass

## Unrelated notes

- The pre-commit hook runs `prettier --write` on library files while the package's `format` script uses `oxfmt`; the two disagree on `useToast/index.ts`, so the hook re-wrapped one pre-existing generic default. That is the 2-line deletion in the diff
- Separately, `BOrchestrator` binds the create payload's `onHide` via `v-bind="componentProps"` **and** re-invokes it from its own `@hide` handler, so a consumer's `onHide`/`onHidden` fires twice for orchestrated items. Pre-existing and untouched here; the new tests compare hide events by identity to work around it. Happy to file it separately

Closes part of #3327

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

- **New Features**
  - Added global toast controls through `useToast`, allowing individual toasts to be hidden by ID or all composable-created toasts to be hidden at once.
  - Hiding supports programmatically created and template-declared toasts, including toasts that have not yet rendered.
  - Added a documentation demo showing how to display and dismiss sample toasts.

- **Documentation**
  - Documented global toast hiding, including ID-based and hide-all behavior.

### DIFF
--- a/apps/docs/src/docs/composables/demo/UseToastGlobalHide.vue
+++ b/apps/docs/src/docs/composables/demo/UseToastGlobalHide.vue
@@ -1,0 +1,39 @@
+<template>
+  <BButtonGroup>
+    <BButton variant="success" @click="notify"> Show Toasts </BButton>
+    <BButton variant="warning" @click="hide('user-request', 'session-expiring')">
+      Hide Session Toast
+    </BButton>
+    <BButton variant="danger" @click="hideAll('route-change')"> Hide All </BButton>
+  </BButtonGroup>
+</template>
+
+<script setup lang="ts">
+import { BButton, BButtonGroup } from 'bootstrap-vue-next/components/BButton'
+import { useToast } from 'bootstrap-vue-next/composables/useToast'
+
+const { create, hide, hideAll } = useToast()
+
+// Shows the toast, then disposes of it once it has been hidden
+const show = async (payload: Parameters<typeof create>[0]) => {
+  await using toast = create(payload)
+  await toast.show()
+}
+
+const notify = () => {
+  void show({
+    id: 'session-expiring',
+    title: 'Session expiring',
+    body: 'You will be signed out soon',
+    variant: 'warning',
+    position: 'bottom-center',
+  })
+  void show({
+    id: 'sync-complete',
+    title: 'Sync complete',
+    body: 'Everything is up to date',
+    variant: 'success',
+    position: 'bottom-center',
+  })
+}
+</script>
--- a/apps/docs/src/docs/composables/useToast.md
+++ b/apps/docs/src/docs/composables/useToast.md
@@ -70,12 +70,14 @@ Toasts can also be hidden from anywhere in the app, without holding on to the co
 
 - `hideAll: (trigger?: string) => void`
 
-  Hides every toast
+  Hides every toast that was created through the composable
 
-Both functions hide toasts that were created through the composable, as well as `BToast` components
-declared in a template that were given a matching `id`. A toast that the orchestrator has not
-rendered yet is hidden through its store entry, so it never becomes visible. Those toasts report
-`modelValue` as their `trigger`, since there is no component to run the hide cycle through.
+`hide(trigger, id)` also hides a `BToast` that was declared in a template with a matching `id`.
+`hideAll` and `hide` without an `id` only cover the toasts that were created through the composable.
+
+A toast that the orchestrator has not rendered yet is hidden through its store entry, so it never
+becomes visible. Such a toast reports `modelValue` as its `trigger`, since there is no component to
+run the hide cycle through. The same applies to a toast rendered through a custom `component`.
 
 Hiding a toast does not remove it from the orchestrator store, see
 [Lifecycle and disposal](#lifecycle-and-disposal).
--- a/apps/docs/src/docs/composables/useToast.md
+++ b/apps/docs/src/docs/composables/useToast.md
@@ -55,6 +55,31 @@ Hiding a `Toast` programmatically is simple. The controller returned by `create`
 
 <<< DEMO ./demo/UseToastProgrammatic.vue
 
+## Globally Hiding Toasts
+
+Toasts can also be hidden from anywhere in the app, without holding on to the controller that
+`create` returned. Give the toast an `id`, then hide it by that `id`:
+
+<<< DEMO ./demo/UseToastGlobalHide.vue
+
+- `hide: (trigger?: string, id?: ControllerKey) => void`
+
+  Hides the toast with the given `id`. The `trigger` is passed to the `trigger` property of the
+  `hide` event and of the resolved `show` promise. When no `id` is given, every toast is hidden,
+  just like `hideAll`
+
+- `hideAll: (trigger?: string) => void`
+
+  Hides every toast
+
+Both functions hide toasts that were created through the composable, as well as `BToast` components
+declared in a template that were given a matching `id`. A toast that the orchestrator has not
+rendered yet is hidden through its store entry, so it never becomes visible. Those toasts report
+`modelValue` as their `trigger`, since there is no component to run the hide cycle through.
+
+Hiding a toast does not remove it from the orchestrator store, see
+[Lifecycle and disposal](#lifecycle-and-disposal).
+
 ## Lifecycle and disposal
 
 Created toast instances persist until you explicitly dispose them. Hiding a toast does not remove it from the orchestrator store.
--- a/packages/bootstrap-vue-next/src/composables/useToast/index.ts
+++ b/packages/bootstrap-vue-next/src/composables/useToast/index.ts
@@ -105,20 +105,28 @@ export const useToast = () => {
    * Hides a mounted BToast that registered itself under `id`. Going through the component, rather
    * than through the store, is what lets the trigger flow through the regular hide cycle
    *
-   * @returns whether an instance was found
+   * @returns whether a toast was found
    */
   const hideRegisteredInstance = (id: ControllerKey, trigger?: string): boolean => {
     // The show/hide registry is keyed by the html id, symbols never reach it
     if (typeof id !== 'string') return false
-    const instance = showHideRegistry?.values.value.get(id)?.getActive()
-    if (!instance) return false
-    instance.hide(trigger, true)
-    return true
+    const instances = showHideRegistry?.values.value.get(id)?.instances
+    if (!instances) return false
+    // Every show/hide component shares this registry, so an id may resolve to a BCollapse or a
+    // BModal. Only toasts are ours to hide. Newest first, the way the registry's getActive does
+    for (let i = instances.length - 1; i >= 0; i--) {
+      const instance = instances[i]
+      if (instance?.component.type !== BToast) continue
+      instance.hide(trigger, true)
+      return true
+    }
+    return false
   }
 
   /**
    * Hides a toast that lives in the orchestrator store. Prefers the mounted instance, falling back
    * to the store entry itself, which also covers toasts that the orchestrator has yet to render
+   * and toasts rendered through a custom `component`
    */
   const hideStoreItem = (item: Ref<ToastOrchestratorArrayValue>, trigger?: string): void => {
     if (hideRegisteredInstance(item.value.props.id ?? item.value.id, trigger)) return
--- a/packages/bootstrap-vue-next/src/composables/useToast/index.ts
+++ b/packages/bootstrap-vue-next/src/composables/useToast/index.ts
@@ -9,10 +9,15 @@ import {
   ref,
   type Ref,
 } from 'vue'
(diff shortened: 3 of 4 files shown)


packages/bootstrap-vue-next/src/composables/useToast/index.ts
```diff
@@ -94,10 +101,71 @@ export const useToast = () => {
...
+  /**
+   * Hides a mounted BToast that registered itself under `id`. Going through the component, rather
+   * than through the store, is what lets the trigger flow through the regular hide cycle
+   *
+   * @returns whether an instance was found
+   */
+  const hideRegisteredInstance = (id: ControllerKey, trigger?: string): boolean => {
+    // The show/hide registry is keyed by the html id, symbols never reach it
+    if (typeof id !== 'string') return false
+    const instance = showHideRegistry?.values.value.get(id)?.getActive()
```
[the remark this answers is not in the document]
IGx89: Valid, and the more serious of the two findings — fixed in af787c6.

Confirmed the collision: `useShowHide` registers `BAlert`, `BCollapse`, `BDropdown`, `BModal`, `BOffcanvas`, `BPopover` and `BToast` into that one registry by id, so `hide(trigger, "sidebar")` would hide a `<BCollapse id="sidebar">`. The lookup now walks the id's instances newest-first and takes the first `BToast`, which keeps `getActive()`'s newest-wins semantics. A side effect worth naming: an orchestrated toast rendered through a custom `component` no longer matches the registry, so it falls through to the store entry, which still hides it (with `modelValue` as the trigger).

Two tests added. `leaves other show/hide components that share the id alone` fails against the previous code, so it is a real regression test; `hides the toast, not the component that shares its id` passes either way, because registration order happens to put the orchestrated toast last — kept as a behavioral guard rather than a regression test.

apps/docs/src/docs/composables/useToast.md
```diff
@@ -55,6 +55,31 @@ Hiding a `Toast` programmatically is simple. The controller returned by `create`
...
+  just like `hideAll`
+
+- `hideAll: (trigger?: string) => void`
+
+  Hides every toast
+
+Both functions hide toasts that were created through the composable, as well as `BToast` components
+declared in a template that were given a matching `id`. A toast that the orchestrator has not
+rendered yet is hidden through its store entry, so it never becomes visible. Those toasts report
+`modelValue` as their `trigger`, since there is no component to run the hide cycle through.
```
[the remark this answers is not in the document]
IGx89: Correct, the original wording overstated it — fixed in af787c6.

Reworded rather than taking the suggestion verbatim, to the same effect: `hide(trigger, id)` is the only form that reaches a template-declared `BToast`, and `hideAll` / `hide` without an id cover only toasts created through the composable. Also narrowed the `hideAll` bullet itself, which still said "hides every toast", and noted that a toast rendered through a custom `component` reports `modelValue` as its trigger for the same reason an unrendered one does.

apps/docs/src/docs/composables/demo/UseToastGlobalHide.vue
VividLemon, line 1: If this is meant to be a demo, it's best in the playground. Otherwise, if it's more useful for unit tests, it could be just transferred to a test. We don't really embed .vue demo files.

apps/docs/src/docs/composables/useToast.md
```diff
@@ -55,6 +55,33 @@ Hiding a `Toast` programmatically is simple. The controller returned by `create`
...
+
+- `hide: (trigger?: string, id?: ControllerKey) => void`
+
+  Hides the toast with the given `id`. The `trigger` is passed to the `trigger` property of the
+  `hide` event and of the resolved `show` promise. When no `id` is given, every toast is hidden,
+  just like `hideAll`
+
+- `hideAll: (trigger?: string) => void`
+
+  Hides every toast that was created through the composable
```
VividLemon, line 73: Can the modal controller do this? I don't remember. I think it can. If it can, why can useToast not?

IGx89: You're asking if useModal's hideAll hides modals declared in templates, and if so, if useToast's hideAll can be enhanced to support that too?

VividLemon: If the modal controller does it, then for consistency this one should do the same, yes. We'd just need to make sure that the modal controller also does it.

IGx89: Thanks for clarifying! I'll take a look.

IGx89: Claude's analysis -- which option would you prefer?

> Confirmed: useModal.hideAll iterates the shared modal stack, and useModalManager pushes every mounted BModal onto it when shown — so yes, it covers template-declared modals. (It only hides visible ones; an orchestrated modal still unrendered in the store isn't touched.)
> 
> Toasts have no equivalent stack, only the shared show/hide registry. The blocker for a straight port: a toast created without an id registers under a generated key, so it can't be matched back to its store entry. A registry sweep in hideAll on top of the store pass would hide those twice — two hide cycles, onHide firing with two different triggers.
> 
> Two ways to close that:
> 
> A. Skip orchestrated instances in the sweep. hideAll hides store entries as it does now, then hides registry BToasts that don't have BOrchestrator in their parent chain. Contained in useToast; same idiom as useModal.current()'s findBModal. Assumes BOrchestrator is what renders store entries.
> 
> B. Give anonymous toasts a generated id in create(). Correlation becomes plain id matching, and every toast then hides through its component, so the caller's trigger survives instead of becoming modelValue. Costs: controller.id and store keys stop being symbols, toasts gain a DOM id, and it needs a new id generator — useId wraps Vue's useId(), which is setup-only, and create() is often called from handlers.
> 
> Happy to do either. B is cleaner but changes create() behavior, and getOrchestratorControllerId is shared with modal/popover/tooltip, so that feels like your call.

apps/docs/src/docs/composables/demo/UseToastGlobalHide.vue
IGx89: Hmm. apps/docs/src/docs/composables/demo/ holds UseToastBasic.vue, UseToastProgrammatic.vue and the rest, and CONTRIBUTING.md documents the <<< DEMO convention. Still prefer the demo dropped, or better to match its neighbors?

IGx89: Sorry, it's AGENTS, not CONTRIBUTING that documents it (`("Use documented demo/fragment conventions for docs snippets (DEMO/FRAGMENT)"`). Whatever your preference!

apps/docs/src/docs/composables/useToast.md
VividLemon: It may be best to skip it for now and open a separate issue for it. I had forgotten that the whole "useModalManager.ts" thing existed. Although the useModalManagere isn't that complicated, and could be "easily" modified to be more generic than just for modals. One could create two separate stacks, and thus be capable of accomplishing the same thing.

apps/docs/src/docs/composables/demo/UseToastGlobalHide.vue
VividLemon: Ignore me completely, I misread the path. This is a docs file, I had thought that you had created a new /demo in /bootstrap-vue-next/src/composables/demo . No, demo files are the way to do this in the docs, yes. So this is correct.

apps/docs/src/docs/composables/useToast.md
IGx89: Sounds good, thanks!

VividLemon (approved):

PR #3341, opened 2026-09-18
docs: replace dead placekitten.com images in avatar demos

NotAFlightRisk:
# Describe the PR

Four of the avatar demos set a `src` and still render the default grey silhouette. `placekitten.com` is gone. 521 on the apex and on both of the paths those demos point at, so its the host and not a path.

Moved them over to `picsum.photos`, which the rest of the docs alredy use, including `AvatarImage.vue` sat in the same folder. The snippet printed under each demo is that same file, so copying the Overview example off the page hands you the dead URL too.

I've left the "Super Kitty" label in the list group example alone. It sits with "J. Circlehead" and "ACME group", so I read it as a daft name rather than a caption, but thats your demo copy and your call.

There's no issue open about it, which figures: `BAvatar` catches the image error and falls back to the silhouette, so the page never looks broken. It just looks like those demos never had a `src` in the first place.

## Small replication

Four URLs, so theres not much to replicate:

```
$ curl -sL -o /dev/null -w '%{http_code}\n' https://placekitten.com/300/300
521
$ curl -sL -o /dev/null -w '%{http_code} %{content_type}\n' https://picsum.photos/300/300
200 image/jpeg
```

On the [avatar page](https://bootstrap-vue-next.github.io/bootstrap-vue-next/docs/components/avatar.html), Overview and Avatar groups show silhouettes, while Image content further down is already on picsum and renders fine.

## PR checklist

- [ ] Feature - `feat(...)`
- [ ] Other (please describe)

- [x] Pull request title and all commits follow the [**Conventional Commits**](https://www.conventionalcommits.org/) convention

---

Six more `placekitten.com` links live over in `apps/playground` (`TAvatar.vue`, `TCard.vue`). Left them out to keep this one quick to read, but it's the same dead host if you want them swept.

## Summary by CodeRabbit

- **Documentation**
  - Updated avatar examples in the documentation demos to use reliable image sources.


VividLemon (approved):

VividLemon:
tyvm

